PR #2, opened 2020-02-15
Update how-to-train.md

BramVanroy:
`.open()` files should be closed. As currently written, the opened file handles were never closed - which might cause issues when the same file is used around different processes at the same time. This proposed change uses pathlib's built-in `read_text` which opens the file, reads contents, and closes it immediately in one go.


julien-c:
Good catch, thanks! will fix the whitespace in next commit

PR #8, opened 2020-02-28
Add Tokenizer config file

vochicong:
Add

```
tokenizer_config = {
	"max_len": 512
}
with open("./EsperBERTo/tokenizer_config.json", 'w') as fp:
    json.dump(tokenizer_config, fp)
```

to fix `ValueError: num_samples should be a positive integer value, but got num_samples=0`


notebooks/01_how_to_train.ipynb
```diff
@@ -246,18 +225,25 @@
...
         "!mkdir EsperBERTo\n",
         "tokenizer.save(\"EsperBERTo\")"
       ],
-      "execution_count": 0,
+      "execution_count": 5,
       "outputs": [
+        {
+          "output_type": "stream",
+          "text": [
+            "mkdir: cannot create directory ‘EsperBERTo’: File exists\n"
```
julien-c: here we could use `mkdir -p` maybe

julien-c:
As expected, it's a bit tedious to review a diff on a `.ipynb` file, but it looks good, thanks for fixing this!

cc @aditya-malte

PR #14, opened 2020-03-19
Fix left quotes throughout article

sheikheddy:
The left quotes were bothering me so I'm fixing them. Using backticks didn't work so I'm using single quotes instead. They should be rendered properly, but feel free to test it first.


patrickvonplaten:
Hey @sheikheddy sorry tried it out - doesn't work either. We compile the latex code via Katex and it seems to have a different behavior than Latex. I guess we gotta accept the annoying quotes for the moment :-/

PR #15, opened 2020-03-19
fix quotes

patrickvonplaten:
Thanks @sheikheddy  for pointing this out. I changed the right quotes as well (think this is the standard to quote in Latex).


sheikheddy:
Cool, I'll close my pull request then!

sheikheddy:
Wait, I just checked the article and it didn't render properly. Did you use `` instead of ''? That causes an issue with the markdown if I'm not wrong.

patrickvonplaten:
> Wait, I just checked the article and it didn't render properly. Did you use `` instead of ''? That causes an issue with the markdown if I'm not wrong.

Yeah you're correct. Problem though is that, the other PR doesn't change the quotes sadly. If you have another idea - let me know :-)

sheikheddy:
Quick hack might be to take the quotes outside of the \text().

PR #21, opened 2020-05-23
fixed example

Nikoschenk:
Most probable sequence at step 1 should be "The, nice" instead of "The, woman"


qqaatw:
Fixed in #139 .

PR #22, opened 2020-05-23
GlueDatabase doesn't have keyword argument evaluate, change to mode s…

rmorain:
…et to dev


julien-c:
Correct, thanks for the fix!

PR #26, opened 2020-07-03
Add reformer notebook

patrickvonplaten:
Adds Reformer blog post


julien-c:
should this be in github.com/huggingface/notebooks instead?

julien-c:
Ooops too late I guess :-/

patrickvonplaten:
Will have to update the notebook, the latex equations are not rendered nicely: https://huggingface.co/blog/reformer

patrickvonplaten:
Actually, the same is true for the generate blog post: https://huggingface.co/blog/how-to-generate

PR #38, opened 2020-07-08
Update how-to-train.md

alexyalunin:


julien-c:
@n1t0 will confirm but I think this is correct

PR #49, opened 2020-09-03
Fixing indentation error

benber86:
The indentation issue fixed by #47 introduced a new error. With the new code only the last file in `src_files` would be read and tokenized instead of every file in the corpus data folder.


julien-c:
Thanks!

PR #62, opened 2020-11-24
01_how_to_train.ipynb: Fix TypeError

orestisfl:
`prediction_loss_only` should go to `TrainingArguments`

### DIFF
--- a/notebooks/01_how_to_train.ipynb
+++ b/notebooks/01_how_to_train.ipynb
@@ -1201,6 +1201,7 @@
         "    per_gpu_train_batch_size=64,\n",
         "    save_steps=10_000,\n",
         "    save_total_limit=2,\n",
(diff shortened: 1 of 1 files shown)


harsham05:
@orestisfl Thanks for fixing this, what is the purpose of setting `prediction_loss_only` equal to True while training a language model from scratch?

ZakMiller:
Any reason this isn't merged? The colab linked [here](https://huggingface.co/blog/how-to-train) is broken.

julien-c:
@sgugger looks good to me, can you give it a quick look and merge if it's right?

sgugger:
Yes, that argument was removed from the main `Trainer` in v4 (it's a `TrainingArgument` now but shouldn't be necessary since not `compute_metrics` is passed).

Thanks for fixing!

PR #65, opened 2020-12-15
Add dependency to notebook

jasonk33:
Add `!pip3 install --upgrade tensorflow-gpu` to install cell in order to avoid import error

### DIFF
--- a/notebooks/02_how_to_generate.ipynb
+++ b/notebooks/02_how_to_generate.ipynb
@@ -81,6 +81,7 @@
       "source": [
         "!pip install -q git+https://github.com/huggingface/transformers.git\n",
         "!pip install -q tensorflow==2.1"
+        "!pip3 install --upgrade tensorflow-gpu"
       ],
       "execution_count": 0,
       "outputs": []


jasonk33:
@julien-c @patrickvonplaten

notebooks/02_how_to_generate.ipynb
```diff
@@ -81,6 +81,7 @@
       "source": [
         "!pip install -q git+https://github.com/huggingface/transformers.git\n",
         "!pip install -q tensorflow==2.1"
+        "!pip3 install --upgrade tensorflow-gpu"
```
patrickvonplaten, line 84:
```suggestion
        "!pip install -q --upgrade tensorflow-gpu"
```

patrickvonplaten: pip should be fine and -q to not output all the download information

patrickvonplaten (approved):

sgugger:
Could you explain the context to me? There is no need to install TensorFlow on colab (that might actually mess up things more than help), and if the user is not in colab, `pip install tensorflow-gpu` is not enough (there is some cudnn to properly install too) unless they are in an env where tensorflow is already installed.

jasonk33:
@sgugger I'm not sure exactly the context of why this needs to be installed, but when running the code on multiple environments, it only worked with that install included.

sgugger:
The environment in which the notebook is executed should be managed outside of the notebook by the user themselves. I think this should be replaced by a markdown cell pointing a link to the installation page of TensorFlow. The `! pip install xxx` are hacks for Colab only, since users don't have access to a terminal to execute those instructions, it shouldn't be generalized.

PR #76, opened 2021-01-19
add resource

stas00:
a very important paper has just been published that explains ZeRO-Offload. Adding it to the resources section.

@samyam


n1t0:
Thanks!

PR #84, opened 2021-02-13
Fixed broken link in 01_how_to_train.ipynb

lrjball:
Fixing broken link to [run_language_modeling.py](https://github.com/huggingface/transformers/blob/master/examples/legacy/run_language_modeling.py)


julien-c:
We should probably upgrade to the new script, but in the meantime, this works. Thanks!

PR #87, opened 2021-03-09
Added image credits

TevenLeScao:
Added image credits for the diagrams - with HTML formatting inside the Markdown


julien-c:
Not 100% sure how it will render but LGTM

PR #88, opened 2021-03-10
Update PyTorch/XLA blog post wheel version

jysohn23:
Colab updated their python runtime from 3.6 to 3.7 (`torch_xla` isn't on pypi yet as it's not manylinux compatible).

cc: @LysandreJik


LysandreJik (approved):
Thanks, LGTM!

PR #93, opened 2021-03-24
Added Julien Simons Video as resource

philschmid:
Added Julien Simons Youtube video as resource. https://youtu.be/leyrCgLAGjM


julien-c:
cc @juliensimon 👍

PR #95, opened 2021-03-31
Big Bird

patrickvonplaten:
Thanks goes to @vasudevgupta7


vasudevgupta7:
Thanks a lot for fixing everything & merging it here!

stas00:
Reading  https://huggingface.co/blog/big-bird#what-tokens-should-be-attended-to I noticed that the text refers to left and right, but the actual images are laid out top and bottom - is it a bug in the blog's layout or was it broken by the redesign of the website?

snapshot_72

Additionally half the code blocks can't be read without scrolling. It appears in the new design you only have  76 chars width:

snapshot_73

Thank you.

PR #136, opened 2021-08-06
add AutoNLP UI blog post on using it to enter a Kaggle competition

ferdi05:
Adding the blog post on AutoNLP web interface


kaggle-competition-without-nlp.md
```diff
@@ -0,0 +1,196 @@
+---
+title: "How to beat the best in an NLP competition without any Machine Learning"
+thumbnail: /blog/assets/25_kaggle_competition/thumbnail.png
```
julien-c, line 3: the `thumbnail.png` you added won't work, it has to be a social thumbnail with the same dimensions as the other ones (check other posts)

julien-c:
Can you also ping Clem for review on this?
In particular, if our goal is to expose AutoNLP to software engineers, it might make more sense to publish this post on a software engineers' community, vs our own blog, no?

kaggle-competition-without-nlp.md
ferdi05: I did not know how to test the thumbnail, thanks for the help

ferdi05:
> Can you also ping Clem for review on this?
done

> In particular, if our goal is to expose AutoNLP to software engineers, it might make more sense to publish this post on a software engineers' community, vs our own blog, no?
There is no blog post about AutoNLP (CLI or web interface), so I thought this would fill a gap.
Also I think that we have to get developers to come on Hugging Face website.
I plan to promote it to the developer communities, including referencing it/cross posting in those communities.
But if this is not the right thing to do, I'll happily publish it on dev.to for instance

clmnt:
I tend to agree with Julien to try to post it somewhere else than on the HF blog.

I don't think it passes the bar for the blog where our articles are more about big announcements or very technical/sciency topics and if the goal is to attract a different audience & with a personal tone, it would be better to try to take advantage of someone else's community. Can you brainstorm other places where we could publish it @ferdi05 ? Maybe also a first step is to do a tweet thread to see if the topic gets people's interest?

ferdi05:
> Can you brainstorm other places where we could publish it @ferdi05 ? Maybe also a first step is to do a tweet thread to see if the topic gets people's interest?

Sure, find it at https://www.notion.so/huggingface2/AutoNLP-UI-20aabcd4a85e4af687b7b2818f16b766#2d011bb52f034a7e88ab7c5d5637af92

cc @clmnt @julien-c

PR #147, opened 2021-09-08
Indicating a higher version of `datasets`

rcgale:
A problem with version 1.4.1 of `datasets` caused this tutorial to fail, since the transcripts for the train set of TIMIT were all labeled erroneously as "Would such an act of refusal be useful?"

I've tested with 1.11, but per https://github.com/huggingface/datasets/issues/2879#issuecomment-915061528 the issue was fixed in 1.5.0.


julien-c:
This looks good to me, WDYT @lhoestq?

lhoestq (approved):
Good catch, thanks for the fix !

PR #160, opened 2021-10-05
Fix html & latex

mishig25:
There were some small errors regarding HTML & latex


osanseviero (approved):
Thanks for this!

julien-c:
wow how did you catch those?

mishig25:
@julien-c I have a branch where blog gets compiled through mdsvex rather than lib/Markdown.ts
And these errors were discovered

julien-c:
(deployed)

PR #183, opened 2021-12-07
Update links (Data Measurements Tool)

mariosasko:


julien-c:
blog deployment is manual, just deployed it

PR #185, opened 2021-12-09
New blog look

mishig25:
Companion to https://github.com/huggingface/moon-landing/pull/1632


_blog.yml
```diff
@@ -2,233 +2,304 @@
   title: How to train a new language model from scratch using Transformers and Tokenizers
   thumbnail: ./assets/01_how-to-train/how-to-train_blogpost.png
   date: February-May, 2020
+  tag: guide
```
julien-c, line 5: should maybe be an array of tags for future-proofness?

_blog.yml
```diff
@@ -140,6 +140,7 @@
 
 - local: big-bird
   title: "Understanding BigBird's Block Sparse Attention"
+  thumbnail: ./assets/18_big_bird/block-sparse-attn.gif
```
julien-c, line 143: maybe mention that in this file it can be GIFs while in the blogpost itself it's better if it's a simple bitmap (because it will be used as a social thumbnail)

_blog.yml
mishig25: done so https://github.com/huggingface/blog/blob/0488d18ef5e3dbb6d36a37e4616a6de648e858b7/_blog.yml#L5-L6

osanseviero (approved):
From my side no further suggestions, this looks good!

_events.yml
```diff
@@ -0,0 +1,46 @@
+- name: Intro to Transfer Learning, end-to-end deployment with TFX, and demo with Streamlit
```
julien-c, line 1: (nit) maybe jump lines between items? for consistency with _blog.yml

_blog.yml
julien-c: up on this (document somewhere as a comment should be enough)

julien-c (approved):
LGTM to me too

PR #205, opened 2022-01-13
Add Infinity CPU case study post

philschmid:
New Blog post of Infinity about cpu performance

P.S. i fixed the order of the blog posts, moved GPTJ before wav2vec2 to fix date order


jeffboudier (approved):
lgtm!

PR #225, opened 2022-02-07
Add modalities tag to blog

mishig25:
Add 4 modalities tags (`nlp, speech, vision, rl`) (technically, I guess, it is 3 modalities + rl) & tag relevant blog posts
(as requested by @patrickvonplaten)

Screenshot 2022-02-07 at 09 41 58


_blog.yml
```diff
@@ -364,6 +394,7 @@
   date: November 15, 2021
   tags:
   - guide
+  - nlp
```
patrickvonplaten, line 397:
```suggestion
  - speech
```

_blog.yml
```diff
@@ -467,6 +506,7 @@
   date: January 21, 2022
   tags:
   -  open-source-collab
+  -  rl
 
 - local: searching-the-hub
   title: "Supercharged Searching on the Hugging Face Hub"
```
patrickvonplaten, line 512: further below at line 525 (can't comment there :-/) - could we also add the speech tag: ?

patrickvonplaten:
Awesome thanks a lot!

_blog.yml
mishig25: done

_blog.yml
```diff
@@ -483,11 +523,13 @@
   tags:
   - guide
   - research
+  - speech
```
osanseviero, line 526: cannot comment, but could we add a new line above asr-chunking ?(518)

osanseviero (approved):
Thanks for the PR!

julien-c:
can we use the same modalities names as in https://github.com/huggingface/huggingface_hub/blob/a217a7e04d4c6785e36fe193ee2576be95815157/js/src/lib/interfaces/Types.ts#L62-L67 for consistency? (and add missing ones there if needed)

More generally let's try to promote naming consistency accross the huggingface platform, going forward :)

julien-c:
do we still want to merge this BTW? 🙏

mishig25:
I've added the modalities tags to newer blog posts as well. 

> can we use the same modalities names as in https://github.com/huggingface/huggingface_hub/blob/a217a7e04d4c6785e36fe193ee2576be95815157/js/src/lib/interfaces/Types.ts#L62-L67 for consistency? (and add missing ones there if needed)

Implemented naming consistency based on below `const`.
```
export const MODALITIES = [
	"nlp",
	"audio",
	"cv",
	"other",
] as const;
```

For reinforcement learning (which does not exist in const MODALITIES), I've chose `rl`. wdyt @simoninithomas ?

julien-c:
you can maybe add `rl` to modalities now, or does it not make sense @osanseviero?

mishig25:
Here is the hub-docs [PR](https://github.com/huggingface/hub-docs/pull/90) in case we want to add _rl to modalities_

osanseviero:
I think you can merge [rl_modality](https://github.com/huggingface/hub-docs/tree/rl_modality) to this branch and then merge all into main.

PR #240, opened 2022-03-18
Optional `thumbnail_small` attribute

mishig25:
## Problem:
Our `thumbnail.png` images are pretty large, which results in a very sharp social thumbnails like this one:
Screenshot 2022-03-18 at 09 53 10

However, you can see those thumbnails loading slow on https://huggingface.co/blog because of their sizes.

## Solution:
Load smaller thumbnails images first (`thumbnail_small.png`) & then load original `thumbnail.png` imgs, which creates an effect where a user can't see img loading (because img is always there)

`thumbnail_small.png` is width 500px version of `thumbnail.png` (which is usually more than 1000px width)

Note: I've used pyyaml to update the `_blog.yml`, which ordered the attributes alphabetically, which resulted in a pretty big diff


julien-c:
what about just running the images through an image optimizer? (and remembering to do it for new images)

Or alternatively, switch the image serving to a CDN?

This PR approach seems a bit complicated to me for something that's better solved by other means (just my 2 cents)

mishig25:
I'm happy to go with either option.

For example, this thumbnail img is `475 KB` `1200 × 628`


When smaller version `500 × 262` of it becomes `86 KB`, which is `475/86 = 5.5` x smaller

I've tried using https://imageoptim.com/ for optimization, and the result is `20%` savings (which probably wont make img load faster). Therefore, CDN is better option that using image optimizer
Screenshot 2022-03-18 at 11 47 51

julien-c:
https://tinypng.com/ outputs it at `160kB` though

mishig25:
> https://tinypng.com/ outputs it at 160kB though
oh I see, unlike imageoptim, tinypng is lossy (but I guess we dont care)

I'll update the imgs with tinypng.com then 😊

mishig25:
Superceded by https://github.com/huggingface/blog/pull/241

PR #243, opened 2022-03-22
Fix links: rename master to main

julien-c:
this is not super crucial b/c GitHub redirects renamed branches, but for consistency this is nice


osanseviero (approved):
Thanks! I hope you did not do this manually

PR #246, opened 2022-03-23
Fixed broken CTA image

BritneyMuller:


feconroses (approved):

BritneyMuller (commented):
Looks good! Thank you!

julien-c:
nit, but the CTA image is a bit too small, so it's blurry on retina screens – maybe you can replace it with one w/ larger dimensions?

PR #252, opened 2022-03-28
Added computer vision study group event

merveenoyan:
Added CV study group event to _events.yml


osanseviero (approved):
Consider changing the name in case you want to have another session in the future. See how it was done for ML for Audio Study Group

PR #269, opened 2022-04-19
Fix Encoder-Decoder: Latex var index

ofikodar:
Hey, I think the target vector index should be i-1.


patrickvonplaten:
Thank you!

PR #279, opened 2022-04-29
Accelerate Large Model Training using PyTorch FSDP

pacman100:
Add blog post about PyTorch FSDP feature and its Accelerate integration


assets/62_pytorch_fsdp/complete_nlp_example.py
```diff
@@ -0,0 +1,372 @@
+# coding=utf-8
```
sgugger, line 1: Maybe we can host this example in `accelerate/examples` instead? Could go well in the [by_feature](https://github.com/huggingface/accelerate/tree/main/examples/by_feature) folder.

pytorch-fsdp.md
```diff
@@ -0,0 +1,266 @@
...
+    <a href="/sgugger">
+        <img class="avatar avatar-user" src="https://aeiljuispo.cloudimg.io/v7/https://s3.amazonaws.com/moonup/production/uploads/1593126474392-5ef50182b71947201082a4e5.jpeg?w=200&h=200&f=face" title="Gravatar">
+        <div class="bfc">
+            <code>sgugger</code>
+            <span class="fullname">Sylvain Gugger</span>
+        </div>
+    </a>
+</div>
+
+<!-- <a href="https://colab.research.google.com/github/huggingface/blog/blob/main/notebooks/13_pytorch_xla.ipynb" target="_parent"><img src="https://colab.research.google.com/assets/colab-badge.svg" alt="Open In Colab"/></a> -->
```
sgugger, line 32: Should be removed?

pytorch-fsdp.md
```diff
@@ -0,0 +1,266 @@
...
+
+# Accelerate 🚀:  Leverage PyTorch FSDP without any code changes
+
+We will look at the task of Causal Language Modelling using GPT-2 Large (762M) and XL (1.5B) model variants.
+
+Below is the code for pre-training GPT-2 model. It is similar to the official causal language modeling example [here](https://github.com/huggingface/transformers/blob/main/examples/pytorch/language-modeling/run_mlm_no_trainer.py) with the addition of 2 arguments `n_train` (2000) and `n_val` (500) to prevent preprocessing/training on entire data in order to perform quick proof of concept benchmarks.
+
+<a href="./assets/62_pytorch_fsdp/run_clm_no_trainer.py" target="_parent">run_clm_no_trainer.py</a>
```
sgugger, line 58: There is an inconsistency between the names (mlm above vs clm here)

pytorch-fsdp.md
```diff
@@ -0,0 +1,266 @@
...
+    
+    > FSDP Warning: When using FSDP, several parameter groups will be conflated into a single one due to nested module wrapping and parameter flattening.
+    > 
+    
+    This is because parameter groups created before wrapping will have no meaning post wrapping due parameter flattening of nested FSDP modules into 1D arrays (which can consume many layers). For instance, below are the named parameters of FSDP model on GPU 0 (When using 2 GPUs. Around 55M (110M/2) params in 1D arrays as this will have the 1st shard of the parameters). Here, if one has applied no weight decay for [bias, LayerNorm.weight] named parameters of unwrapped BERT-Base model, it can’t be applied to the below FSDP wrapped model as there are no named parameters with either of those strings and the parameters of those layers are concatenated with parameters of various other layers. More details mentioned in this [issue](https://github.com/pytorch/pytorch/issues/76501) (`The original model parameters' .grads are not set, meaning that they cannot be optimized separately (which is why we cannot support multiple parameter groups)`).
```
sgugger, line 210:
```suggestion
This is because parameter groups created before wrapping will have no meaning post wrapping due parameter flattening of nested FSDP modules into 1D arrays (which can consume many layers). For instance, below are the named parameters of FSDP model on GPU 0 (When using 2 GPUs. Around 55M (110M/2) params in 1D arrays as this will have the 1st shard of the parameters). Here, if one has applied no weight decay for [bias, LayerNorm.weight] named parameters of unwrapped BERT-Base model, it can’t be applied to the below FSDP wrapped model as there are no named parameters with either of those strings and the parameters of those layers are concatenated with parameters of various other layers. More details mentioned in this [issue](https://github.com/pytorch/pytorch/issues/76501) (`The original model parameters' .grads are not set, meaning that they cannot be optimized separately (which is why we cannot support multiple parameter groups)`).
```

pytorch-fsdp.md
```diff
@@ -0,0 +1,266 @@
...
+
+- Mixed precision is currently not supported with FSDP as we wait for PyTorch to fix support for it.
+
+# How it works 📝
+
+![FSDP Workflow](./assets/62_pytorch_fsdp/FSDP_workflow.png)
+
+(Source: [link](https://pytorch.org/blog/introducing-pytorch-fully-sharded-data-parallel-api/))
+
+Above workflow gives an overview of what happens behind the scenes when FSDP is activated. Let's first understand how DDP works and how FSDP improves it. In DDP, each worker/accelerator/GPU has a replica of the entire model parameters, gradients and optimizer states. Each worker gets a different batch of data, it goes through forwards pass, loss is computed followed by backward pass to generate gradients. Now, all-reduce operation is performed wherein each worker gets the gradients from the remaining workers and averaging is done. In this way, each worker now has the same global gradients which is used by optimizer to update the model parameters. We can see that having full replicas consume a lot of redundant memory on each GPU, which limits the batch size as well as loading of large models.
```
sgugger, line 231:
```suggestion
The above workflow gives an overview of what happens behind the scenes when FSDP is activated. Let's first understand how DDP works and how FSDP improves it. In DDP, each worker/accelerator/GPU has a replica of the entire model parameters, gradients and optimizer states. Each worker gets a different batch of data, it goes through the forwards pass, a loss is computed followed by the backward pass to generate gradients. Now, an all-reduce operation is performed wherein each worker gets the gradients from the remaining workers and averaging is done. In this way, each worker now has the same global gradients which are used by the optimizer to update the model parameters. We can see that having full replicas consume a lot of redundant memory on each GPU, which limits the batch size as well as the size of the models. 
```

pytorch-fsdp.md
```diff
@@ -0,0 +1,266 @@
...
... addresses this by sharding the optimizer states, gradients and model parameters across the data parallel workers. It further facilitates CPU offloading thereby enabling loading large models which won't fit the available GPU memory. Similar to DDP, each worker gets a different batch of data. During the forward pass, if the CPU offload is enabled, the parameters of the local shard are first copied to the GPU/accelerator. Then, each worker performs all-gather operation for a given FSDP wrapped module/layer(s) to get all the needed parameters, computation is performed followed by releasing/empyting the parameter shards of other workers. This continues for all the FSDP modules. Loss gets computed after the forward pass. During backward pass, again an all-gather operation is performed to get all the needed parameters for a given FSDP module, computation is performed to get local gradients followed by releasing the shards of other workers. Now, local gradients are averaged and sharded to relevant workers using reduce-scatter operation. This allows each worker to update the parameters of its local shard. If CPU offload is activated, the gradients are passed to CPU for updating parameters.
```
sgugger, line 233:
```suggestion
FSDP precisely addresses this by sharding the optimizer states, gradients and model parameters across the data parallel workers. It further facilitates CPU offloading of all those tensors, thereby enabling loading large models which won't fit the available GPU memory. Similar to DDP, each worker gets a different batch of data. During the forward pass, if the CPU offload is enabled, the parameters of the local shard are first copied to the GPU/accelerator. Then, each worker performs all-gather operation for a given FSDP wrapped module/layer(s) to all get the needed parameters, computation is performed followed by releasing/emptying the parameter shards of other workers. This continues for all the FSDP modules. The loss gets computed after the forward pass and during the backward pass, again an all-gather operation is performed to get all the needed parameters for a given FSDP module, computation is performed to get local gradients followed by releasing the shards of other workers. Now, the local gradients are averaged and sharded to each relevant workers using reduce-scatter operation. This allows each worker to update the parameters of its local shard. If CPU offload is activated, the gradients are passed to CPU for updating parameters directly on CPU. 
```

pytorch-fsdp.md
```diff
@@ -0,0 +1,266 @@
...
+
+Without reinventing the wheel, please refer [7, 8, 9] for all the in-depth details on the workings of the PyTorch FSDP and the extensive experimentation carried out using this feature.
```
sgugger, line 235:
```suggestion
Please refer [7, 8, 9] for all the in-depth details on the workings of the PyTorch FSDP and the extensive experimentation carried out using this feature.
```

pytorch-fsdp.md
```diff
@@ -0,0 +1,266 @@
...
+
+# Issues
+
+If you encounter any issues with the integration part of PyTorch FSDP, please open an Issue in [accelerate](https://github.com/huggingface/accelerate/issues).
+
+But if you have problems with PyTorch FSDP configuration, and deployment - you need to ask the experts in their domains, therefore, please, use [PyTorch Issue](https://github.com/pytorch/pytorch/issues) instead.
```
sgugger, line 241:
```suggestion
But if you have problems with PyTorch FSDP configuration, and deployment - you need to ask the experts in their domains, therefore, please, open a [PyTorch Issue](https://github.com/pytorch/pytorch/issues) instead.
```

sgugger (approved):
Thanks a lot for working on this!

@Pierrci I don't know if it's okay to have a python script in the assets? (One should go in Accelerate examples but the other one is a slight tweak of an existing Transformers example, not sure if we can have it here as an asset).

Pierrci:
> @Pierrci I don't know if it's okay to have a python script in the assets? (One should go in Accelerate examples but the other one is a slight tweak of an existing Transformers example, not sure if we can have it here as an asset).

It's definitely ok if needed (everything in `/assets` is served as static content)

PR #289, opened 2022-05-06
Clarification for the usage of  load_from_hub in sb3.md

saadaosa:
Added .zip extension to filename example in load_from_hub. This is way it is consistent with the intended usage (without the extension, we would get "HTTPError: 404 Client Error")


osanseviero (approved):
Looks good, thanks! 

cc @simoninithomas

PR #267, opened 2022-04-13
Blog post of ambassadors program

VioletteLepercq:

PR #299, opened 2022-05-10
Update ambassadors-program.md

VioletteLepercq:
following @omarespejel, I think we should open more the program and turn it into a simple "Ambassador Program". wdyt? @omarespejel @osanseviero @merveenoyan @NimaBoscarino


osanseviero:
Merging this so we review content directly in https://github.com/huggingface/blog/pull/267

PR #350, opened 2022-06-09
Improve blog

NielsRogge:
This PR fixes some typos (missing brackets), and adds a mention to tell users that Phil Wang removed ConvNeXT blocks from his repo entirely, although we obtained nice results with them.

Also fixes #354

cc @kashif @lucidrains

PR #352, opened 2022-06-09
Import sneaky LayerNorm.

kgourgou:
Adding a `nn.` in front of `LayerNorm` in the example.


NielsRogge:
Will fix this in #350

PR #362, opened 2022-06-11
fix gamma value

jianyangma:
The gamma value in text shows 0.99, but the image shows the correct value of 1

image


simoninithomas:
Awesome, thanks for pointing that out. I merge it 🤗

PR #380, opened 2022-06-22
transformers.onnx typo fix

jessecambon:
Fixed a typo (base_model -> model) so that the transformers.onnx script runs successfully


philschmid (approved):
Thank you for fixing

PR #383, opened 2022-06-23
Update blog post

abidlabs:
Noticed that a code snippet was incomplete as I was working through this blog post, so added what I think needs to be added for the code to run!

### DIFF
--- a/fine-tune-vit.md
+++ b/fine-tune-vit.md
@@ -355,7 +355,7 @@ def compute_metrics(p):
 ```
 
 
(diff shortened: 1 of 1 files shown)


fine-tune-vit.md
```diff
@@ -364,7 +364,7 @@ from transformers import ViTForImageClassification
 labels = ds['train'].features['labels'].names
 
 model = ViTForImageClassification.from_pretrained(
-    model_name_or_path,
+    "google/vit-base-patch16-224-in21k",
```
nateraw, line 367: This was defined on line 200 (`model_name_or_path`).

nateraw (commented):
I'm fine with updating this, but it was defined earlier in the post 😅

abidlabs:
Oh my bad! Somehow I didn't see that when I was running your code. Ignore this!!

PR #405, opened 2022-07-04
Add notebooks/85_sentiment_analysis_twitter.ipynb

feconroses:


stas00:
@feconroses, your assets dir points to a previous blog post's number:

```
  thumbnail: /blog/assets/85_policy_gradient/thumbnail.gif
[...]
  thumbnail: /blog/assets/85_sentiment_analysis_twitter/thumbnail.png
```

ideally it should be `86_` - but if you're going to fix it could you make it `87_` as I have already claimed `86_` in PR which will be committed shortly.

I can of course edit my PR to use 87 if you prefer 86 

Thanks.

feconroses:
@stas00 thanks for pointing this out. Policy Gradient PR came after my original PR but the post was published sooner. I can use 87, so you don't have to make changes too. I'll make this PR later today.

stas00:
ah, perfect. Thank you!

I wonder if we should switch to a different approach for assets, otherwise there is often this race condition with new articles. I'm not sure what role the numbers play. Looks more of a convention than a requirement. 

```
ls -1 *md | wc -l
```
should be enough to see how many posts we have if someone wants to know.

feconroses:
@stas00 yeah I also think it's a convention, not a requirement. Maybe I can leave it as it is? And have both blog posts as 85 (Policy Gradient PR and Sentiment Analysis Twitter). From what I'm seeing, this happened a bunch of times, and teammates just left the same numbers. Personally, I don't think it's a big issue if two blog posts posted on the same week have the same number.

stas00:
ok, then let's do nothing ;) thank you for the feedback, @feconroses

julien-c:
yes, your assessments are correct, it's just a convention to keep folders roughly sorted inside a file view, but is not used functionally in any way

it's a roughly monotonic increment

PR #407, opened 2022-07-06
The Technology Behind BLOOM

stas00:
This is an article on the engineering behind BLOOM, with the specific intention to showcase the technologies developed by the Megatron and Deepsped teams, which were the crucial part of this project.

I'm currently waiting for feedback from the Deepspeed team.

If you want to preview in the easy way go to: https://github.com/stas00/blog/blob/bloom-tech/bloom-megatron-deepspeed.md


bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,218 @@
...
+
+Parallelizing the multi-headed attention layers is even simpler, since they are already inherently parallel, due to having multiple independent heads!
+![parallel self-attention](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/parallelism-tp-parallel_self_attention.png)
+
+Special considerations: Due to the two all reduces per layer in both the forward and backward passes, TP requires a very fast interconnect between devices. Therefore it's not advisable to do TP across more than one node. Practically, if a node has 4 GPUs, the highest TP degree is therefore 4. If you need a TP degree of 8, you need to use nodes that have at least 8 GPUs.
+
+This component is implemented by Megatron-LM. Megatron-LM has recently expanded tensor parallelism to include sequence parallelism that splits the operations that cannot be split as above, such as LayerNorm, along the sequence dimension. The paper [Reducing Activation Recomputation in Large Transformer Models] (https://arxiv.org/abs/2205.05198) provides details for this technique. Sequence parallelism was developed after BLOOM was trained so not used in the BLOOM training.
```
stefan-it, line 87:
```suggestion
This component is implemented by Megatron-LM. Megatron-LM has recently expanded tensor parallelism to include sequence parallelism that splits the operations that cannot be split as above, such as LayerNorm, along the sequence dimension. The paper [Reducing Activation Recomputation in Large Transformer Models](https://arxiv.org/abs/2205.05198) provides details for this technique. Sequence parallelism was developed after BLOOM was trained so not used in the BLOOM training.
```

Otherwise the hyperlink will look like:

image

stas00: Thank you for spotting my typo, @stefan-it!

lvwerra:
Maybe @gary149 or @Reeray could help with a thumbnail and @osanseviero might know about the metadata.

_blog.yml
```diff
@@ -977,4 +977,13 @@
   tags:
     - sentiment-analysis
     - nlp
-    - guide
\ No newline at end of file
+    - guide
+
+- local: bloom-megatron-deepspeed-technology
```
osanseviero, line 982: this will be the url as well, so I would suggest to make it a bit shorter. E.g. bloom-technology for example

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,345 @@
...
+<div class="blog-metadata">
+    <small>Published July 9, 2022.</small>
+    <a target="_blank" class="btn no-underline text-sm mb-5 font-sans" href="https://github.com/huggingface/blog/blob/main/bloom-megatron-deepspeed-technology.md">
+        Update on GitHub
+    </a>
+</div>
+
+<div class="author-card">
+    <a href="/stas">
+        <img class="avatar avatar-user" src="/blog/assets/86_bloom_megatron_deepspeed_technology/stas-bekman-300x300.jpg">
```
osanseviero, line 17: You can also use https://aeiljuispo.cloudimg.io/v7/https://s3.amazonaws.com/moonup/production/uploads/1594311341799-5f07383b19cb630495b812cd.jpeg?w=200&h=200&f=face instead

osanseviero (commented):
I just looked at the metadata, it looks good :+1: !

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,345 @@
...
+
+ALIBI:
+
+-  [Train Short, Test Long: Attention with Linear Biases Enables Input Length Extrapolation](https://arxiv.org/abs/2108.12409)
+
+BitsNBytes:
+
+- [8-bit Optimizers via Block-wise Quantization](https://arxiv.org/abs/2110.02861) (in the context of Embedding LayerNorm but the rest of the paper and the technology is amazing - the only reason were weren't using the 8-bit optimizer is because we were already saving the optimizer memory with DeepSpeed-ZeRO).
+
+## Training Difficulties
```
lvwerra, line 329: I think a little transition here would be nice. Until here all the techniques were described and now we actually start training. 

> With the architecture, hardware and software in place we were able to start training on XXX. However it was not just smooth sailing from there. In this section we discuss some of the main hurdles we encountered.

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,345 @@
...
+
+The main type of issue encountered during training were hardware failures. As this was a new cluster with about 400 GPUs, on average we were getting 1-2 GPU failures a week. We were saving a checkpoint every 3h (100 iterations) so on average we would lose 1.5h of training on hardware crash. The JeanZay sysadmins would then replace the faulty GPUs and bring the node back up. Meanwhile we had backup nodes to use instead.
```
lvwerra, line 331: You mentioned at some point that the time zones between the collaborators/JeanZay were also an issue - I think that would be an interesting insight most people don't think about. Essentially that for that scale you need somebody on call. You could maybe even add a $/h estimate to indicate how expensive 1h downtime is (roughly).

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,345 @@
...
+
+## Conclusion
+
+The most difficult and intense part of the training was the 2 months leading to the start of the training. We were under a lot of pressure to start training ASAP, since the resources allocation was limited in time and we didn't have access to A100s until the very last moment. So it was a very difficult time, considering that the `BF16Optimizer` was written in the last moment and we needed to debug it and fix various bugs.
+
+Additionally we found several issues that manifested themselves only once we started training on 48 nodes, and won't appear at small scale.
+
+But once we sorted those out, the training itself was surprisingly smooth and without major problems. Most of the time we had one person monitoring the training and only a few times several people were involved to troubleshoot. We enjoyed great support from JeanZay's administration who quickly addressing most needs that emerged during the training.
+
+Overall it was a super-intense but very rewarding experience.
```
lvwerra, line 345: What do you think about adding something like the following to conclude the proposed intro:

> Training large language models is still a challenging task, but we hope by building and sharing this technology in the open others can build on top of our experience.

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,345 @@
...
+5. The PyTorch team who created a super powerful framework, on which the rest of the software was based, and who were very supportive to us during the preparation for the training, fixing multiple bugs and improving the usability of the PyTorch components we relied on during the training.
+6. The volunteers in the BigScience Engineering workgroup
+
+It'd be very difficult to name all the amazing people who contributed to the engineering side of the project, so I will just name a few key people outside of Hugging Face who were the engineering foundation of this project for the last 14 months:
+
+Olatunji Ruwase, Deepak Narayanan, Jeff Rasley, Jared Casper and Rémi Lacroix
+
+Also we are grateful to all the companies who allowed their employees to contribute to this project.
+
+## Overview
```
lvwerra, line 55: I would add a very condensed version of the Overview to the intro (between the two sections proposed above):

> Here's a summary of project:
 |            |  | 
 | :----                 | :----       |
 | Architecture            |[GPT3](https://en.wikipedia.org/wiki/GPT-3)  |
 | Dataset | 46 Languages in 1.5TB of deduplicated massively cleaned up text, converted into 350B unique tokens, see [The BigScience Corpus A 1.6TB Composite Multilingual Dataset](https://openreview.net/forum?id=UoEw6KigkUn)
 | Hardware    | 48 Nodes with 8 A100 GPUs (384 GPUs total)  |
 |Software| Megatron-LM + DeepSpeed |
 |Training time | 3.5 months |

Then I would remove the overview and instead add a `Architecture` and `Hardware` section.

- `Architecture`: here I would gather all the architecture related information incl. the Embedding LayerNorm and Positional Encoding section as well as the specs described in the overview. I think logically they should come first (this is what we want to train) and then the megatron/deepspeed discussion (that's how we do it)
- `Hardware`: here I would add the hardware info from the overview. I think what's missing is a link to the institute and the name of the cluster. I think a few sentences in addition to the specs would be nice.

lvwerra (commented):
Hi @stas00 

The article looks great overall. I mostly added comments of two categories: move a few things around and maybe add a little story component around the article (see intro and difficulties sections).

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,345 @@
...
+Besides other improvements we believe that using BF16 mixed precision training turned a potential nightmare into a relatively smooth process which can be observed from the following lm loss graph:
+
+![176B-fail](assets/86_bloom_megatron_deepspeed_technology/176b-lm-loss.png)
+
+## Fused CUDA Kernels
+
+To train BLOOM fast and efficiently it was necessary to use several custom fused CUDA kernels provided by Megatron-LM. In particular there is an optimized kernel to perform LayerNorm as well as kernels to fuse various combinations of the scaling, masking, and softmax operations. The addition of a bias term is also fused with the GeLU operation using PyTorch's JIT functionality. These operations are all memory bound, so it is important to fuse them to maximize the amount of computation done once a value has been retrieved from memory. So, for example, adding the bias term while already doing the memory bound GeLU operation adds no additional time. These kernels are all available in the [Megatron-LM repository](https://github.com/NVIDIA/Megatron-LM).
+
+
+## Datasets
```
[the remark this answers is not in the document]
stas00: The way I laid it out is that the most important crucial feature of 3D parallelism is explained first - and it takes roughly half the article. 

Then I go into smaller aspects, in no particular order.

Same with the People section, the article is ordered by the significance, rather than by functionality. As many readers will never finish reading this both technical and long article. So trying to put the important things first.

So if only part of the article was read, I want readers to at least join in celebration of the people as technology, then 3D parallelism, and if they have brain power left then everything else.

bloom-megatron-deepspeed-technology.md
stas00: An excellent suggestion, Leandro!

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,345 @@
...
+        <div class="bfc">
+            <code>stas</code>
+            <span class="fullname">Stas Bekman</span>
+        </div>
+    </a>
+</div>
+
+This article will start by thanking companies and key people and groups that made the amazing feat of training a 176 Billion parameter model by a small group of dedicated people possible.
+
+Then the hardware setup and main technological components will be discussed.
```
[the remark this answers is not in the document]
stas00: Wonderful!

bloom-megatron-deepspeed-technology.md
stas00: > You mentioned at some point that the time zones between the collaborators/JeanZay were also an issue - I think that would be an interesting insight most people don't think about.

Did we? I personally have more of an issue in general collaborating with HF work as most employees are far away from where I live timezone-wise. With BLOOM the key players - Megatron and Deepspeed teams are in my timezone so it was perfect.

And having different timezones was actually a boon and not a difficulty since others could support me while I was asleep.

But yes, it'd be good to write out that someone needs to watch the training 24/7 - will add that.
 
> You could maybe even add a $/h estimate to indicate how expensive 1h downtime is (roughly).
 
OK, so we sorted out - it's about 393€/h - but since we weren't paying for it what would be the point of saying that? in other words I'm not sure how to say it in this situation.

bloom-megatron-deepspeed-technology.md
stas00: > Summary 

excellent idea - but will use a much shorter version - all that long text is not suitable for a table summary.

> ... I think logically they should come first (this is what we want to train) and then the megatron/deepspeed discussion (that's how we do it)

I think some reviewers are unaware of the intention of this article. It was conceived to showcase Megatron and Deepspeed technologies as the titans this project was standing on. I was initially hoping that each will write an article showcasing their technology in the context of BLOOM, but we talked about it many times and it didn't quite happen. So I'm writing it instead.

> .I think what's missing is a link to the institute and the name of the cluster. 

Excellent point. I will query on slack what to write.

_blog.yml
julien-c: i also favor shorter URLs (including for SEO) and that's consistent with (some/most) other blogposts

stas00: sure, then let's change it when the reviews are over - otherwise we will make it unnecessarily hard for reviewers with the file rename.

So this article is primarily about megatron and deepspeed, so drop technology and leave 

/bloom-megatron-deepspeed

is that short enough?

julien-c: works for me:)

lvwerra (commented):
Looks really good to me! I added a few minor comments.

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,375 @@
...
+
+When ZeRO-DP is combined with PP (and optionally TP) it typically enables only ZeRO stage 1, which shards only optimizer states. ZeRO stage 2 additionally shards gradients, and stage 3 also shards the model weights.
+
+While it's theoretically possible to use ZeRO stage 2 with Pipeline Parallelism, it will have bad performance impacts. There would need to be an additional reduce-scatter collective for every micro-batch to aggregate the gradients before sharding, which adds a potentially significant communication overhead. By nature of Pipeline Parallelism, small micro-batches are used and instead the focus is on trying to balance arithmetic intensity (micro-batch size) with minimizing the Pipeline bubble (number of micro-batches). Therefore those communication costs are going to hurt.
+
+In addition, there are already fewer layers than normal due to PP and so the memory savings won't be huge. PP already reduces gradient size by ``1/PP``, and so gradient sharding savings on top of that are less significant than pure DP.
+
+ZeRO stage 3 is not a good choice either for the same reason - more inter-node communications are required.
```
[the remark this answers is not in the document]
stas00: It's complicated. BF16Optimizer didn't quite fit any stages - it's sort of a variation on stage 1, but was different, so in the config it was actually stage 0. But it's a transient state - the deepspeed team needs to sort out which stage it belongs to.

There is another BF16/stage 1, but it does accumulation in bf16 and isn't really suitable for training.

So the article makes it clear that 3D parallelism uses stage 1 and since we use 3D parallelism, we are doing that (only optimizer sharding)

@tjruwase, would you like to address this? that is how to best describe which stage `BF16Optimizer` belongs to for the sake of the article? Thanks.

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,375 @@
...
+All PyTorch components have been updated to ensure that they perform any accumulation in FP32, so no loss happening there.
+
+One crucial issue is gradient accumulation, and it's one of the main features of pipeline parallelism as the gradients from each microbatch processing get accumulated. It's crucial to implement gradient accumulation in FP32 to keep the training precise, and this is what `BF16Optimizer` does.
+
+Besides other improvements we believe that using BF16 mixed precision training turned a potential nightmare into a relatively smooth process which can be observed from the following lm loss graph:
+
+![176B-fail](assets/86_bloom_megatron_deepspeed_technology/176b-lm-loss.png)
+
+## Fused CUDA Kernels
+
```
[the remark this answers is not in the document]
stas00: That's an excellent suggestion. Will add one now.

bloom-megatron-deepspeed-technology.md
```diff
@@ -288,6 +288,10 @@ Besides other improvements we believe that using BF16 mixed precision training t
 
 ## Fused CUDA Kernels
 
+When executing `c=a+b; e=c+d`, first the `a+b` is executed - the GPU fetches `a` and `b` from GPU memory, computes the sum (`c`) and deposits it back to memory. Then to execute the next operation `c+d`, `c` and `d` are fetched from the memory, the sum is computed and the result is deposited in the memory. As you can see `c` travels to/from memory twice. If we were to fuse these two operations, we won't send the intermediary result `c` back to the memory, but leave it in the GPU registers and only need to fetch `d` to complete the computation. This saves a lot of overhead and prevents GPU idling and making the whole operation much more efficient.
+
+Fused kernels are just that. They replace multiple discrete computations and data movements to/from memory into fused computations that have very few memory movements.
+
```
stas00, line 294: @jaredcasper, could you please have a quick peek if I explained fused kernels correctly?

Thanks a lot!

jaredcasper: I think it might be helpful to include the idea of compute kernels. If the code "c=a+b; e=c+d" is in one kernel then it won't write then read c, but the compiler will do as expected. So the first two sentences could be confusing without the added detail that it is describing what happens when you put them into two kernels. Maybe something like:

When instructing the GPU to compute `c=a+b; e=c+d`, a naive approach, and what PyTorch will do unless instructed otherwise, is to launch two separate "add" kernels, one to perform `c=a+b` and one to perform `e=c+d`. In this case, the GPU fetches `a` and `b` <...> If we were to fuse these two operations, i.e. put them into a single "fused kernel", and just launch that one kernel we won't send the intermediary ...

This draws the distinction and further explains the actual mechanics of what it means to "fuse" them. It does assume the reader knows what a "kernel" is and what it means to "launch" it, so if necessary we might need to explain those, but I think an understanding of those is necessary to understand what a fused kernel is anyway.

 First paragraph last sentence I think you want to remove the "and" to make it "... idling, making the whole" or "idling and makes".

stas00: wonderful! thank you, Jared. Indeed, I didn't think of compilers ;)

I integrated your suggestions and extended the explanation to include Kernels

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+Now, when instructing the GPU to compute `c=a+b; e=c+d`, a naive approach, and what PyTorch will do unless instructed otherwise, is to launch two separate "add" kernels, one to perform `c=a+b` and another to perform `e=c+d`. In this case, the GPU fetches from its memory `a` and `b`, performs the addition, and then copies the result back into the memory. It then fetches `c` and `d` and performs the same thing.
+
+If we were to fuse these two operations, i.e. put them into a single "fused kernel", and just launch that one kernel we won't copy the intermediary result `c` to the memory, but leave it in the GPU registers and only need to fetch `d` to complete the computation. This saves a lot of overhead and prevents GPU idling and makes the whole operation much more efficient.
+
+Fused kernels are just that. Primarily they replace multiple discrete computations and data movements to/from memory into fused computations that have very few memory movements. Additionally, some fused kernels rewrite the math so that certain groups of computations can be performed faster.
```
stas00, line 299: @jaredcasper, here is the rewritten version.

jaredcasper: Looks good to me!

stas00: Thank you, Jared.

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+
+If we look at the computation in matrix form, it's easy to see how the matrix multiplication can be split between multiple GPUs:
+![Parallel GEMM](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/parallelism-tp-parallel_gemm.png)
+
+If we split the weight matrix `A` column-wise across `N` GPUs and perform matrix multiplications `XA_1` through `XA_n` in parallel, then we will end up with `N` output vectors `Y_1, Y_2, ..., Y_n` which can be fed into `GeLU` independently:
+![independent GeLU](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/parallelism-tp-independent-gelu.png). Notice with the Y matrix split along the columns, we can split the second GEMM along its rows so that it takes the output of the GeLU directly without any extra communication.
+
+Using this principle, we can update an MLP of arbitrary depth, without the need for any synchronization between GPUs until the very end, when we need to reconstruct the output vector from shards using an all reduce. The Megatron-LM paper authors provide a helpful illustration for that:
```
stas00, line 161: @jaredcasper, we are discussing this with Julien Launay and he suggests that this invalid as it can only go into a depth of 2 and not arbitrary. As this was taken from the paper, could you please have another peek and see if we can rewrite this a better way.

perhaps I got it wrong when reading the paper. My understanding is that the block can have many sequential ops and thus arbitrary depth, but clearly it is ambiguous to some.

thank you!

stas00: https://arxiv.org/abs/2104.04473 2.3 Tensor Model Parallelism

It looks like I may have hallucinated that explanation as I don't see it in the paper.


2.3     Tensor Model Parallelism
With tensor model parallelism, individual layers of the model are
partitioned over multiple devices. In this paper, we use the particular
partitioning strategy used by Megatron [33] for transformer layers
that are the bedrock of language models. We can apply similar ideas
to other types of models, like CNNs, as well. We briefly outline this
strategy below.
   A transformer layer consists of a self-attention block followed
by a two-layer multi-layer perceptron (MLP). Further details of the
transformer layer can be found in Vaswani et al [35].
   The MLP block consists of two GEMMs with a GeLU non-linearity
in the middle:
                𝑌 = GeLU(𝑋𝐴) 𝑍 = Dropout(𝑌 𝐵)
We can split 𝐴 along its columns 𝐴 = [𝐴1, 𝐴2 ]. This partitioning
allows the GeLU non-linearity to be independently applied to the
output of each partitioned GEMM:
               [𝑌1, 𝑌2 ] = [GeLU(𝑋𝐴1 ), GeLU(𝑋𝐴2 )]
This is advantageous as it removes the need for synchronization
(needed if 𝐴 is split along its rows since GeLU is non-linear).
   The rows of the second weight matrix 𝐵 can then be split along
its rows to remove the need for any communication between the
GEMMs (shown in Figure 5a), as shown below:
                             
                             𝐵1
                        𝐵=        , 𝑌 = [𝑌1, 𝑌2 ].
                             𝐵2
The output of the second GEMM is then reduced across the GPUs
before the dropout layer.
   We exploit the inherent parallelism in the multi-head attention
operation to partition the self-attention block (shown in Figure 5b).
The key (𝐾), query (𝑄), and value (𝑉 ) matrices can be partitioned in
a column-parallel fashion. The output linear layer can then directly
operate on the partitioned output of the attention operation (weight
matrix partitioned across rows).
   This approach splits GEMMs in the MLP and self-attention blocks
across GPUs while requiring only two all-reduce operations in the
forward pass (𝑔 operator) and two all-reduces in the backward pass
(𝑓 operator). We implemented 𝑓 and 𝑔 in a few lines of code.

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+## Megatron-DeepSpeed
+
+The 176B BLOOM model has been trained using [Megatron-DeepSpeed](https://github.com/bigscience-workshop/Megatron-DeepSpeed), which is a combination of 2 main technologies:
+
+* [Megatron-LM](https://github.com/NVIDIA/Megatron-LM) is a large, powerful transformer developed by the Applied Deep Learning Research team at NVIDIA.
+* [DeepSpeed](https://github.com/microsoft/DeepSpeed) is a deep learning optimization library that makes distributed training easy, efficient, and effective.
+
+The DeepSpeed team extended the Megatron-LM framework to support its ZeRO sharding protocol and their Pipeline Parallelism implementation. Thus Megatron-DeepSpeed was born.
+
+Please note that the BigScience's [Megatron-DeepSpeed](https://github.com/bigscience-workshop/Megatron-DeepSpeed) is a fork of the original [Megatron-DeepSpeed](https://github.com/microsoft/Megatron-DeepSpeed) repository, to which we added multiple additions.
```
lvwerra, line 105: What do you think about adding a few highlights?
```suggestion
Please note that the BigScience's [Megatron-DeepSpeed](https://github.com/bigscience-workshop/Megatron-DeepSpeed) is a fork of the original [Megatron-DeepSpeed](https://github.com/microsoft/Megatron-DeepSpeed) repository, to which we added multiple additions such as ...
```

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+        <div class="bfc">
+            <code>stas</code>
+            <span class="fullname">Stas Bekman</span>
+        </div>
+    </a>
+</div>
+
+
+
+In recent years, training ever larger language models has become the norm. While the issues of those models' not being released for further study is frequently discussed, the hidden knowledge about how to train such models rarely gets any attention. This article aims to change this by shedding some light on the technology and engineering behind training such models both in terms of hardware and software on the the example of the 176B parameter language model BLOOM.
```
douwekiela, line 27: "the the" example

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+
+## People
+
+The project was conceived by Thomas Wolf, who dared to compete with the huge corporations not only to train one of the largest multilingual models, but also to make the final result accessible to all people, thus making what was but a dream to most people a reality.
+
+This article focuses specifically on the engineering side of the training of the model. The most important part of the technology behind BLOOM were the people and companies who shared their expertise and helped us with coding and training.
+
+There are 6 main groups of people to thank:
+
+1. The HuggingFace's BigScience team who dedicated more than half a dozen full time employees to figure out and run the training from inception to the finishing line and provided and paid for all the infrastructure beyond the JeanZay's compute.
```
douwekiela, line 53: I'd remove the first "The"

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+2. The Microsoft DeepSpeed team, who developed DeepSpeed and later integrated it with Megatron-LM, and whose developers spent many weeks working on the needs of the project and provided lots of awesome practical experiential advice before and during the training.
+3. The NVIDIA Megatron-LM team, who developed Megatron-LM and who were super-helpful answering our numerous questions and providing first class experiential advice.
```
douwekiela, line 55: super helpful (remove dash)

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+4. The IDRIS / GENCI team managing the JeanZay supercomputer, who donated to the project an insane amount of compute and a great system administration support.
```
douwekiela, line 56: Insert space on Jean Zay?

douwekiela: "and great system administration support" (remove 'a')

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+| Tensor Parallelism    | V           |           |
+| Pipeline Parallelism  |             | V         |
+| ZeRO Data Parallelism |             | V         |
+| BF16Optimizer         |             | V         |
+
+Please note that both Megatron-LM and DeepSpeed have a Pipeline Parallelism implementation, but we used the one from DeepSpeed as it's integrated with ZeRO.
+
+Megatron-DeepSpeed implements 3D Parallelism to allow huge models to train in a very efficient way. Let’s briefly discuss the 3D components.
+
+1. **DataParallel (DP)** - the same setup is replicated multiple times, and each being fed a slice of the data. The processing is done in parallel and all setups are synchronized at the end of each training step.
```
douwekiela, line 122: Can we be consistent with capitalization of these sections :) "Also" below is upper case the rest is lower case.

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+2. **TensorParallel (TP)** - each tensor is split up into multiple chunks, so instead of having the whole tensor reside on a single GPU, each shard of the tensor resides on its designated GPU. During processing each shard gets processed separately and in parallel on different GPUs and the results are synced at the end of the step. This is what one may call horizontal parallelism, as the splitting happens on horizontal level.
```
douwekiela, line 123: on _a_ horizontal level

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+
+## Tensor Parallelism
+
+In Tensor Parallelism (TP) each GPU processes only a slice of a tensor and only aggregates the full tensor for operations that require the whole thing.
+
+In this section we use concepts and diagrams from the [Megatron-LM](https://github.com/NVIDIA/Megatron-LM) paper: [Efficient Large-Scale Language Model Training on GPU Clusters](https://arxiv.org/abs/2104.04473).
+
+The main building block of any transformer is a fully connected `nn.Linear` followed by a nonlinear activation `GeLU`.
+
+Following the Megatron's paper notation, we can write the dot-product part of it as `Y = GeLU(XA)`, where `X` and `Y` are the input and output vectors, and `A` is the weight matrix.
```
douwekiela, line 153: Following the Megatron paper's notation

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+This insight came from experimenting with [bitsandbytes](https://github.com/facebookresearch/bitsandbytes) which contains a `StableEmbedding` which is a normal Embedding with layernorm and it uses a uniform xavier initialization.
+
+## Positional Encoding
+
+We also replaced the usual positional embedding with an AliBi - based on the paper: [Train Short, Test Long: Attention with Linear Biases Enables Input Length Extrapolation](https://arxiv.org/abs/2108.12409), which allows to extrapolate for longer input sequences than the ones the model was trained on. So even though we train on sequences with length 2048 the model can also deal with much longer sequences during inference.
+
+
+## Training Difficulties
+
+With the architecture, hardware and software in place we were able to start training in the early March 2022. However, it was not just smooth sailing from there. In this section we discuss some of the main hurdles we encountered.
```
douwekiela, line 321: in early March

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+
+There were a lot of issues to figure out before the training started. In particular we found several issues that manifested themselves only once we started training on 48 nodes, and won't appear at small scale. e.g. `CUDA_LAUNCH_BLOCKING=1` was needed to prevent the framework from hanging, and we needed to split the optimizer groups into smaller groups, otherwise the framework would again hang. You can read about those in detail in the [training prequel chronicles](https://github.com/bigscience-workshop/bigscience/blob/master/train/tr11-176B-ml/chronicles-prequel.md).
```
douwekiela, line 323: E.g.,

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+
+
+## Conclusion
+
+The most difficult and intense part of the training was the 2 months leading to the start of the training. We were under a lot of pressure to start training ASAP, since the resources allocation was limited in time and we didn't have access to A100s until the very last moment. So it was a very difficult time, considering that the `BF16Optimizer` was written in the last moment and we needed to debug it and fix various bugs. And as explained in the previous section we discovered new problems that that manifested themselves only once we started training on 48 nodes, and won't appear at small scale.
+
+But once we sorted those out, the training itself was surprisingly smooth and without major problems. Most of the time we had one person monitoring the training and only a few times several people were involved to troubleshoot. We enjoyed great support from JeanZay's administration who quickly addressing most needs that emerged during the training.
```
douwekiela, line 340: Jean Zay (insert whitespace)

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+
+- [main training document](https://github.com/bigscience-workshop/bigscience/blob/master/train/tr11-176B-ml/README.md)
+- [tensorboard](https://huggingface.co/bigscience/tr11-176B-ml-logs/tensorboard)
+- [training slurm script](https://github.com/bigscience-workshop/bigscience/blob/master/train/tr11-176B-ml/tr11-176B-ml.slurm)
+- [training chronicles](https://github.com/bigscience-workshop/bigscience/blob/master/train/tr11-176B-ml/chronicles.md)
+
+
+### Papers and Articles
+
+We couldn't have possibly explained everything in details in this article, so if the technology presented here piqued your curiosity and you'd like to know more here are the papers to read:
```
douwekiela, line 358: in detail

douwekiela (commented):
This is awesome! Some minor nits

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+
+Regardless of whether one uses BF16 or FP16 there is also a copy of weights which is always in FP32 - this is what gets updated by the optimizer. So the 16-bit formats are only used for the computation, the optimizer updates the FP32 weights with full precision and then casts them into the 16-bit format for the next iteration.
+
+All PyTorch components have been updated to ensure that they perform any accumulation in FP32, so no loss happening there.
+
+One crucial issue is gradient accumulation, and it's one of the main features of pipeline parallelism as the gradients from each microbatch processing get accumulated. It's crucial to implement gradient accumulation in FP32 to keep the training precise, and this is what `BF16Optimizer` does.
+
+Besides other improvements we believe that using BF16 mixed precision training turned a potential nightmare into a relatively smooth process which can be observed from the following lm loss graph:
+
+![176B-fail](assets/86_bloom_megatron_deepspeed_technology/176b-lm-loss.png)
```
lvwerra, line 287: are there any guesses what the spike was? I think this is a natural question readers would ask themselves.

bloom-megatron-deepspeed-technology.md
osanseviero: I don't think so (cc @julien-c). The image is small so I think it's good!

bloom-megatron-deepspeed-technology.md
```diff
@@ -290,6 +288,16 @@ Besides other improvements we believe that using BF16 mixed precision training t
...
+
+Now, when instructing the GPU to compute `c=a+b; e=c+d`, a naive approach, and what PyTorch will do unless instructed otherwise, is to launch two separate "add" kernels, one to perform `c=a+b` and another to perform `e=c+d`. In this case, the GPU fetches from its memory `a` and `b`, performs the addition, and then copies the result back into the memory. It then fetches `c` and `d` and performs the same thing.
+
+If we were to fuse these two operations, i.e. put them into a single "fused kernel", and just launch that one kernel we won't copy the intermediary result `c` to the memory, but leave it in the GPU registers and only need to fetch `d` to complete the computation. This saves a lot of overhead and prevents GPU idling and makes the whole operation much more efficient.
```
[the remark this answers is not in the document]
jaredcasper: The key part is going from two kernels to one kernel. If the additions are in a single kernel `c=a+b; e=c+d` and `e=a+b+d` are essentially equivalent and imo saying that fusing is going from one to the other just makes it more confusing and detracts from the key part of getting them into the same kernel.

bloom-megatron-deepspeed-technology.md
jaredcasper: Oy, missed this in my pass earlier. :) Yes, you would need communication between every two layers to extend the MLP, so you can do column, then row, then communication, column, row, comm, etc...

stas00: ok, I have just changed it to a shorter:

> Using this principle, we can update an MLP of arbitrary depth, while synchronizing the GPUs after each row-column sequence.

bloom-megatron-deepspeed-technology.md
```diff
@@ -0,0 +1,383 @@
...
+| :-----        | :-------------              |
+| Hardware      | 384 80GB A100 GPUs          |
+| Software      | Megatron-DeepSpeed          |
+| Architecture  | GPT3 w/ extras              |
+| Dataset       | 350B tokens of 46 Languages |
+| Training time | 3.5 months                  |
+
+## People
+
+The project was conceived by Thomas Wolf, who dared to compete with the huge corporations not only to train one of the largest multilingual models, but also to make the final result accessible to all people, thus making what was but a dream to most people a reality.
```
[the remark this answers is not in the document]
stas00: may be ask Thom? But normally you'd say "CSO of Hugging Face" I think

bloom-megatron-deepspeed-technology.md
```diff
@@ -290,6 +288,16 @@ Besides other improvements we believe that using BF16 mixed precision training t
 
 ## Fused CUDA Kernels
 
+The GPU performs two things. It can copy data to/from memory and perform computations on that data. While the GPU is busy copying the GPU's computations units idle. If we want to efficiently utilize the GPU we want to minimize the idle time.
+
+A kernel is a set of instructions that implements a specific PyTorch operation. For example, when you call `torch.add`, it goes through a [PyTorch dispatcher](http://blog.ezyang.com/2020/09/lets-talk-about-the-pytorch-dispatcher/) which looks at the input tensor(s) and various other things and decides which code it should run, and then runs it. A CUDA kernel is a specific implementation that uses the CUDA API library and can only run on NVIDIA GPUs.
+
+Now, when instructing the GPU to compute `c=a+b; e=c+d`, a naive approach, and what PyTorch will do unless instructed otherwise, is to launch two separate "add" kernels, one to perform `c=a+b` and another to perform `e=c+d`. In this case, the GPU fetches from its memory `a` and `b`, performs the addition, and then copies the result back into the memory. It then fetches `c` and `d` and performs the same thing.
```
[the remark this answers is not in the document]
stas00: I was trying to stress that it performs the exact same sequence of operations here. another addition doesn't necessarily mean it's exactly the same sequence, does it?

bloom-megatron-deepspeed-technology.md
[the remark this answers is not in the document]
stas00: what Jared said.

It's really:

```
c = op1('a', 'b')
e = op2('c','d')
```
so you can't combine these as with addition into a single code, since `e` depends on `c` .

perhaps we should use a different example?

bloom-megatron-deepspeed-technology.md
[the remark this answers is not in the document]
stas00: not " at "

bloom-megatron-deepspeed-technology.md
stas00: where? All `The`'s each item of 6?

You click on "Add suggestion" button and edit the text how it should be and hit submit ;)

snapshot:
snapshot_125

bloom-megatron-deepspeed-technology.md
stas00: Honestly nobody seems to know. At least based on my asking the experts. There all kinds of hypotheses and sometimes some of them are that, but most of the time it's a mystery.

Here are some examples: If you somehow messed up the weights you get a spike, if you have a bug in your code that impacts the optimizer - often you get a spike, when there is a huge gradient spike, you always get a matching loss spike,

We started investigating and writing tools to get the answers, but the needs of the project took over and we weren't able to finish those. we developed the first stages of ModelInspector https://github.com/bigscience-workshop/Megatron-DeepSpeed/pull/155 but haven't had a chance to finish it.

bloom-megatron-deepspeed-technology.md
stas00: I won't even know where to start - we have 200 merged PRs. and the Meg-DS development moved to our fork, so I'm not sure how to address this.

Bottom-line our Meg-DS fork is currently the most advanced version, and of course many contributions came from Deepspeed - it was just simpler to work directly on the fork that is being used rather than constantly try to sync the branches, which is too error-prone.

bloom-megatron-deepspeed-technology.md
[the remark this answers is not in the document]
stas00: I agree with the general proposal, just want to ask for clarification - isn't research community "public" too?

May I suggest to just add "to the public"?

bloom-megatron-deepspeed-technology.md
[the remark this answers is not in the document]
stas00: ok, I rewrote the whole example to use 2 different operators, so hopefully now it's more clear. 

Please let me know if it's better now and doesn't look like the obvious solution should be done by the compiler rather than fused kernel.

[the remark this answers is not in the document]
stas00: Here it is for your convenience:

> Now, when instructing the GPU to compute `c = torch.add(a, b); e = torch.max([c,d])`, a naive approach, and what PyTorch will do unless instructed otherwise, is to launch two separate kernels, one to perform the addition of `a` and `b` and another to find the maximum value between `c` and `d`. In this case, the GPU fetches from its memory `a` and `b`, performs the addition, and then copies the result back into the memory. It then fetches `c` and `d` and performs the `max` operation and again copies the result back into the memory.

If we were to fuse these two operations, i.e. put them into a single "fused kernel", and just launch that one kernel we won't copy the intermediary result `c` to the memory, but leave it in the GPU registers and only need to fetch `d` to complete the last computation. This saves a lot of overhead and prevents GPU idling and makes the whole operation much more efficient.

bloom-megatron-deepspeed-technology.md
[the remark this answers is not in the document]
stas00: I rewrote this example, so it's no longer relevant.

bloom-megatron-deepspeed-technology.md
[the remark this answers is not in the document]
lvwerra: ok sounds good, i really don't know anything about kernels :)

bloom-megatron-deepspeed-technology.md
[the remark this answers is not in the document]
lvwerra: What do you think @thomwolf?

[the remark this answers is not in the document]
stas00: perfect, thank you, Thom

bloom-megatron-deepspeed.md
```diff
@@ -0,0 +1,399 @@
...
+| Component                                       | DeepSpeed | Megatron-LM |
+| :----                                           | :----     | :----       |
+| [ZeRO Data Parallelism](#zero-data-parallelism) | V         |             |
+| [Tensor Parallelism](#tensor-parallelism)       |           | V           |
+| [Pipeline Parallelism](#pipeline-parallelism)   | V         |             |
+| [BF16Optimizer](#bf16optimizer)                 | V         |             |
+| [Fused CUDA Kernels](#fused-cuda-kernels)       |           | V           |
+| [DataLoader](#datasets)                         |           | V           |
+
+Please note that both Megatron-LM and DeepSpeed have a Pipeline Parallelism implementation, but we used the one from DeepSpeed as it's integrated with ZeRO.
```
jaredcasper, line 119: Can we mention that Megatron-LM also has a BF16Optimizer?

```suggestion
Please note that both Megatron-LM and DeepSpeed have both a Pipeline Parallelism and BF16 Optimizer implementation, but we used the one from DeepSpeed as it's integrated with ZeRO.
```

PR #415, opened 2022-07-11
Fix typo in big-bird.md

novice03:
Changes

> For calculating attention score for tokens in seconcd block

to

> For calculating attention score for tokens in second block


novice03:
Closing since change was proposed in https://github.com/huggingface/blog/pull/378.

PR #425, opened 2022-07-16
[DRL course] Add missing parenthesis

roykoand:
Hi Team,
I've noticed one obvious typo in the https://huggingface.co/blog/deep-rl-q-part1:
image
Not sure if this worth the PR. Please refer me to the place where we can report things like this in that case.
Thanks a lot for the course! Really awesome!


julien-c:
This looks good to me, thanks a ton! cc @simoninithomas for visibility

PR #430, opened 2022-07-20
Fix syntax highlighting in mnist-adversarial

julien-c:
cc @chrisemezue


chrisemezue:
thank you so much @julien-c

PR #432, opened 2022-07-20
Efficient BLOOM Inference on PyTorch

stas00:
WIP: Efficient BLOOM Inference on PyTorch


bloom-inference-pytorch.md
```diff
@@ -0,0 +1,173 @@
...
+        <div class="bfc">
+            <code>narsil</code>
+            <span class="fullname">Nicolas Patry</span>
+        </div>
+    </a>
+</div>
+
+This article discusses various PyTorch-based solution to efficient 176B parameter [BLOOM model](https://huggingface.co/bigscience/bloom).
+
+As the model needs 352GB in bf16 weights (`176*2`), the most efficient set-up is 8x80GB A100. The second best is 2x8x40GB A100s. The main reason for using A100s And, is that at as of this writing they provide the largest GPU memory. But other GPUs can be used as well. It'd probably take 24x32GB V100 as another possibility.
```
sgugger, line 42:
```suggestion
As the model needs 352GB in bf16 weights (`176*2`), the most efficient set-up is 8x80GB A100. The second best is 2x8x40GB A100s. The main reason for using A100s AMD, is that at as of this writing they provide the largest GPU memory. But other GPUs can be used as well. It'd probably take 24x32GB V100 as another possibility.
```

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,173 @@
...
+
+If you don't have that much hardware, it's still possible to run BLOOM inference on smaller GPUs, by using CPU or NVME offload, but of course the generation time will be much slower.
+
+For the sake of consistency, unless stated differently, the benchmarks in this article were all done on the same 8x80GB A100 node on [Jean Zay HPC](http://www.idris.fr/eng/jean-zay/index.html). The HPC users enjoy a very fast IO of about 3GB/s read speed (GPFS). This is important for loading speed time. A slow disc will result in slow loading time. Especially since we are concurrently doing IO in multiple processes.
+
+## Performance metrics
+
+For doing occasional inference on someone's desktop, that is a few inputs at a time the two critical metrics are the time to load the framework and the checkpoint shards and then the generation time.
+
+For doing inference on the server the two important metrics for inference is the overall latency and the throughput of tokens. The loading time is less important since once loaded it becomes persistent.
```
sgugger, line 54:
```suggestion
For doing inference on a server the two important metrics are the overall latency and the throughput of tokens. The loading time is less important since once loaded, the model becomes persistent.
```

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,173 @@
...
+
+Using a single node is ideal since PCIe speed is typically much faster than inter-node network.
+
+If you don't have that much hardware, it's still possible to run BLOOM inference on smaller GPUs, by using CPU or NVME offload, but of course the generation time will be much slower.
+
+For the sake of consistency, unless stated differently, the benchmarks in this article were all done on the same 8x80GB A100 node on [Jean Zay HPC](http://www.idris.fr/eng/jean-zay/index.html). The HPC users enjoy a very fast IO of about 3GB/s read speed (GPFS). This is important for loading speed time. A slow disc will result in slow loading time. Especially since we are concurrently doing IO in multiple processes.
+
+## Performance metrics
+
+For doing occasional inference on someone's desktop, that is a few inputs at a time the two critical metrics are the time to load the framework and the checkpoint shards and then the generation time.
```
sgugger, line 52:
```suggestion
For doing occasional inference on someone's desktop, that is a few inputs at a time, the two critical metrics are the time to load the framework and the checkpoint shards and then the generation time.
```

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,173 @@
...
+
+The latency is more difficult to define. If the server is idle and the model is ready to generate, and one or more prompts are sent from the same user at once, the latency would be just the time to generate a single token multiplied by the number of new tokens to generate. However, if you have multiple users sending prompts at different times, things become much more complicated. The first user gets the fastest response and thus fast latency, but subsequent users if they are unlucky to submit their request while the previous request is being processed may have to first wait till the first request has been generated and only their request will be attended. It's possible to design a very efficient system that uses a pipeline parallelism, where a new request can join the pipeline at any moment. But this of course will make the overall throughput lower. Then of course there is a small overhead of processing the input, sending data to the server and sending the results back to the user, but this is usually insignificant compared to the cost of generation on a huge model.
```
sgugger, line 58:
```suggestion
The latency is more difficult to define. If the server is idle and the model is ready to generate, and one or more prompts are sent from the same user at once, the latency would be just the time to generate a single token multiplied by the number of new tokens to generate. However, if you have multiple users sending prompts at different times, things become much more complicated. The first user gets the fastest response and thus fast latency, but subsequent users if they are unlucky to submit their request while the previous request is being processed may have to first wait till the first request has been generated and only their request will be attended. It's possible to design a very efficient system that uses a pipeline parallelism, where a new request can join the pipeline at any moment. But this of course will make the overall throughput lower. Then there is a small overhead of processing the input, sending data to the server and sending the results back to the user, but this is usually insignificant compared to the cost of generation on a huge model.
```
(to avoid two successions of "of course")

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,173 @@
...
+
+This article primarily discusses simple solutions and new more efficient solutions are being worked on as well.
```
sgugger, line 60:
```suggestion
This article primarily discusses simple solutions in PyTorch. More efficient solutions are being worked on as well.
```

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,173 @@
...
+[...]
+
+*** Performance stats:
+Throughput per token including tokenize: 10.74 msecs
+Start to ready to generate: 107.892 secs
+Tokenize and generate 20000 (bs=40) tokens: 52.059 secs
+Start to finish: 159.952 secs
+```
+
+The highest batch size I was able to run without OOM was 40 in this case.
```
sgugger, line 93:
```suggestion
The highest batch size we were able to run without OOM was 40 in this case.
```

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,173 @@
...
+[...]
+*** Performance stats:
+Throughput per token including tokenize: 0.73 msecs
+Start to ready to generate: 591.902 secs
+Tokenize and generate 64000 (bs=128) tokens: 9.421 secs
+Start to finish: 601.323 secs
+
+```
+
+The highest batch size I was able to run without OOM was 128 in this case.
```
sgugger, line 124:
```suggestion
The highest batch size we were able to run without OOM was 128 in this case.
```

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,173 @@
...
+
+
+
+
+
+
+## DeepSpeed-ZeRO
+
+[DeepSpeed-ZeRO](https://www.deepspeed.ai/tutorials/zero/) was primarily developed for training huge models, and wasn't targeting inference, but it still produces pretty decent results. ZeRO shards the weights across multiple GPUs and then gathers the needed weights magically before each `forward` and `backward` calls, so that the model isn't aware that its weights were ever split up.
```
sgugger, line 133:
```suggestion
[DeepSpeed-ZeRO](https://www.deepspeed.ai/tutorials/zero/) was primarily developed for training huge models, and does not target inference, but it still produces pretty decent results. ZeRO shards the weights across multiple GPUs and then gathers the needed weights magically before each `forward` and `backward` calls, so that the model isn't aware that its weights were ever split up.
```

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,173 @@
...
+deepspeed --num_gpus 8 scripts/inference/bloom-ds-inference.py --name bigscience/bloom --batch_size 8 --benchmark
+[...]
+*** Performance stats:
+Throughput per token including tokenize: 37.86 msecs
+Start to ready to generate: 464.724 secs
+Tokenize and generate 32000 (bs=8) tokens: 242.090 secs
+Start to finish: 706.813 secs
+```
+
+The highest batch size I was able to run without OOM was 8 in this case.
```
sgugger, line 156:
```suggestion
The highest batch size we were able to run without OOM was 8 in this case.
```

sgugger (approved):
Thanks for drafting this! Left some nits on the text.

bloom-inference-pytorch.md
stas00: AMD? I just need to remove And ;) no?

bloom-inference-pytorch.md
stas00: you must have meant to edit this, but the change is exactly the same as the original.

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,183 @@
...
+2. Analyze the size of each layer and the available space on each device (GPUs, CPU) to decide where each layer should go.
+3. Load the model checkpoint bit by bit and put each weight on its device
+
+It then ensures the model runs properly with hooks that transfer the inputs and outputs on the right device and that the model weights offloaded on the CPU (or even the disk) are loaded on a GPU just before the forward pass, before being offloaded again once the forward pass is finished.
+
+In a situation where there are multiple GPUs with enough space to accomodate the whole model, it switches control from one GPU to the next until all layers have run. Only one GPU works at any given time, which sounds very inefficient but it does produce excellent throughput despite the idling of the GPUs.
+
+It is also very flexible since the same code can run on any given setup. Accelerate will use all available GPUs first, then offload on the CPU until the RAM is full, and finally on the disk. It will be slower than running the full model on GPUs of course. As an example, users have reported running BLOOM with no code changes on just 2 A100s with a throughput of 15s per token.
```
stas00, line 78:
```suggestion
It is also very flexible since the same code can run on any given setup. Accelerate will use all available GPUs first, then offload on the CPU until the RAM is full, and finally on the disk. Offloading to CPU and even more so to disk will make things slower. As an example, users have reported running BLOOM with no code changes on just 2 A100s with a throughput of 15s per token.
```

Trying to disambiguate what "It" refers to.

stas00: should we give a reference point here and add:

```
As an example, users have reported running BLOOM with no code changes on just 2 A100s with a throughput of 15s per token as compared to 10 msec on 8x80 A100s.
```

sgugger: SGTM!

bloom-inference-pytorch.md
sgugger: Wasn't sure what you meant here actually 😅

stas00: AMD was a superfluous word - now removed. Is this OK:

> The main reason for using A100s is that at as of this writing they provide the largest GPU memory

stas00: I think `at` is wrong too, this?

> The main reason for using A100s is that as of this writing they provide the largest GPU memory

sgugger: Perfect!

Narsil (commented):
I left a few comments.

I think for the big picture we need an extra paragraph explaining that for a live server / a few queries we're bottlenecked by memory bandwidth, but that as batch_size increases it's less and less relevant because we're more and more compute bound.

This is a very important fact to keep in mind, especially comparing throughputs like we're doing since measuring throughput when increasing batch_size is essentially measuring how well GPU can distribute across new cores (so what I am trying to say is that it's not really a fair comparison to when we're starting to be compute bound)

Since memory bandwidth is the limiting factor it should be a little more present in the message we're trying to convey IMO (because that's how DS obtains better performance, not really by using TP)

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,183 @@
...
+3. Load the model checkpoint bit by bit and put each weight on its device
+
+It then ensures the model runs properly with hooks that transfer the inputs and outputs on the right device and that the model weights offloaded on the CPU (or even the disk) are loaded on a GPU just before the forward pass, before being offloaded again once the forward pass is finished.
+
+In a situation where there are multiple GPUs with enough space to accomodate the whole model, it switches control from one GPU to the next until all layers have run. Only one GPU works at any given time, which sounds very inefficient but it does produce excellent throughput despite the idling of the GPUs.
+
+It is also very flexible since the same code can run on any given setup. Accelerate will use all available GPUs first, then offload on the CPU until the RAM is full, and finally on the disk. Offloading to CPU or disk will make things slower. As an example, users have reported running BLOOM with no code changes on just 2 A100s with a throughput of 15s per token as compared to 10 msecs on 8x80 A100s.
```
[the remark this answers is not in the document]
sgugger: Yes it was with batch size 1.

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,183 @@
...
+```
+
+The highest batch size we were able to run without OOM was 128 in this case.
+
+
+
+
+
+
+## DeepSpeed-ZeRO
```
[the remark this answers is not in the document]
stas00: Not sure, perhaps we can ask the deepspeed team and see if they want it or not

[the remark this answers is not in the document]
stas00: @jeffra, @tjruwase - what do you guys think? Should we not mentioned DS-ZeRO as an inference solution and only discuss DS-Inference? Thank you!

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Performance metrics
+
+For doing occasional inference on someone's desktop, that is a few inputs at a time, the two critical metrics are the time to load the framework and the checkpoint shards and then the generation time.
+
+For doing inference on a server the two important metrics are the overall latency and the throughput of tokens. The loading time is less important since once loaded, the model becomes persistent.
+
+The time to generate a token is straightforward - this is just how long it takes for a new token to be generated. As long as `use_cache=True` (the default argument to `generate`) all the previous logits are saved and therefore each token takes the same amount of time to generate. This of course comes at a memory cost - as the model is huge it can easily take GBs of memory to keep the cache. One can turn the caching off, and save memory, but it'd mean recalculating all the previous logits - which would make a long sequence generation very slow.
```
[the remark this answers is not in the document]
stas00: yes please - perhaps it was a very short output - otherwise it's super strange!

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,183 @@
...
+
+The latency is more difficult to define. If the server is idle and the model is ready to generate, and one or more prompts are sent from the same user at once, the latency would be just the time to generate a single token multiplied by the number of new tokens to generate. However, if you have multiple users sending prompts at different times, things become much more complicated. The first user gets the fastest response and thus fast latency, but subsequent users if they are unlucky to submit their request while the previous request is being processed may have to first wait till the first request has been generated and only their request will be attended. It's possible to design a very efficient system that uses a pipeline parallelism, where a new request can join the pipeline at any moment. But this of course will make the overall throughput lower. Then of course there is a small overhead of processing the input, sending data to the server and sending the results back to the user, but this is usually insignificant compared to the cost of generation on a huge model.
```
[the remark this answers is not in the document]
stas00: It'd be slower since 

1. in the simple case you give all gpus to a single batch.
2. with the pipeline you share gpus to process multiple batches.

So the time to process a single batch is now slower. a throughput of a single batch is slower.

However the back-to-back latency of a user entering the busy queue is faster.

Are we on the same page now?

That paragraph was there to specifically disambiguate the definition of latency in this context. So clearly it needs some work if it was unclear.

stas00:
> I think for the big picture we need an extra paragraph explaining ...

I was hoping you'd write a whole section on the server-style inference. I have even added you as an author ahead of time ;)

Perhaps this article could just discuss the server inference, summarizing your discoveries, but not bothering with the code and perhaps another article down the road with details of the code if you'd be inspired to do so?

Because, otherwise this article so far focuses on fast inference and the server solution is a related but a very different problem to solve,  but the token generation approach is the same w/ or w/o the server, IMHO.

bloom-inference-pytorch.md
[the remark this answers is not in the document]
tjruwase: @stas00, I am fine with dropping DS-ZeRO because when there are enough GPUs to fit the model, DS-Inference is DS recommendation. We find that DS-ZeRO is more suitable when there is insufficient GPUs to fit the model and offloading to CPU or NVMe is necessary.   Since the blog focuses on non-offload scenarios, I think including just one (the best) DS solution for such scenarios enhances clarity and avoids reader confusion. Just my two cents.

bloom-inference-pytorch.md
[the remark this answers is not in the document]
Narsil: > It'd be slower since
> 
>  1. in the simple case you give all gpus to a single batch.
>  2. with the pipeline you share gpus to process multiple batches.

I think I need to devise drawings since in both cases you send the batch to all GPUs (doesn't matter how you share the weights/computation across). The only thing that matters, is the overhead added by having another user on the same machine as you. And if a user blocks the machine for a single `forward` loop, or if it blocks for the full `generate` loop.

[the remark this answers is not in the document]
stas00: The logic is quite simple - the gpus will process a BS=1 and BS=64 at about the same speed. So if you can't fit inputs into large batches the processing time of sequential batches that will start after the first batch will lead to a slower througput no matter how you cut it.

bloom-inference-pytorch.md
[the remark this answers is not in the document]
stas00: Thank you for the feedback, @tjruwase - I will remove it then.

In general the offloading should work, but with 176B it's just too slow to be practical. 

Perhaps leaving a mentioning of DS-ZeRO to say that its offloading features might be useful for smaller models? What do you think?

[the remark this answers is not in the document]
tjruwase: @stas00, thanks for offering to put a plug for offload inference solution. By the way, the correct name as proposed in our [inference paper](https://arxiv.org/pdf/2207.00032.pdf) is ZeRO-Inference not DS-ZeRO.  I think DS-Inference will remain faster even for smaller models. So, in terms of pitch, I think the real value of ZeRO-Inference is when offloading is needed, for example to inference 176b bloom on a single GPU. I am not sure how such a pitch fits into the server inference theme of this blog. I will let you decide either way. Thanks!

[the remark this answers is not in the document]
stas00: I started using DS-ZeRO to differentiate it from DS-Inference, since DS-ZeRO-Inference was too confusing with DS-Inference.

So ZeRO and Inference are clear stand apart.

what do you think? I can of course use the full DS-ZeRO-Inference but I don't think it's the best approach, given the notes above.

I don't dare to even try doing offload with a single GPU, after my experiment with ZeRO/NVME offload took 8h to process one batch. Surely, CPU Offload will be better, but how many people who have only one GPU have 512GB of CPU RAM - so this doesn't quite make sense.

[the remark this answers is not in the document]
tjruwase: > I don't dare to even try doing offload with a single GPU, after my experiment with ZeRO/NVME offload took 8h to process one batch. Surely, CPU Offload will be better, but how many people who have only one GPU have 512GB of CPU RAM - so this doesn't quite make sense.

@stas00, sorry you had such a horror experience with NVMe offload. I agree that if BLOOM token generation with zero-inference took 8 hours per token then it wouldn't make sense. However, this is neither the expected behavior nor my experience, so zero-inference was still quite broken when you tried this. The expectation for NVMe offload is that since most devices are capable of 3GB/sec reads, streaming ~360GB BLOOM weights from a single NVMe into GPU to generate a token should take about 120 seconds, and not 8 hours. Using more NVMe devices should improve performance up to the PCIe bandwidth (e.g., ~16GB/sec) of a single GPU. These are the kind of numbers that I get in my experiments. We can sync offline if needed.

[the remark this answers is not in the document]
stas00: or really, my experiment was done just recently - I think 2 months ago - has something changed since then? I'd be happy to try again. I think it'd be great if anybody with a small GPU could run generation even if it takes a longish time, but not hours.

It was 100 tokens, not one with bs=1. and 1x A100 and very fast very recent gen4 nvme and PCIe-4.

And yes we can discuss this for sure if you think we can speed it up significantly.

bloom-inference-pytorch.md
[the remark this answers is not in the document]
Narsil: It is a very short input/output, still surprising to me.

Narsil:
@stas00 I added a few things, don't hesitate to comment or modify directly.

I felt that the first part and clarification about single forward latency vs throughput (token/ms) was needed otherwise readers would think we were actually getting 1 token in `0.7ms`. (which isn't correct afaik).

Latency = total run time / 5 (number of cyles) / 100 (tokens per cycle) should be roughly the time T showed in my diagram.
(Past vs initial inference *are* different) 

Usually I prefer to do PR over yours to not modify your work directly, but you seemed ok with the idea of pushing on top.

I checked the images on the blog on light/dark theme and everything should be readable.

bloom-inference-pytorch.md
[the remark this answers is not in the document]
tjruwase: Yeah, I think the code stabilized about 2 weeks ago. Hmm, generating 100 tokens for a bs=1 prompt will require reading the model 100 times which brings e2e latency to 100*120secs = ~3.3hours. Did I understand the scenario correctly? Perhaps this the hours latency you are concerned about. Of course, we could amortize by increasing batch size to around 64 as the extra compute time is negligible compared to PCIe transfer time. My experiments were focused on time per token generation, and not for all the tokens.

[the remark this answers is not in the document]
tjruwase: @stas00, but I am interested in exploring a bit further, so please share your script with me. I can try on my side. Thanks!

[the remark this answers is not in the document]
stas00: It's the same script from the inference PR here:

https://github.com/bigscience-workshop/Megatron-DeepSpeed/pull/308/files#diff-b772a56ca7676205ef4bcf14ad5fd704246e81d7e35e6d41807775ebbb8b6bbf

The scenario is of a practical need - how long will it take a user who wants to do a single 100 token generation - and the desire is for the script to finish in a reasonable length of time. If we can do it in 3h with a single small GPU I think that would be quite amazing, given the size of the model.

stas00:
Hmm, editing the PR directly is fine. but you force-pushed which is not the same as now I have no idea what you have changed other than manually doing a diff of the last version before your force-pushes. If you were to edit the files and commit them directly github would have shown the diff.

Usually please avoid force pushing in PRs that are a collaboration work. As besides the diff issue, it also makes it very difficult to push local changes if they were made before the force-push but weren't committed.

stas00:
I think the images weren't linked correctly, you can see in the preview that https://github.com/stas00/blog/blob/bloom-inference-pytorch/bloom-inference-pytorch.md that they are 404.

Narsil:
> but you force-pushed which is not the same as now I have no idea what you have changed other than manually doing a diff of the last version before your force-pushes.

That was simply a rebase. But I understand. I never get why github forces us to force-push to rebase but there's no real way around it afaik. Here I shouldn't have, but did it by habit because of the merge conflict.

And the commit list is the same, so my changes are still readable as independant commits. https://github.com/huggingface/blog/pull/432/commits/60d5fb8b08767934b6639b4f548e01ba0306d9ca (Ofc I could have swept the rug under you, but I didn't, at least to the best of my knowledge)

Narsil:
> I think the images you added weren't linked correctly, you can see in the preview that https://github.com/stas00/blog/blob/bloom-inference-pytorch/bloom-inference-pytorch.md that they are 404.

Actually the links I correct. I verified on a local build of the blog. I seems your original PR had the same URL and same failure for you image: https://github.com/stas00/blog/blob/1a62600fa3ce36a9949b00908e4ed3a7681a3b68/bloom-inference-pytorch.md

The thing is that the github markdown is confused by the absolute path of the image (which isn't correct in github viewing but is correct once on the hub).

stas00:
You don't need to force push to rebase, here is a tool I wrote that I use everywhere to re-base
https://github.com/stas00/git-tools/tree/master/git-rebase

stas00:
basically my local clone is completely broken right now after these force pushes. I'm just going to make a new clone.

stas00:
I'm not sure how to proceed - you wiped out all the editorial work that has been done so far, deleting all the fixes :( 

Please advise.

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,211 @@
...
+
+Let's try and show the metrics in a drawing:
+
+![initial_drawing](assets/bloom-inference-pytorch/initial_drawing.png)
+
+In this image, **T** is the simple forward latency. This is determined by how efficiently the **forward** pass is coded. But as you can see, using `8xA100` gives plenty of computing power so we can increase the batch size, and that's (to a point) not going to change **T**. But you will be generating more tokens within **T** so your throughput is effectively going up. As you increase your batch size, your memory consumption will likely go up too, so you might hit OOM errors, **before** actually reaching your computational boundary.
```
stas00, line 67: Your definition of T is very ambiguous - you first defined it as all of forwards together here and later in the images you used a single forward as T.

Moreover, not all forwards are created equal. some are faster some are slower. You probably need to state which of them you're referring here.

bloom-inference-pytorch.md
```diff
@@ -0,0 +1,211 @@
...
+
+![overall_latency](assets/bloom-inference-pytorch/overall_latency.png)
+
+As you can see here, the first request gets a latency of `3 x T` which is what we would expect. But **request 2** has to wait for `4.5 x T`, which is longer. If there was a **request 3** with yet another parameter set, then you would have to wait for, say,`7.5 x T`. And it piles up. The simple solution is to force a single (or handful) parameter sets so that we can ensure we're capping the overall latency.
+
+Another approach is to split the generation loop entirely. We're going to ignore the `use_cache=True` complexity for the sake of readability.
+
+So now, what we're going to do is to have each request, handle the `generate` part on its own, and simply send back the required tensors only for the `forward` pass. A single thread will be responsible for batching those simple requests.
+
+Since each request is handling entirely its own way of choosing next ids, there is no problem in handling an infinite amount of different parameters.
```
stas00, line 172: what is "next ids"?

what is "different parameters"?

I think it's very difficult to follow as the text jumps into technical jargon without defining which is which. I lost you several times while reading this text.

PR #435, opened 2022-07-22
[DRL course] Fix latex formula

roykoand:
Hi Team,
Thanks a lot for the new lesson! Since the first unit wanted to play with PyBullet. 
It looks like here we have a LaTeX issue:
image
I've checked this in the online latex editor, and it should be fixed just by deleting the backslash:
image


osanseviero (approved):
Thanks so much for the fix!

cc @simoninithomas for visibility

PR #438, opened 2022-07-25
Misc minor changes of Deep RL Class unit 7

osanseviero:
The most important one was the table of contents, but apart from that all changes are minor


simoninithomas (approved):
Thanks for the nice updates, I merge it 🤗

PR #439, opened 2022-07-25
added mturk on spaces blog post

TristanThrush:
This PR adds a blog post/tutorial about how to do mechanical turk data collection with a hugging face space.


mturk-on-spaces.md
```diff
@@ -0,0 +1,176 @@
+---
+title: "How to collect a dataset with crowdworkers, all from a Hugging Face space"
+thumbnail: /blog/assets/91_mturk_on_spaces/thumbnail.png
+---
+
+<h1>
+    How to collect a dataset with crowdworkers, all from a Hugging Face space
```
douwekiela, line 7: I'm not sold on the "all from a" - this makes it sound like we can trigger the data collection from a space itself, while it requires a separate script

mturk-on-spaces.md
```diff
@@ -0,0 +1,176 @@
...
+        </div>
+    </a>
+</div>
+
+
+### TLDR
+
+Do you want to gather data from crowdworkers but don’t know where to start? This tutorial is for you! Here, you will learn to automate the "crowdworker to Hugging Face dataset" pipeline, in only a few lines of code. Specifically, you will learn how to:
+
+- Run your Hugging Face space as a data-collection interface in Mechanical Turk (a popular data collection platform)
```
douwekiela, line 46: Insert link to mturk?

mturk-on-spaces.md
```diff
@@ -0,0 +1,176 @@
...
+- Make your space store data in a Hugging Face dataset automatically
+- Do Dynamic Adversarial Data Collection (DADC), to improve the quality of the collected data. In DADC, there is a model providing feedback to human crowdworkers.
```
douwekiela, line 48: I don't think we have to emphasize this here - it'll look similar to Chris's post which is very different, and technically we're not "dynamic" here, it's just a single model in the loop. I'd prefer "human and model in the loop data collection" here, or simply "model in the loop data collection" since it's obvious that humans are involved.

mturk-on-spaces.md
```diff
@@ -0,0 +1,176 @@
...
+
+
+## Overview
+In this post, we showcase a space which collects a sentiment analysis dataset. Examples in this dataset are `text, target` pairs like this:
```
douwekiela, line 52: The space doesn't do the collection, we do - maybe rephrase?

mturk-on-spaces.md
```diff
@@ -0,0 +1,176 @@
...
+
+```
+text: “the movie was great”, target: “positive”
```
douwekiela, line 55: put the target on a separate line?

mturk-on-spaces.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+Following the DADC paradigm, we also deploy a Hugging Face model in the space, so it can give crowdworkers feedback about whether they are creating hard examples which fool it.
```
douwekiela, line 58: I don't think we can assume people will know what "the DADC paradigm is", and I don't think it matters that much (see above). I also think that we should be clear immediately about whether we're collecting the dataset _from scratch_ or not (the way it's written, people will immediately wonder "huh I'm collecting a sentiment dataset but I already have a model for it?")

mturk-on-spaces.md
```diff
@@ -0,0 +1,176 @@
...
+
+We make the UI with [Gradio](https://huggingface.co/blog/gradio-spaces). Below is the space, embedded in an iframe. You can also access the space [here](https://huggingface.co/spaces/Tristan/dadc). In the space, assignments are saved in chunks of two examples. Try creating a couple examples. They will be saved in a dataset which you will see later. Afterwards, a “Submit Work” button will appear for you. For turkers, the “Submit Work” button will send a message to Mechanical Turk to give them credit for their two-example chunk, so they can get paid. For you, the “Submit Work” button will just refresh the space. On Mechanical Turk, these quick chunks of work are called Human Intelligence Tasks (HITs).
```
douwekiela, line 60: You need to walk people through this much more. The fact that it's embedded in an iframe doesn't mean anything to most readers, I'd just end this para with "Here's the space:" or something

douwekiela: By walking through I mean:
* What is MTurk
* Why do people use it for data collection
* What is difficult about data collection with Mturk right now
* How does a space make it easier
* What is the usual flow when doing data collection on Mturk
Assume you're explaining this to someone who doesn't know anything about these things.

mturk-on-spaces.md
```diff
@@ -0,0 +1,176 @@
...
+
+<iframe src="https://hf.space/embed/Tristan/dadc/+?__theme=light" frameBorder="0" width="100%" height="1400px" title="Gradio app" allow="accelerometer; ambient-light-sensor; autoplay; battery; camera; document-domain; encrypted-media; fullscreen; geolocation; gyroscope; layout-animations; legacy-image-formats; magnetometer; microphone; midi; oversized-images; payment; picture-in-picture; publickey-credentials-get; sync-xhr; usb; vr ; wake-lock; xr-spatial-tracking" sandbox="allow-forms allow-modals allow-popups allow-popups-to-escape-sandbox allow-same-origin allow-scripts allow-downloads"></iframe>
+
+
+## How does the space store examples into a Hugging Face dataset?
```
douwekiela, line 65: This comes out of the blue to me

mturk-on-spaces.md
```diff
@@ -0,0 +1,176 @@
...
+            repo.git_push()
+    if not f_stop.is_set():
+        # call again in 60 seconds
+        threading.Timer(60, asynchronous_push, [f_stop]).start()
+
+f_stop = threading.Event()
+asynchronous_push(f_stop)
+```
+
+Every minute, this function checks the `.jsonl` file for updates. If it was updated, then the space pushes the `.jsonl` to the dataset [here](https://huggingface.co/datasets/Tristan/dadc-data/blob/main/data.jsonl). If you entered two examples following the section above, then you should see your examples stored in this dataset after a minute.
```
douwekiela, line 101: Is there a risk of data loss actually, if we are relying on the timer? What if the app is killed before the 60 seconds?

mturk-on-spaces.md
```diff
@@ -0,0 +1,176 @@
...
+
+## How do we run the space on Mechanical Turk?
+
+To run this app on Mechanical Turk, clone the space locally with:
+
+```
+git clone https://huggingface.co/spaces/Tristan/dadc
+```
+
+Then, follow the `README.md`. Essentially, all you need to do is run [`collect.py`](https://huggingface.co/spaces/Tristan/dadc/blob/main/collect.py), after providing it with your API keys for Mechanical Turk.
```
douwekiela, line 112: I think you need to write the steps out in the blog and not just point people to the README

mturk-on-spaces.md
```diff
@@ -0,0 +1,176 @@
...
+
+submit_hit_button.click(
+    lambda state: state,
+    inputs=[state],
+    outputs=[state],
+    _js=post_hit_js,
+)
+```
+
+In order for Mechanical Turk to notice the form submission, it must happen from the frontend of app.py. This is fine, because Gradio allows us to run javascript when a button is clicked.
```
douwekiela, line 164: I don't think many people will understand this

mturk-on-spaces.md
```diff
@@ -0,0 +1,176 @@
...
+Link to generated dataset:
+[https://huggingface.co/datasets/Tristan/dadc-data/blob/main/data.jsonl](https://huggingface.co/datasets/Tristan/dadc-data/blob/main/data.jsonl)
+
+Questions/Suggestions/Comments? Feel free to open a discussion on the space:
+[https://huggingface.co/spaces/Tristan/dadc/discussions](https://huggingface.co/spaces/Tristan/dadc/discussions)
+
+
+### Ethical considerations
+
+Note that crowdworker platforms offer very limited protections for their workers. It is up to you as the task owner to ensure that you pay crowdworkers enough and do not exploit them in any other ways. For example, see [https://arxiv.org/abs/1712.05796](https://arxiv.org/abs/1712.05796).
```
douwekiela, line 183: Please loop in eg Giada on this, I think we have more to say than just this

douwekiela (commented):
Nice one! I think this needs quite a bit of work: it's written for someone who already knows all of this stuff, and you should be much clearer on the problem you're solving, what the exact novelty of this solution is, and in general why this is so nice (no more clunky JS interfaces!)

mturk-on-spaces.md
TristanThrush: Yes. So someone should wait 60 seconds after the last hit was collected before they do something that kills their space. Although, they shouldn't need to kill their space.

mturk-on-spaces.md
TristanThrush: hmm yeah maybe a bit esoteric. I'll remove this in the revised version

mturk-on-spaces.md
TristanThrush: An alternative that I actually did first was to have the space push an update every time a requestor submits a hit, but I think that this would put unnecessary load on the space and the hugging face website. And data loss would still be possible if someone kills the space before a bunch of the pushes are complete.

mturk-on-spaces.md
TristanThrush: Thanks! Updating the draft with a transition.

PR #445, opened 2022-08-01
fix datasets version

nbroad1881:
@patrickvonplaten, after receiving an email about the datasets version being wrong, I made a correction.


patrickvonplaten:
Thanks!

PR #455, opened 2022-08-09
Add some clarity to deep rl, unit 2

theicfire:


deep-rl-q-part1.md
```diff
@@ -358,8 +358,6 @@ The idea with **TD is to update the \\(V(S_t)\\) at each step.**
 
 But because we didn't play during an entire episode, we don't have \\(G_t\\) (expected return). Instead, **we estimate \\(G_t\\) by adding \\(R_{t+1}\\) and the discounted value of the next state.**
 
-We speak about **bootstrap because TD bases its update part on an existing estimate \\(V(S_{t+1})\\) and not a complete sample \\(G_t\\).**
```
theicfire, line 361: Bootstrap was never mentioned. I think this can just be removed.

deep-rl-q-part2.md
```diff
@@ -329,7 +329,7 @@ Because I go to the poison state, **I get \\(R_{t+1} = -10\\), and I die.**
   
 Because we're dead, we start a new episode. But what we see here is that **with two explorations steps, my agent became smarter.**
 
-As we continue exploring and exploiting the environment and updating Q-values using TD target, **Q-Table will give us better and better approximations. And thus, at the end of the training, we'll get an optimal Q-Function.**
```
theicfire, line 332: Not optimal. Or at least, we didn't define optimal.

deep-rl-q-part1.md
simoninithomas: I think we could keep it but adding something, "we call it bootstrap". Because that's a term they're going to see a lot in papers and other resources.

deep-rl-q-part2.md
simoninithomas: In fact we talk about optimality in the first chapter and in this second rapidly where we explain why having an optimal value function leads to an optimal policy.

But I'm agree this phrase is not right because we don't reach the optimal Q-Function Q* but an estimation of this Q*. So I can modify this phrase remove this mistake.

theicfire: done, with more minor changes

deep-rl-q-part1.md
theicfire: done, with more minor changes

theicfire: I still think this could simply be removed.. it doesn't seem to add anything to this post. But for the sake of this PR, I've left it in.

theicfire:
Added new commits to handle your comments

simoninithomas:
Awesome, thanks for the update I merge it 🤗

PR #463, opened 2022-08-11
Blog post on `bitsandbytes` integration on Hugging Face

younesbelkada:
A blog post explaining the method `8-bit Matrix Multiplication for transformers at scale` and how this is used under the hood in `transformers` !

- Still a WIP PR

# TODOs

- [x] change the main thumbnail
- [x] add more content
- [x] correct content
- [x] re-capture the gif `out-quant.gif` because we can see the mouse


_blog.yml
```diff
@@ -1105,3 +1105,13 @@
   tags:
     - guide
     - cv
+
+- local: hf-bitsandbytes-integration
+  title: "A Gentle Introduction to 8-bit Matrix Multiplication for transformers at scale using transformers, accelerate and bitsandbytes"
+  author: ybelkada
+  thumbnail: /blog/assets/96_hf_bitsandbytes_integration/thumbnail.png
```
younesbelkada, line 1112:
```suggestion
  thumbnail: /blog/assets/96_hf_bitsandbytes_integration/thumbnail_blue.png
```
Use the blue one

stas00:
You might want to share in the OP the preview link that will remain fixed through this PR's mods.
https://github.com/younesbelkada/blog/blob/add_bnb_inference/hf-bitsandbytes-integration.md
it's much easier to read that way.

younesbelkada:
Great thanks for the tip! I have changed the previous names of some pictures and now all images should be rendered correctly

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,199 @@
...
+    </a>
+</div>
+
+# Introduction
+
+Language models are becoming larger: at the time of writing this blogpost, PaLM has 540B parameters, OPT, GPT-3 and BLOOM have around 176B parameters, and the current trend is still towards larger language models. Below is a qualitative diagram showing the size of some recent language models (original content).
+
+![LLM](assets/96_hf_bitsandbytes_integration/LLM.png)
+
+Therefore these models are hard to run on easily accessible devices. To properly run OPT-175B you would need to have around 5-6 NVIDIA A100 80GB, with each GPU costing ~15.000$.
```
stas00, line 39:
```suggestion
Therefore these models are hard to run on easily accessible devices. To properly run BLOOM-175B you would need to have around 5-6 NVIDIA 80GB A100 GPUs , with each GPU costing ~15.000$.
```

also perhaps use our own BLOOM as an example?

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,199 @@
...
+
+While running an open-source version of PaLM would be even more expensive, we at Hugging Face will certainly want to host these even larger models, just as we did for BLOOM, OPT, YaLM, so that the community can benefit from these powerful tools.
```
stas00, line 41:
```suggestion
While running an open-source version of PaLM would be even more expensive, we at Hugging Face will certainly want to host these even larger models, just as we did for BLOOM, OPT, YaLM, so that the community can benefit from these powerful models. 
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,199 @@
...
+
+![LLM](assets/96_hf_bitsandbytes_integration/LLM.png)
+
+Therefore these models are hard to run on easily accessible devices. To properly run OPT-175B you would need to have around 5-6 NVIDIA A100 80GB, with each GPU costing ~15.000$.
+
+While running an open-source version of PaLM would be even more expensive, we at Hugging Face will certainly want to host these even larger models, just as we did for BLOOM, OPT, YaLM, so that the community can benefit from these powerful tools. 
+
+Because these large models require a lot of resources to be run, it is a priotity to develop methods that allow us to run large models on less devices. To represent these large models faithfully however, we want to reduce the required number of devices and preserve performance at the same time – not an easy challenge for quantization and distillation approaches.
+
+Running large models on fewer devices with no performance degradation will be an open challenge for the next few years. In the future, Hugging Face will certainly be hosting these large models as we did it for BLOOM, OPT and YaLM, and we want users to benefit from the most powerful tools to run these models efficiently.
```
stas00, line 45: the second half of this para is a repeat of the 2 paras up. "host these even larger..."

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,199 @@
...
+
+At Hugging Face and BigScience, while training BLOOM-176B we were interested in reducing the main model’s size with the least performance degradation possible. That is how we came out collaborating with bitsandbytes to integrate the recent “GPT3.int8(): 8-bit Matrix Multiplication for Transformers at Scale'' paper on transformers. We decided to integrate it since no post-training quantization is required to run this feature and you can reduce the memory footprint of the models by 2x for the largest models with few lines of code. Let’s understand in this blogpost how this method works in a nutshell and how to use it in transformers!
+
+# A high level look
+
+Let us start from the beginning. The size of the model is determined by the number of its parameters, and the precision of the model is mostly determined by the precision of the parameters.
```
stas00, line 51:
```suggestion
Let us start at the beginning. 

The size of the model is determined by the number of its parameters, and their precision, typically one of float32, float16 or bfloat16. 
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,199 @@
...
+
+# A high level look
+
+Let us start from the beginning. The size of the model is determined by the number of its parameters, and the precision of the model is mostly determined by the precision of the parameters. 
+Let’s imagine 1 bit as an available memory case to store a binary value (0 or 1), and 1 byte being a bigger memory case that can store 8 bits. With one byte we can make 2^8=256 different patterns. Usually, 256 different values offers too limited precision and one should store numbers in 2 bytes (float16 or bfloat16), 4 bytes (float32) or 8 bytes (float64) to get more precision - aka to cover more possible patterns and store very precise numbers. For example, 2 bytes can store 65536 patterns.
```
stas00, line 52: this is very confusing. in particular patterns as it's difficult to follow what you mean by that. Why not say 256 numbers instead of 256 patterns.

"256 different values offers too limited precision" doesn't quite make logical sense 

If you really want to go into this low-level of details you will need to explain quite differently. I propose to have a very brief explanation of only 3 types: float32, float16 or bfloat16. and include the wikipedia graphics that show the mantissa. that's loud and clear. I'd also have a table showing the min/max representation - .e.g. fp16 is 64k max. so fp16 can't represent a number larger than 64k and will overflow instead.

Precision is then how many bytes you cut off and what is the loss comparatively to the normal number with full precision.

In our world fp32 is full precision and everything is relative to it.

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,199 @@
...
+
+
+![byte](assets/96_hf_bitsandbytes_integration/byte.png)
+
+By default, most common Deep Learning models are stored in float32 - fp32 (4 bytes per parameter) but Large language models are usually stored in half-precision, so either fp16 or bf16 precision (2 bytes per parameter) since the performance gap between fp32 models and fp16 is relatively acceptable, and fp16/bf16 allows for faster training that requires less memory.
```
stas00, line 57: hmm, why do you say so? All typical models are trained and stored in full-precision - only some parts of it are performed in half-precision. BLOOM's weights are in fp32 and not in half-precision, we chose to go with bf16 version of the weights because it's smaller. 

Here is the script that extracts the fp32 weights:
https://github.com/bigscience-workshop/Megatron-DeepSpeed/pull/327

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,199 @@
...
+
+![Model-storage](assets/96_hf_bitsandbytes_integration/Model-storage.png)
+
+How is a model's size calculated in practice? For instance BLOOM-176B that has been trained and stored in bf16 has roughly a total size of 176.000.000.000 x 2 bytes = 352GB! Fitting correctly such a high memory requirement into GPU RAM would be quite challenging.
```
stas00, line 61: same as above comment.

```suggestion
To calculate the model size you multiply the number of parameters by the size of the chosen precision in bytes.  For instance, If we use the bfloat16 version of the BLOOM-176B model we have `176*10**9 x 2 bytes = 352GB`! As discussed earlier this is quite a challenge to fit into a few GPUs.
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,199 @@
...
+
+Running large models on fewer devices with no performance degradation will be an open challenge for the next few years. In the future, Hugging Face will certainly be hosting these large models as we did it for BLOOM, OPT and YaLM, and we want users to benefit from the most powerful tools to run these models efficiently.
+
+At Hugging Face and BigScience, while training BLOOM-176B we were interested in reducing the main model’s size with the least performance degradation possible. That is how we came out collaborating with bitsandbytes to integrate the recent “GPT3.int8(): 8-bit Matrix Multiplication for Transformers at Scale'' paper on transformers. We decided to integrate it since no post-training quantization is required to run this feature and you can reduce the memory footprint of the models by 2x for the largest models with few lines of code. Let’s understand in this blogpost how this method works in a nutshell and how to use it in transformers!
+
+# A high level look
```
stas00, line 49: a low level look?

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,199 @@
...
+![byte](assets/96_hf_bitsandbytes_integration/byte.png)
+
+By default, most common Deep Learning models are stored in float32 - fp32 (4 bytes per parameter) but Large language models are usually stored in half-precision, so either fp16 or bf16 precision (2 bytes per parameter) since the performance gap between fp32 models and fp16 is relatively acceptable, and fp16/bf16 allows for faster training that requires less memory.
+
+![Model-storage](assets/96_hf_bitsandbytes_integration/Model-storage.png)
+
+How is a model's size calculated in practice? For instance BLOOM-176B that has been trained and stored in bf16 has roughly a total size of 176.000.000.000 x 2 bytes = 352GB! Fitting correctly such a high memory requirement into GPU RAM would be quite challenging. 
+But what if we can store those weights using less memory using a different precision? A trick called quantization has been widely used in Deep Learning and let’s see how it works! 
+
+# Quantizing models in few words
```
stas00, line 64:
```suggestion
# Quantizing models in a few words
```

stas00 (commented):
first part

I will stop here for now, the text needs heavy editing - if you'd trust my approach - perhaps it'd be easier if I edited the text directly? as it's quite painful to edit each sentence in github. You will still see a diff and we can tweak further.

younesbelkada:
Thank you very much for your review! This will be extremely helpful
Yes we can go for this approach, I can give you access to the google doc I used to write the text if you prefer

hf-bitsandbytes-integration.md
stas00: if you were to train in half-precision it'd be very bad and will take much longer to train. At least with the current tech.

hf-bitsandbytes-integration.md
stas00: steal the image from here:
https://en.wikipedia.org/wiki/Bfloat16_floating-point_format

younesbelkada: Ohh thanks! Yes the images are better there and easier to understand

hf-bitsandbytes-integration.md
younesbelkada: I think that I will remove this sentence, the 2 paragraphs before seems to be enough

younesbelkada:
```suggestion
```

hf-bitsandbytes-integration.md
younesbelkada: Totally yes!
```suggestion
# A low level look
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,201 @@
...
+
+Language models are becoming larger: at the time of writing this blogpost, PaLM has 540B parameters, OPT, GPT-3 and BLOOM have around 176B parameters, and the current trend is still towards larger language models. Below is a qualitative diagram showing the size of some recent language models (original content).
+
+![LLM](assets/96_hf_bitsandbytes_integration/LLM.png)
+
+Therefore these models are hard to run on easily accessible devices. To properly run BLOOM-175B you would need to have around 5-6 NVIDIA 80GB A100 GPUs , at ~$15k a piece.
+
+While running an open-source version of PaLM would be even more expensive, we at Hugging Face will certainly want to host these even larger models, just as we did for BLOOM, OPT, YaLM, so that the community can benefit from these powerful models. 
+
+Because these large models require a lot of resources to be run, it is a priotity to develop methods that allow us to run large models on less devices. To represent these large models faithfully however, we want to reduce the required number of devices and preserve performance at the same time – not an easy challenge for quantization and distillation approaches.
```
younesbelkada, line 43:
```suggestion
Because these large models require a lot of resources to be run, it is a priotity to develop methods that allow us to run large models on less devices. To represent these large models faithfully however, we want to reduce the required number of devices and preserve performance at the same time – not an easy challenge for quantization and distillation approaches. Therefore, running these models on fewer devices with no performance degradation will be an open challenge for the next few years.
```

stas00:
no google doc needed, we are already editing here, it'd be extremely difficult to keep 2 separate docs in sync. So let's switch to editing here 100%.

As I shared earlier the preview link should serve as a nice visual that's easier to read than raw MD.

stas00:
I've pushed the rewrite of the first few sections. Please let me know if it feels good and then I will continue reworking the rest when you're asleep next time, so that we don't collide on our edits.

I wonder if perhaps you'd want to first finish the rough version first and I will do the heavy editing afterwards? It's your call.

younesbelkada:
Thank you very much @stas00 ! I have updated more content and tried to rephrase as much as I could! I would love to have a review pass until `How to use it in transformers` section 💪

younesbelkada:
I have addressed your points and comment on them! Let me know if they make more sense now 💪

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+| hellaswag  | acc\_norm |           0.7274 |       0.7303 |         0.0044 |     0.0044 | 0.0029 |
+| hellaswag  | acc       |           0.5563 |       0.5584 |          0.005 |      0.005 | 0.0021 |
+| piqa       | acc       |           0.7835 |       0.7884 |         0.0096 |     0.0095 | 0.0049 |
+| piqa       | acc\_norm |           0.7922 |       0.7911 |         0.0095 |     0.0095 | 0.0011 |
+| lambada    | ppl       |           3.9191 |        3.931 |         0.0842 |     0.0846 | 0.0119 |
+| lambada    | acc       |           0.6808 |       0.6718 |         0.0065 |     0.0065 |  0.009 |
+| winogrande | acc       |           0.7048 |       0.7048 |         0.0128 |     0.0128 |      0 |
+
+
+We indeed observe 0 performance degradation for those models since the absolute difference of the metrics are all below the standard error (except for BLOOM-int8 which is slightly better than the native model on lambada). For more detailed performance evaluation against state of the art approaches you may look closely at the paper!
```
younesbelkada, line 199: TODO: add paper link

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+![out-quant.gif](assets/96_hf_bitsandbytes_integration/out-quant.gif)
+
+For an unsigned int8, we would subtract the minimum and scale by the absolute maximum. This is close to what zero-point quantization does, a min-max scaling with the difference that zero-point quantization maintains scales the values in such a way that the value “0” is always represented by an integer without any quantization error. 
+
+XXX: I lost you here, this example is half-baked - as you didn't divide by the total range and it's very difficult to make sense of the above para. May I suggest to use the exact same example of `0.3` here so that the reader could build upon what he already learnt? -> proposed a suggestion above let me know if it is more clear!
+
+
+These tricks can be combined in several ways, for example row-wise or vector-wise quantization when it comes to matrix multiplication for more accurate results.
+
+If you want to read more details about how classic quantization techniques work, we recommend to read this [blogpost](https://intellabs.github.io/distiller/algo_quantization.html), or read the GPT3.int8() paper (XXX: link).
```
younesbelkada, line 132: Do not forget to put paper link

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+This approach, in our opinion, greatly improves access to very large models. With no performance degradation, it enables users with less compute to access models that were previously inaccessible.
+We've found several areas for improvement that can be worked on in the future to make this method even better for large models!
+
+## Inference speed and slowing down on smaller models
+
+For very large language models, we have observed that we nearly maintain the same inference speed using the native model as opposed to the mixed-8bit model (see attached experiments on BLOOM-176B).
+
+However, due to the numerous internal casting steps, together with the outlier detection procedure that take place inside each 8bit-Linear layer, this method can significantly slow down inference speed on small models (models with less than 6b parameters).
+
+One could attempt to improve that in the future and see how the inference speed can be decreased, probably by avoiding the casting operations or writing more efficient CUDA kernels.
```
younesbelkada, line 463: TODO: add link to the GH issue

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+B = self.data.contiguous().half().cuda(device)
+CB, CBt, SCB, SCBt, coo_tensorB = bnb.functional.double_quant(B)
+del CBt
+del SCBt
+self.data = CB
+setattr(self, 'CB', CB)
+setattr(self, 'SCB', SCB)
+```
+
+Here, setting a parameter's device step is extremely crucial since the quantization statistics fails when calling it twice. We had to came up with our implementation of `accelerate`'s `set_module_tensor_to_device` function (termed as `set_module_8bit_tensor_to_device`) and propose some Pull Requests on `accelerate` library; all of that to make sure this function is never called twice. Let's in detail in the section below!
```
stevhliu, line 402:
```suggestion
Here, setting a parameter's device step is extremely crucial since the quantization statistics fails when calling it twice. We had to come up with an implementation of `accelerate`'s `set_module_tensor_to_device` function (termed as `set_module_8bit_tensor_to_device`) to make sure we don't call it twice. Let's discuss this in detail in the section below!
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+
+## Be very careful on how to set devices with `accelerate`
+
+Here we played a very delicate balancing act with `accelerate` library!
```
stevhliu, line 406:
```suggestion
Here we played a very delicate balancing act with the `accelerate` library!
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+Once you load your model and set it on the correct devices, sometimes you still need to call `set_module_tensor_to_device` to dispatch the model with hooks on all devices. This is done inside `dispatch_model` function from `accelerate`. This will involve potentially calling `.to` several times, which is something we want to avoid.
```
stevhliu, line 407:
```suggestion
Once you load your model and set it on the correct devices, sometimes you still need to call `set_module_tensor_to_device` to dispatch the model with hooks on all devices. This is done inside the `dispatch_model` function from `accelerate`, which involves potentially calling `.to` several times and is something we want to avoid.
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+2 Pull Requests were needed to achieve what we wanted! The initial PR proposed [here](https://github.com/huggingface/accelerate/pull/539/) broke some tests but [this PR](https://github.com/huggingface/accelerate/pull/576/) successfully fixed everything!
+
+## Wrapping up all together
```
stevhliu, line 410:
```suggestion
## Wrapping it all up
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+
+Therefore the ultimate recipe is:
+1. Initialize a model in the `meta` device with the correct modules
+2. Set the parameters one by one on the correct GPU device and make sure you never do this procedure twice!
+3. Put new keyword arguments in the correct place everywhere, and add some nice documentation
+4. Add very extensive tests! Check our tests [here](https://github.com/huggingface/transformers/blob/main/tests/mixed_int8/test_mixed_int8.py) for more details
+This may sound quite easy, but actually went through a lot of hard debugging sessions with sometimes debugging CUDA kernels together!
```
stevhliu, line 417:
```suggestion
This may sound quite easy, but we went through many hard debugging sessions together, often times involving CUDA kernels!
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+
+All that said, this integration adventure was very fun, deep diving and doing some "surgery" on different libraries to align everything and make it work was pure fun!
```
stevhliu, line 419:
```suggestion
All said and done, this integration adventure was very fun; from deep diving and doing some "surgery" on different libraries to aligning everything and making it work!
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+
+Now time to see how to benefit from this integration and how to successfully use it in `transformers`!
+
+# How to use it in `transformers`
+
+## Hardware requirements
+
+8-bit tensor core are not supported on the CPU. bitsandbytes can be run on 8-bit tensor cores-supported hardwares, which are Turing and Ampere GPUs (RTX 20s, RTX 30s, A40-A100, T4+). For example, Google Colab GPUs are usually NIVIDIA T4 GPUs and the latest generation of GPUs does support 8-bit cores. Our demos are based on Google Colab so check it out below!
```
stevhliu, line 427:
```suggestion
8-bit tensor cores are not supported on the CPU. bitsandbytes can be run on 8-bit tensor core-supported hardware, which are Turing and Ampere GPUs (RTX 20s, RTX 30s, A40-A100, T4+). For example, Google Colab GPUs are usually NVIDIA T4 GPUs, and their latest generation of GPUs does support 8-bit cores. Our demos are based on Google Colab so check them out below!
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+
+```bash
+pip install accelerate
+pip install bitsandbytes
+pip install git+https://github.com/huggingface/transformers.git
+```
+
+## Example demos - running T5 11b on a Google Colab
+
+Checkout the Google colab demos for running 8bit models on a Google Colab using BLOOM-3b model !
```
stevhliu, line 441:
```suggestion
Check out the Google Colab demos for running 8bit models on a BLOOM-3B model!
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+
+Here is the demo for running T5-11b (42GB in fp16)! Using 8-bit modules on Google Colab:
```
stevhliu, line 443:
```suggestion
Here is the demo for running T5-11B (42GB in fp16) with 8-bit modules on Google Colab:
```

stevhliu (commented):
Very cool blog post! Good job explaining the basics, then gradually building on it to explain the different quantization techniques and showing how it is all implemented in `bitsandbytes` and `transformers`.

Sorry for the huge number of comments, but most of them are just minor grammatical things to improve readability and flow 🤗

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+
+![thumbnail](assets/96_hf_bitsandbytes_integration/Thumbnail_blue.png)
+
+# Introduction
+
+Language models have been becoming larger all the time. At the time of this writing, PaLM has 540B parameters, OPT, GPT-3 and BLOOM have around 176B parameters, and the current trend is towards even larger models. Below is a qualitative diagram showing the size of some recent language models.
+
+![LLM](assets/96_hf_bitsandbytes_integration/LLM.png)
+
+At this same time, PaLM has 540B parameters, OPT, GPT-3 and BLOOM have around 176B parameters, and the current trend is towards even larger models.
```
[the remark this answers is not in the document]
younesbelkada:
```suggestion

```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+![Model-storage](assets/96_hf_bitsandbytes_integration/Model-storage.png)
+
+To calculate the model size in bytes one multiplies the number of parameters by the size of the chosen precision in bytes.  For example, if we use the bfloat16 version of the BLOOM-176B model we have `176*10**9 x 2 bytes = 352GB`! As discussed earlier this is quite a challenge to fit into a few GPUs.
+
+But what if we can store those weights using less memory using a different data type? A methodology called quantization has been used widely in Deep Learning.
+
+
+# Introduction to model quantization
+
+Experientially we have discovered that instead of using the 4-byte fp32 precision, we can get an almost identical inference outcome with 2-byte bf16/fp32 half-precision, which halves the model size. It'd be amazing to cut it further, but the inference quality outcome starts to drop dramatically at lower precision.
```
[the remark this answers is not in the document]
younesbelkada:
```suggestion
Experientially we have discovered that instead of using the 4-byte FP32 precision, we can get an almost identical inference outcome with 2-byte BF16/FP16 half-precision, which halves the model size. It'd be amazing to cut it further, but the inference quality outcome starts to drop dramatically at lower precision.
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+
+XXX: I lost you here, this example is half-baked - as you didn't divide by the total range and it's very difficult to make sense of the above para. May I suggest to use the exact same example of `0.3` here so that the reader could build upon what he already learnt? -> proposed a suggestion above let me know if it is more clear!
+
+
+These tricks can be combined in several ways, for example row-wise or vector-wise quantization when it comes to matrix multiplication for more accurate results.
+
+If you want to read more details about how classic quantization techniques work, we recommend to read this [blogpost](https://intellabs.github.io/distiller/algo_quantization.html), or read the GPT3.int8() paper (XXX: link).
+
+While these basic techniques enable us to quantize transformers models, they usually lead to a drop in accuracy for larger models. The bnb-int8 implementation that we integrated into Hugging Face Transformers and Accelerate libraries is the first technique that does not degrade performance even for large models with 176B parameters, such as BLOOM.
```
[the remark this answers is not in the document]
younesbelkada: Actually it refers to all Deep Learning models, let me update this!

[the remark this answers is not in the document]
younesbelkada:
```suggestion
While these basic techniques enable us to Deep Learning models, they usually lead to a drop in accuracy for larger models. The bnb-int8 implementation that we integrated into Hugging Face Transformers and Accelerate libraries is the first technique that does not degrade performance even for large models with 176B parameters, such as BLOOM.
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+
+Now let's look at how it works.
```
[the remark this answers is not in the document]
younesbelkada: Great agreed on that!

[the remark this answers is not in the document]
younesbelkada:
```suggestion
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+
+# A gentle summary of mixed int8 matrix multiplication for Large Language Models
+
+Authors have demonstrated that it is crucial to comprehend the scale-dependent emergent properties of transformers in order to understand why traditional quantization fails for large models. They demonstrate that performance deterioration is caused by outlier features.
+
+XXX: it's not clear what is meant by "outlier features" - an example would help. I think the paper goes into details. -> I added a small explanation and explicitly said tha the gif provides a visual example
+
+In essence, 8-bit Matrix Multiplication at Scale for Transformers seeks to complete the matrix multiplication computation in three steps:
+1. From the input hidden states, extract the outliers (i.e. values that are larger than a certain threshold) by column.
+2. Perform the matrix multiplication of the outliers in fp16 and the non-outliers in int8
```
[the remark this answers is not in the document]
younesbelkada:
```suggestion
2. Perform the matrix multiplication of the outliers in fp16 and the non-outliers in int8.
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,488 @@
...
+3. Dequantize the non-outlier results and retrieve the full result in fp16
```
younesbelkada, line 150:
```suggestion
3. Dequantize the non-outlier results and retrieve the full result in fp16.
```

hf-bitsandbytes-integration.md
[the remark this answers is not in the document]
younesbelkada: Thanks for pointing out this nit!

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,486 @@
...
+![out-quant.gif](assets/96_hf_bitsandbytes_integration/out-quant.gif)
+
+For an unsigned int8, we would subtract the minimum and scale by the absolute maximum. This is close to what zero-point quantization does; a min-max scaling with the difference being zero-point quantization maintains the value scales in such a way that the value “0” is always represented by an integer without any quantization error. 
+
+XXX: I lost you here, this example is half-baked - as you didn't divide by the total range and it's very difficult to make sense of the above para. May I suggest to use the exact same example of `0.3` here so that the reader could build upon what he already learnt? -> proposed a suggestion above let me know if it is more clear!
+
+
+These tricks can be combined in several ways, for example row-wise or vector-wise quantization, when it comes to matrix multiplication for more accurate results.
+
+If you want to read more details about how classic quantization techniques work, we recommend reading this [blog post](https://intellabs.github.io/distiller/algo_quantization.html) or the GPT3.int8() paper (XXX: link).
```
younesbelkada, line 133: TODO: add link to paper

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,486 @@
...
+
+While these basic techniques enable us to Deep Learning models, they usually lead to a drop in accuracy for larger models. The bnb-int8 implementation that we integrated into Hugging Face Transformers and Accelerate libraries is the first technique that does not degrade performance even for large models with 176B parameters, such as BLOOM.
+
+
+
+# A gentle summary of mixed int8 matrix multiplication for Large Language Models
+
+Authors have demonstrated that it is crucial to comprehend the scale-dependent emergent properties of transformers in order to understand why traditional quantization fails for large models. They demonstrate that performance deterioration is caused by outlier features.
+
+XXX: it's not clear what is meant by "outlier features" - an example would help. I think the paper goes into details. -> I added a small explanation and explicitly said tha the gif provides a visual example
```
younesbelkada, line 143: TODO: fix the animation (add the last frame)

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,486 @@
...
+
+To retrieve the latest, one can just divide in full precision the int8 number with the quantization factor, but since the result above is "rounded" some precision will be lost.
+
+![out-quant.gif](assets/96_hf_bitsandbytes_integration/out-quant.gif)
+
+For an unsigned int8, we would subtract the minimum and scale by the absolute maximum. This is close to what zero-point quantization does; a min-max scaling with the difference being zero-point quantization maintains the value scales in such a way that the value “0” is always represented by an integer without any quantization error. 
+
+XXX: I lost you here, this example is half-baked - as you didn't divide by the total range and it's very difficult to make sense of the above para. May I suggest to use the exact same example of `0.3` here so that the reader could build upon what he already learnt? -> proposed a suggestion above let me know if it is more clear!
```
younesbelkada, line 128: TODO: add last frame

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,484 @@
...
+
+# Common data types used in Machine Learning
+
+We start with the basic understanding of different floating point data types, which are also referred to as "precision" in the context of Machine Learning.
+
+The size of a model is determined by the number of its parameters, and their precision, typically one of float32, float16 or bfloat16 (image below from: https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format/).
+
+![Summary](assets/96_hf_bitsandbytes_integration/tf32-Mantissa-chart-hi-res-FINAL.png)
+
+Float32 (FP32) stands for the standard 32-bit floating point representation. With this data type it is possible to represent a wide range of floating numbers. In FP32, 8 bits are reserved for the "exponent", 23 bits for the "mantissa" and 1 bit for the sign of the number. In addition to that, most of the hardware supports FP32 operations and instructions.
```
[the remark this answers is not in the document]
younesbelkada:
```suggestion
Float32 (FP32) stands for the standard 32-bit floating point representation. With this data type it is possible to represent a wide range of floating numbers. Note that any floating point number can be written as a product of three terms: *sign* x *mantissa* x *base^exponent* (Image below taken from: https://en.wikipedia.org/wiki/Floating-point_arithmetic). 
![mantissa](assets/96_hf_bitsandbytes_integration/mantissa.svg)
In FP32, 8 bits are reserved for the "exponent", 23 bits for the "mantissa" and 1 bit for the sign of the number. In addition to that, most of the hardware supports FP32 operations and instructions.
```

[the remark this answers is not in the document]
younesbelkada: THank you very much! Added more comments and an image from the net for a more explicit explanation of this concept to put every reader on the same page

younesbelkada:
Thank you very much @stevhliu for the complete review! I validated them and addressed the points that needed more clarification :pray:

younesbelkada:
I left some comments for myself to integrate @stas00 comment regarding the animations that I will process later ;)

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,483 @@
...
+| name       | metric    | value - int8 - 6 | value - bf16 | err - int8 - 6 | err - bf16 |     -  |
+| hellaswag  | acc\_norm |           0.7274 |       0.7303 |         0.0044 |     0.0044 | 0.0029 |
+| hellaswag  | acc       |           0.5563 |       0.5584 |          0.005 |      0.005 | 0.0021 |
+| piqa       | acc       |           0.7835 |       0.7884 |         0.0096 |     0.0095 | 0.0049 |
+| piqa       | acc\_norm |           0.7922 |       0.7911 |         0.0095 |     0.0095 | 0.0011 |
+| lambada    | ppl       |           3.9191 |        3.931 |         0.0842 |     0.0846 | 0.0119 |
+| lambada    | acc       |           0.6808 |       0.6718 |         0.0065 |     0.0065 |  0.009 |
+| winogrande | acc       |           0.7048 |       0.7048 |         0.0128 |     0.0128 |      0 |
+
+
```
stas00, line 194: in both tables you have 4 input columns but only 1 diff column - which column the last one makes the diff of? it looks like the first 2 - so perhaps place it right after the first 2 columns?

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,483 @@
...
+| BLOOM-176-bf16 | 176B                 | 14xA100 40GB | 285                                             |                                            36.5 |                                             10.4 |      |
+| T5-11b-fp16         | 11B                         | 2xT4 15GB                                       |                                            11.7 |                                              1.7 |  0.5 |
+| T5-11b-int8         | 11B                         | 1xT4 15GB                                       |                                            43.5 |                                              5.3 |  1.3 |
+| T5-3b-fp32          | 3B                           | 2xT4 15GB                                       |                                              45 |                                              7.2 |  3.1 |
+| T5-3b-int8          |   3B                         | 1xT4 15GB                                       |                                             312 |                                             39.1 | 10.2 |
+
```
stas00, line 212: while it's commendable that you made many hardware set up benchmarks for 176B, I think the results make it much more difficult to make sense of the table. as it's very hard to see what is compared to what.

I'd say stick to the same hardware setup, e.g. the 40 or the 80, and no variation of the number of gpus, especially since 6xA100 and 5xA100 give the same results more or less.

or if you want to keep both A100s at least drop 6x 40 so that the table is a bit more symmetrical.

If possible I'd add some visual separation as well, so the reader reads the table in pairs of rows.

the 2nd column is sort of redundant as well, as it repeat the first column. Unless you want to switch the first column to just the precision use the second column as the model type+size.

Basically, less data is better readability wise.

stas00: I'd use this:

```
| Model          | Number of parameters | Hardware     | Time per token in milliseconds for Batch Size 1 | Time per token in milliseconds for Batch Size 8 | Time per token in milliseconds for Batch Size 32 |   |
| -------------- | -------------------- | ------------ | ----------------------------------------------- | ----------------------------------------------- | ------------------------------------------------ |   |
| bf16           | 176B                 | 8xA100 80GB  |                                             239 |                                              32 |                                              9.9 |   |
| int8           | 176B                 | 4xA100 80GB  |                                             282 |                                            37.5 |                                             10.2 |   |
| bf16           | 176B                 | 14xA100 40GB |                                             285 |                                            36.5 |                                             10.4 |   |
| int8           | 176B                 | 5xA100 40GB  |                                             367 |                                            46.4 |                                              oom |   |
| fp16           | 11B                  | 2xT4 15GB    |                                            11.7 |                                             1.7 |                                              0.5 |   |
| int8           | 11B                  | 1xT4 15GB    |                                            43.5 |                                             5.3 |                                              1.3 |   |
| fp32           | 3B                   | 2xT4 15GB    |                                              45 |                                             7.2 |                                              3.1 |   |
| int8           | 3B                   | 1xT4 15GB    |                                             312 |                                            39.1 |                                             10.2 |   |
|                |                      |              |                                                 |                                                 |                                                  |   |

The 3 models are BLOOM-176B, T5-11B and T5-3B

```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,483 @@
...
+| lambada    | ppl       |           3.9191 |        3.931 |         0.0842 |     0.0846 | 0.0119 |
+| lambada    | acc       |           0.6808 |       0.6718 |         0.0065 |     0.0065 |  0.009 |
+| winogrande | acc       |           0.7048 |       0.7048 |         0.0128 |     0.0128 |      0 |
+
+
+We indeed observe 0 performance degradation for those models since the absolute difference of the metrics are all below the standard error (except for BLOOM-int8 which is slightly better than the native model on lambada). For a more detailed performance evaluation against state-of-the-art approaches, take a look at the paper!
+
+## Is it faster than native models?
+
+We also benchmarked the inference speed of int8 models on different models. Although we are close to having the same speed as the native model for large models (tested on BLOOM-176), the inference speed seems to be much slower than the native model on smaller models.
```
stas00, line 199: > close to having the same speed 

is 15% and 23% slower correspondingly can be called close to having the same speed?

stas00: also you really one have one data point for larger models - it being bloom. Would probably be helpful to run OPT as well and perhaps some model in the middle range?

so it's hard to draw conclusions based on that. Unless you'd say that here is the theory and here are some examples that support it. May be...

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,483 @@
...
+
+# Which technology to use for `transformers` integration?
+
+How were these technologies incorporated into the `transformers` library? What were the difficulties we faced and the main techniques we employed in the integration project? Let's examine everything in the next sections!
+
+## How to use it in `bitsandbytes` library?
+
+The module responsible for the whole magic described in this blog post is called `Linear8bitLt` and you can easily import it from the `bitsandbytes` library. It is derived from a classic `torch.nn` Module and can be easily used and deployed in your architecture with the commands described below.
+
+Let's walk through step by step with a specific use case: let's say you want to convert a shallow model in int8 using `bitsandbytes`.
```
stas00, line 224: what did you mean by `shallow` here?

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,483 @@
...
+int8_model[0].weight
+Parameter containing:
+tensor([[   3,  -47,   54,  ...,   -5,  -44,   47],
+        [-104,   40,   81,  ...,  101,   61,   17],
+        [ -89, -127, -125,  ...,  120,   19,  -55],
+        ...,
+        [  82,   74,   65,  ...,  -49,  -53, -109],
+        [ -21,  -42,   68,  ...,   13,   57,  -12],
+        [  -4,   88,   -1,  ...,  -43,  -78,  121]], device='cuda:0',
+       dtype=torch.int8, requires_grad=True)
```
stas00, line 296: `requires_grad=True` shouldn't be consistent in both versions.

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,483 @@
...
+
+6. Now you can safely infer using your model by making sure your model is on the correct GPU:
+
+```py
+input_ = torch.randn(8, 64, dtype=torch.float16)
+hidden_states = int8_model(input_.to(0))
+```
+
+Check out [this gist](https://gist.github.com/younesbelkada/9035e247b066d1cf18682e9e4c21032d) for the full minimal code! Now the time has come to understand how to integrate that into the `transformers` library!
+
```
stas00, line 329: gist is likely to disappear in time, perhaps include the code snippet as a file in the assets? then it'll always remain found.

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,483 @@
...
+```py
+param_cls = type(module._parameters[name])
+kwargs = module._parameters[name].__dict__
+module._parameters[name] = param_cls(module._parameters[name].to(torch.device("meta")), **kwargs)
+```
+
+Now that this is fixed, we can easily leverage this context manager and play with it to replace all `nn.Linear` modules to `bnb.nn.Linear8bitLt` with no cost using a custom function!
+
+```py
+def replace_8bit_linear(model, threshold=6.0, modules_to_not_convert="lm_head"):
```
stas00, line 363: the mnemonics of `modules_to_not_convert` appears to be ambiguous - if it's **modules** it probably should expect a list instead and below check if `name in modules_to_not_convert`

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,483 @@
...
+                )
+    return model
+```
+
+This function recursively replaces all `nn.Linear` layers of a given model initialized on the `meta` device and replaces them with a `Linear8bitLt` module. The attribute `has_fp16_weights` has to be set to `False` in order to directly load the weights in `int8` together with the quantization statistics.
+
+We also discard the replacement for some modules (here the `lm_head`) since we want to keep the latest in their native precision for more precise and stable results.
+
+But it isn't over yet! The function above is executed under the `init_empty_weights` context manager which means that the new model will be still in the `meta` device.
+For models that are initialized under this context manager, `accelerate` later manually loads the parameters of each module and sets it on the correct device.
```
stas00, line 385: you need to define `later`

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,483 @@
...
+# we store the 8-bit rows-major weight
+# we convert this weight to the turning/ampere weight during the first inference pass
+B = self.data.contiguous().half().cuda(device)
+CB, CBt, SCB, SCBt, coo_tensorB = bnb.functional.double_quant(B)
+del CBt
+del SCBt
+self.data = CB
+setattr(self, 'CB', CB)
+setattr(self, 'SCB', SCB)
+```
```
stas00, line 398: is this really needed? The user has no idea what this code is and what to do with it, it should be enough to link to the code as you did for the few % who would want the details.

The article is already quite complicated and you're likely to lose the reader here.

hf-bitsandbytes-integration.md
younesbelkada: I have put `difference - value` in the description of the last column. I have left this column at the end so that readers can easily compare it against the standard errors and notice that it is lower

hf-bitsandbytes-integration.md
younesbelkada: I meant small / dummy. Let me modify it!

hf-bitsandbytes-integration.md
younesbelkada: Thanks for spotting this! 
Will also modify it on https://github.com/huggingface/transformers/pull/18631

hf-bitsandbytes-integration.md
stas00: if you feel it's better. I'd also replace `err` with `std err` - so that it better correlates with the text following it as you're referring to some std err but it's unclear which is that.

hf-bitsandbytes-integration.md
younesbelkada: I just refactored it a little bit (will push it on the next commit)

hf-bitsandbytes-integration.md
younesbelkada: Totally makes sense, I would also try to add a separator between each hardware setup! Thanks for providing the new table

hf-bitsandbytes-integration.md
stas00: Also, I have already heard from others on the team that said that 8bit implementation makes no speed impact, but clearly it is slower, so I think we should be very careful which words we use as people remember only parts...

hf-bitsandbytes-integration.md
younesbelkada: Yes actually. And I would also remove one `err` column and keep only the one in fp16 and rename it std err

hf-bitsandbytes-integration.md
younesbelkada: Yeah we clearly do not want to make false promises here, I will change it

hf-bitsandbytes-integration.md
younesbelkada: I removed it but kept the link to the code for those who are curious!

hf-bitsandbytes-integration.md
younesbelkada: Should be fixed in a5826d513c0906e90337f26d31e1e1c91b48e0ed

hf-bitsandbytes-integration.md
younesbelkada: Should be fixed in c71e04f6c5dd09bbbe0151e3643e334331c47b5c

hf-bitsandbytes-integration.md
younesbelkada: Should be fixed in 84fe2cbe6f80adc6de9f8922e0f2e26315d93dba

hf-bitsandbytes-integration.md
younesbelkada: added the small change in 84fe2cbe6f80adc6de9f8922e0f2e26315d93dba

hf-bitsandbytes-integration.md
younesbelkada: I just went to a more explicit explanation in c7495781e90be21d840f9350452ad027296a6e52 
I think that the readers needs to understand that the main goal of this method is not about speed but more about accessibility. Now people can get access to models that were not possible to access before due to hardware constraints (e.g. T5 on Google colab).

younesbelkada:
Just finished my round of editing @stas00 ! Please let me know if the proposed modifications make sense to you 🙏 
Regarding https://github.com/huggingface/blog/pull/463#discussion_r946058286 I will probably let Tim answer on that one! Again thank you for your feedback so far 🔥

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,460 @@
...
+                )
+    return model
+```
+
+This function recursively replaces all `nn.Linear` layers of a given model initialized on the `meta` device and replaces them with a `Linear8bitLt` module. The attribute `has_fp16_weights` has to be set to `False` in order to directly load the weights in `int8` together with the quantization statistics.
+
+We also discard the replacement for some modules (here the `lm_head`) since we want to keep the latest in their native precision for more precise and stable results.
+
+But it isn't over yet! The function above is executed under the `init_empty_weights` context manager which means that the new model will be still in the `meta` device.
+For models that are initialized under this context manager, `accelerate` will manually load the parameters of each module and sets it on the correct device.
```
stas00, line 374:
```suggestion
For models that are initialized under this context manager, `accelerate` will manually load the parameters of each module and set them on the correct devices.
```

hf-bitsandbytes-integration.md
stas00: yes, definitely! Perhaps it'd help to add a few bullet points at the beginning of the article to show what accomplishments this technique offers and the cost - e.g. save 2-4x memory at a cost of of a slowdown? Not sure what's a safe word to quantify the slowdown. Is 25% small? Perhaps just saying the actual "at a cost of 15-25% slowdown for large models"

stas00: Also I'm not quite sure it's very clear to the reader that the whole article is talking about the inference.

Related, I'm not sure the article title is enticing - because most users won't know that they should care about matmul speed. I'd drop it and say 8-bit inference instead. But that's just IMHO.

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,460 @@
...
+
+![out-quant.gif](assets/96_hf_bitsandbytes_integration/out-quant.gif)
+
+For an unsigned int8, we would subtract the minimum and scale by the absolute maximum. This is close to what zero-point quantization does. It's is similar to a min-max scaling but the latter maintains the value scales in such a way that the value “0” is always represented by an integer without any quantization error.
+
+These tricks can be combined in several ways, for example, row-wise or vector-wise quantization, when it comes to matrix multiplication for more accurate results. Looking at the matrix multiplication, A\*B=C, instead of regular quantization that normalize by a absolute maximum value per tensor, vector-wise quantization finds the absolute maximum of each row of A and each column of B. Then we normalize A and B by dividing these vectors. We then multiply A\*B to get C. Finally, to get back the FP16 values, we denormalize by computing the outer product of the absolute maximum vector of A and B. More details on this technique can be found in the LLM.int8() paper (XXX: link here).
```
younesbelkada, line 115:
```suggestion
These tricks can be combined in several ways, for example, row-wise or vector-wise quantization, when it comes to matrix multiplication for more accurate results. Looking at the matrix multiplication, A\*B=C, instead of regular quantization that normalize by a absolute maximum value per tensor, vector-wise quantization finds the absolute maximum of each row of A and each column of B. Then we normalize A and B by dividing these vectors. We then multiply A\*B to get C. Finally, to get back the FP16 values, we denormalize by computing the outer product of the absolute maximum vector of A and B. More details on this technique can be found in the [LLM.int8() paper](https://arxiv.org/abs/2208.07339).
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,460 @@
...
+2. Perform the matrix multiplication of the outliers in FP16 and the non-outliers in int8.
+3. Dequantize the non-outlier results and add both outlier and non-outlier results together to receive the full result in FP16.
+
+These steps can be summarized in the following animation:
+
+![Mixed-int8.gif](assets/96_hf_bitsandbytes_integration/Mixed-int8.gif)
+
+### The importance of outlier features
+
+A value that is outside the range of some numbers' global distribution is generally referred to as an outlier. Outlier detection has been widely used and covered in the current literature, and having prior knowledge of the distribution of your features helps with the task of outlier detection. More specifically, authors have observed that classic quantization at scale fails for transformer-based models >6B parameters. While large outlier features are also present in smaller models, the authors observe that a certain threshold these outliers from highly systematic patterns across transformers which are present in every layer of the transformer. For more details on these phenomena see the LLM.int8() paper.
```
younesbelkada, line 135:
```suggestion
A value that is outside the range of some numbers' global distribution is generally referred to as an outlier. Outlier detection has been widely used and covered in the current literature, and having prior knowledge of the distribution of your features helps with the task of outlier detection. More specifically, authors have observed that classic quantization at scale fails for transformer-based models >6B parameters. While large outlier features are also present in smaller models, the authors observe that a certain threshold these outliers from highly systematic patterns across transformers which are present in every layer of the transformer. For more details on these phenomena see the [LLM.int8() paper](https://arxiv.org/abs/2208.07339).
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,460 @@
...
+| hellaswag  | acc\_norm |           0.7274 |       0.7303 |         0.0044 | 0.0029 |
+| hellaswag  | acc       |           0.5563 |       0.5584 |           0.005 | 0.0021 |
+| piqa       | acc       |           0.7835 |       0.7884 |         0.0095 | 0.0049 |
+| piqa       | acc\_norm |           0.7922 |       0.7911 |         0.0095 | 0.0011 |
+| lambada    | ppl       |           3.9191 |        3.931 |      0.0846 | 0.0119 |
+| lambada    | acc       |           0.6808 |       0.6718 |            0.0065 |  0.009 |
+| winogrande | acc       |           0.7048 |       0.7048 |         0.0128 |      0 |
+
+
+We indeed observe 0 performance degradation for those models since the absolute difference of the metrics are all below the standard error (except for BLOOM-int8 which is slightly better than the native model on lambada). For a more detailed performance evaluation against state-of-the-art approaches, take a look at the paper!
```
younesbelkada, line 179:
```suggestion
We indeed observe 0 performance degradation for those models since the absolute difference of the metrics are all below the standard error (except for BLOOM-int8 which is slightly better than the native model on lambada). For a more detailed performance evaluation against state-of-the-art approaches, take a look at the [paper](https://arxiv.org/abs/2208.07339)!
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,460 @@
...
+
+![LLM](assets/96_hf_bitsandbytes_integration/LLM.png)
+
+Therefore, these models are hard to run on easily accessible devices. For example, just to do inference on BLOOM-175B, you would need to have 8x 80GB A100 GPUs (~$15k each). To fine-tune BLOOM-175B, you'd need 72 of these GPUs! Much larger models, like PaLM would require even more resources.
+
+Because these huge models require so many GPUs to run, we need to find ways to reduce these requirements while preserving the model's performance. Various technologies have been developed that try to shrink the model size, you may have heard of quantization and distillation, and there are many others.
+
+At Hugging Face and BigScience, after completing the training of BLOOM-176B, one of the approaches we took was to collaborate with `bitsandbytes` and integrate the technology described in the recent "LLM.int8(): 8-bit Matrix Multiplication for Transformers at Scale" paper into Hugging Face Transformers. We chose to integrate it since no post-training quantization is required to run this method, and you can reduce the memory footprint of any model by 2x by adding just a few lines of code.
```
younesbelkada, line 45:
```suggestion
At Hugging Face and BigScience, after completing the training of BLOOM-176B, one of the approaches we took was to collaborate with `bitsandbytes` and integrate the technology described in the recent ["LLM.int8(): 8-bit Matrix Multiplication for Transformers at Scale" paper](url) into Hugging Face Transformers. We chose to integrate it since no post-training quantization is required to run this method, and you can reduce the memory footprint of any model by 2x by adding just a few lines of code.
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,462 @@
...
+
+Because these huge models require so many GPUs to run, we need to find ways to reduce these requirements while preserving the model's performance. Various technologies have been developed that try to shrink the model size, you may have heard of quantization and distillation, and there are many others.
+
+After completing the training of BLOOM-176B, we at HuggingFace and BigScience were looking for ways to make this big model easier to run on less GPUs. Through our BigScience community we were made aware of research on Int8 inference that does not degrade predictive performance of large models and reduces the memory footprint of large models by a factor or 2x. Soon we started collaboring on this research which ended with a full integration into Hugging Face Transformers. With this blog post, we offer LLM.int8() integration for all Hugging Face Transformer models which we explain in more detail below. If you want to read more about our research, you can read our paper, [LLM.int8(): 8-bit Matrix Multiplication for Transformers at Scale](https://arxiv.org/abs/2208.07339).
```
younesbelkada, line 45:
```suggestion
After completing the training of BLOOM-176B, we at HuggingFace and BigScience were looking for ways to make this big model easier to run on less GPUs. Through our BigScience community we were made aware of research on Int8 inference that does not degrade predictive performance of large models and reduces the memory footprint of large models by a factor or 2x. Soon we started collaboring on this research which ended with a full integration into Hugging Face `transformers`. With this blog post, we offer LLM.int8() integration for all Hugging Face Transformer models which we explain in more detail below. If you want to read more about our research, you can read our paper, [LLM.int8(): 8-bit Matrix Multiplication for Transformers at Scale](https://arxiv.org/abs/2208.07339).
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,462 @@
+---
+title: "A Gentle Introduction to 8-bit Matrix Multiplication for transformers at scale using transformers, accelerate and bitsandbytes"
+thumbnail: /blog/assets/96_hf_bitsandbytes_integration/Thumbnail_blue.png
+---
+
+# A Gentle Introduction to 8-bit Matrix Multiplication for transformers at scale using Hugging Face Transformers, Accelerate and BitsAndBytes
```
younesbelkada, line 6:
```suggestion
# A Gentle Introduction to 8-bit Matrix Multiplication for transformers at scale using Hugging Face Transformers, Accelerate and bitsandbytes
```

_blog.yml
```diff
@@ -1132,3 +1131,13 @@
     - announcement
     - guide
 
+
+- local: hf-bitsandbytes-integration
+  title: "A Gentle Introduction to 8-bit Matrix Multiplication for transformers at scale using transformers, accelerate and bitsandbytes"
+  author: ybelkada
+  thumbnail: /blog/assets/96_hf_bitsandbytes_integration/thumbnail_blue.png
+  date: August 12, 2022
```
younesbelkada, line 1139:
```suggestion
  date: August 17, 2022
```

hf-bitsandbytes-integration.md
```diff
@@ -0,0 +1,462 @@
...
+
+Because these huge models require so many GPUs to run, we need to find ways to reduce these requirements while preserving the model's performance. Various technologies have been developed that try to shrink the model size, you may have heard of quantization and distillation, and there are many others.
+
+After completing the training of BLOOM-176B, we at HuggingFace and BigScience were looking for ways to make this big model easier to run on less GPUs. Through our BigScience community we were made aware of research on Int8 inference that does not degrade predictive performance of large models and reduces the memory footprint of large models by a factor or 2x. Soon we started collaboring on this research which ended with a full integration into Hugging Face `transformers`. With this blog post, we offer LLM.int8() integration for all Hugging Face Transformer models which we explain in more detail below. If you want to read more about our research, you can read our paper, [LLM.int8(): 8-bit Matrix Multiplication for Transformers at Scale](https://arxiv.org/abs/2208.07339).
```
younesbelkada, line 45:
```suggestion
After completing the training of BLOOM-176B, we at HuggingFace and BigScience were looking for ways to make this big model easier to run on less GPUs. Through our BigScience community we were made aware of research on Int8 inference that does not degrade predictive performance of large models and reduces the memory footprint of large models by a factor or 2x. Soon we started collaboring on this research which ended with a full integration into Hugging Face `transformers`. With this blog post, we offer LLM.int8() integration for all Hugging Face Transformer which we explain in more detail below. If you want to read more about our research, you can read our paper, [LLM.int8(): 8-bit Matrix Multiplication for Transformers at Scale](https://arxiv.org/abs/2208.07339).
```

younesbelkada:
Looks good to me now! 
cc @stevhliu I think that we are ready to merge now 💪

PR #481, opened 2022-08-20
Remove `time_emb` check

Mustapha-AJEGHRIR:
No need to check for `time_emb` when the condition is already True


osanseviero (approved):
Thanks a lot for the cleanup!

PR #492, opened 2022-08-23
Add HF int8 demo colab

mrm8488:


terarachang:
Thanks for the demo! 
But according to [the document](https://huggingface.co/docs/transformers/main/en/main_classes/model) about `int8_threshold `, _"Any hidden states value that is above this threshold will be considered an outlier and the operation on those values will be done in fp16"_. 
Correct me if I'm wrong, isn't _"Lowering the threshold means also less parameters in fp16 so breaking down the threshold to 0 leads to a full model in int8"_ (in your demo) contradicts this?

PR #494, opened 2022-08-25
Tutorial for training in Megatron-LM

loubnabnl:
This blog gives a gentle introduction to [Megatron-LM](https://github.com/NVIDIA/Megatron-LM), how to train a GPT2 model in it (to profit from the speedup) and convert the model to `transformers`

cc @lvwerra


megatron-training.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Why Megatron-LM?
+
+Before getting into the training details, let’s first understand what makes this framework more efficient than others. This section is inspired by this great [blog](https://huggingface.co/blog/bloom-megatron-deepspeed) about BLOOM training in [Megatron-DeepSpeed](https://github.com/bigscience-workshop/Megatron-DeepSpeed), please refer to it for more details as this blog is intended to give a gentle introduction to Megatron-LM.
+
+### DataLoader
+
+Megatron-LM comes with an efficient DataLoader, where the data is tokenized and shuffled before the training. It is also split into numbered sequences with indexes that are saved in a file, to avoid recomputing them in the training loop. The way the data is loaded makes the learning curve smooth and saves time during the training. 
+
+### Fused CUDA Kernels
```
lvwerra, line 43: what do you think about adding a small cartoon with before/after fusion? Just something very simple. I find visual elements always help a bit :)

megatron-training.md
```diff
@@ -0,0 +1,163 @@
...
+* Training
+* Model conversion to 🤗 Transformers
+
+## Why Megatron-LM?
+
+Before getting into the training details, let’s first understand what makes this framework more efficient than others. This section is inspired by this great [blog](https://huggingface.co/blog/bloom-megatron-deepspeed) about BLOOM training in [Megatron-DeepSpeed](https://github.com/bigscience-workshop/Megatron-DeepSpeed), please refer to it for more details as this blog is intended to give a gentle introduction to Megatron-LM.
+
+### DataLoader
+
+Megatron-LM comes with an efficient DataLoader, where the data is tokenized and shuffled before the training. It is also split into numbered sequences with indexes that are saved in a file, to avoid recomputing them in the training loop. The way the data is loaded makes the learning curve smooth and saves time during the training.
```
lvwerra, line 41: maybe also adding a figure here would help. In the book we made something like this but for this it could be much simpler: https://github.com/nlp-with-transformers/notebooks/blob/main/images/chapter10_ddp.png

also maybe spending 1-2 sentence more explaining why sequences are indexed and why that makes learning curves smooth would help.

megatron-training.md
```diff
@@ -0,0 +1,163 @@
...
+    <a href="/loubnabnl">
+        <img class="avatar avatar-user" src="https://avatars.githubusercontent.com/u/44069155?v=4" width="100" title="Gravatar">
+        <div class="bfc">
+            <code>loubnabnl</code>
+            <span class="fullname">Loubna Ben Allal</span>
+        </div>
+    </a>
+</div>
+
+In this tutorial, you will learn how to train a Language Model on NVIDIA GPUs with [Megatron-LM](https://github.com/NVIDIA/Megatron-LM), a powerful transformer model framework developed by the Applied Deep Learning Research team at NVIDIA, which gives [2x speedup](https://arxiv.org/pdf/2205.14135.pdf) compared to Hugging Face [Transformers](https://github.com/huggingface/transformers.git).
```
lvwerra, line 25: it be probably a bit more conservative and just say "which can give up to 2x speedup"

megatron-training.md
```diff
@@ -0,0 +1,163 @@
...
+<div class="author-card">
+    <a href="/loubnabnl">
+        <img class="avatar avatar-user" src="https://avatars.githubusercontent.com/u/44069155?v=4" width="100" title="Gravatar">
+        <div class="bfc">
+            <code>loubnabnl</code>
+            <span class="fullname">Loubna Ben Allal</span>
+        </div>
+    </a>
+</div>
+
```
lvwerra, line 24: What do you think about a entry hook like this:

> In the past few months several large language models have been released, usually with a mention of a mystical tool called Megatron-LM. It has become the de-facto framework to train enormous models efficiently across many GPUs but it is also very useful to pretrain smaller models.

Then the "In this tutorial"

_blog.yml
```diff
@@ -1188,3 +1188,12 @@
   date: August 24, 2022
   tags:
     - research
+
+- local: megatron-training
+  title: "Training a Language Model with Megatron-LM"
```
lvwerra, line 1193: I think "How to" blog posts worked quite well - maybe it's also just biased because those we done really well :) 

```suggestion
  title: "How to Train a Language Model with Megatron-LM"
```

megatron-training.md
```diff
@@ -0,0 +1,163 @@
...
+### Setup
+The easiest way to setup the environement is to pull an NVIDIA PyTorch Container that comes with all the required installations from [NGC](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/pytorch). See [documentation](https://docs.nvidia.com/deeplearning/frameworks/pytorch-release-notes/index.html) for more details. If you don't want to use this container you will need to install the latest pytorch, cuda, nccl, and NVIDIA [APEX](https://github.com/NVIDIA/apex#quick-start) releases and the `nltk` library.
+
+So after having installed Docker, you can run the container with the following command (`xx.xx` denotes your Docker version), and then clone [Megatron-LM repository](https://github.com/NVIDIA/Megatron-LM) inside it:
+```bash
+docker run --gpus all -it --rm nvcr.io/nvidia/pytorch:xx.xx-py3
+git clone https://github.com/NVIDIA/Megatron-LM
+```
+
+You also need to add the vocabulary file and merges table of your tokenizer inside Megatron-LM folder of your container. If you want to copy this data from outside the container you can use the following commands
```
lvwerra, line 62: Maybe we can explain where those can be found for a normal tokenizer and maybe point to the codeparrot repo for people who want to train their own tokenizer?

megatron-training.md
```diff
@@ -0,0 +1,163 @@
...
+
+```bash
+# to execute outside the container:
+mkdir -p nvidia/megatron-codeparrot-small
+# copy the weights from the container
+sudo docker cp CONTAINER_ID:/workspace/Megatron-LM/experiments/codeparrot-small/iter_0150000/mp_rank_00/model_optim_rng.pt nvidia/megatron-codeparrot-small
+git clone https://github.com/huggingface/transformers.git
+git clone https://github.com/NVIDIA/Megatron-LM.git
+export PYTHONPATH=Megatron-LM
+python transformers/src/transformers/models/megatron_gpt2/convert_megatron_gpt2_checkpoint.py nvidia/megatron-codeparrot-small/model_optim_rng.pt
```
lvwerra, line 157: I would add the steps to push the checkpoint afterwards to the Hub and maybe do an example at the end with the transformers pipeline generating something

megatron-training.md
```diff
@@ -0,0 +1,163 @@
...
+```
+Be careful, you will need to replace the generated vocabulary file and merges table after the conversion, with the original ones if you plan to load the tokenizer from there.
+
+Disclaimer: This framework adds some time overhead because of the extra preprocessing and conversion steps. So it is important that you decide for your case and given your model size which framework is more appropriate. But in general, for large models it's worth giving it a try.
```
lvwerra, line 161: Maybe we can add that for pretraining a model this might be a good option but just to fine-tune a medium sized model it might not be worth the extra effort?

lvwerra (commented):
Looks great! Left some comments, I can do a more detailed review in a second pass :)

megatron-training.md
loubnabnl: solved in https://github.com/huggingface/blog/pull/494/commits/f5272b6d4cd50200b04b2b050319a0eb54d4edaf

megatron-training.md
loubnabnl: solved in https://github.com/huggingface/blog/pull/494/commits/9c86dfd2a0c426b852e65568a32485841b0e97d1

megatron-training.md
loubnabnl: I have added some sentences to explain the indexing better. But I'm not sure what should be in the figure do you mean some indexes that get shuffled like the example of the list from 0 to 9 in [BLOOM's blog](https://huggingface.co/blog/bloom-megatron-deepspeed#datasets) ?

loubnabnl: I don't think data parallelism represents what makes Megatron-LM data loading different, as we also have that in the Trainer and Accelerate

megatron-training.md
```diff
@@ -0,0 +1,194 @@
...
+* Training
+* Model conversion to 🤗 Transformers
+
+## Why Megatron-LM?
+
+Before getting into the training details, let’s first understand what makes this framework more efficient than others. This section is inspired by this great [blog](https://huggingface.co/blog/bloom-megatron-deepspeed) about BLOOM training in [Megatron-DeepSpeed](https://github.com/bigscience-workshop/Megatron-DeepSpeed), please refer to it for more details as this blog is intended to give a gentle introduction to Megatron-LM.
+
+### DataLoader
+
+Megatron-LM comes with an efficient DataLoader, in which the data is tokenized and shuffled before the training. It is also split into numbered sequences with indexes that are saved in a file, so that they are computed only once. To make the indexes, the number of epochs is calculated based on the training parameters and an ordering for that number is created and then shuffled, unlike most cases where we iterate through the entire dataset until it is exhausted and then repeat. This smoothes the learning curve and saves time during the training.
```
lvwerra, line 41:
```suggestion
Megatron-LM comes with an efficient DataLoader where the data is tokenized and shuffled before the training. It is also splits the data into numbered sequences with indexes that are stored such that they need to be computed only once. To build the index, the number of epochs are computed based on the training parameters and an ordering is created and then shuffled. This is unlike most cases where we iterate through the entire dataset until it is exhausted and then repeat for the second epoch. This smoothes the learning curve and saves time during the training.
```

megatron-training.md
```diff
@@ -0,0 +1,194 @@
...
+
+### Fused CUDA Kernels
+
+In simple words, the idea of fused kernels here is that similar operations that are normally performed separately by Pytorch, are combined into a single hardware operation. So they reduce the number of memory movements done in multiple discrete computations by merging them into one. The figure below illustrates the idea of Kernel Fusion. It is inspired from this [paper](https://www.arxiv-vanity.com/papers/1305.1183/) which discusses the concept in detail.
```
lvwerra, line 45:
```suggestion
When a computation is run on the GPU the necessary data is fetched from memory then the computation is run and the result saved back into memory. In simple terms, the idea of fused kernels is that similar operations that are normally performed separately by Pytorch, are combined into a single hardware operation. So they reduce the number of memory movements done in multiple discrete computations by merging them into one. The figure below illustrates the idea of Kernel Fusion. It is inspired from this [paper](https://www.arxiv-vanity.com/papers/1305.1183/) which discusses the concept in detail.
```

megatron-training.md
```diff
@@ -0,0 +1,194 @@
...
+
+<p align="center">
+    <img src="assets/100_megatron_training/kernel_fusion.png" width="600" />
+</p>
+
+When f, g and h are fused in one kernel, the intermediary results x’ and y’ of f and g are stored in the GPU registers and immediately used by h. But without fusion,  x’ and y’ would need to be copied to the memory and then loaded by h. Therefore, Kernel Fusion gives a significant speed up to the computations.
+Megatron-LM also uses a Fused implementation of AdamW from [Apex](https://github.com/NVIDIA/apex) which is faster than the Pytorch implementation.
+
+While one can customize the DataLoader like Megatron-LM and use Apex’s Fused optimizer with `transformers`, it is hard to build Fused CUDA Kernels.
```
lvwerra, line 54:
```suggestion
While one can customize the DataLoader like Megatron-LM and use Apex’s Fused optimizer with `transformers`, it is not a beginner friendly undertaking to build custom Fused CUDA Kernels.
```

megatron-training.md
```diff
@@ -0,0 +1,194 @@
...
+
+So after having installed Docker, you can run the container with the following command (`xx.xx` denotes your Docker version), and then clone [Megatron-LM repository](https://github.com/NVIDIA/Megatron-LM) inside it:
+```bash
+docker run --gpus all -it --rm nvcr.io/nvidia/pytorch:xx.xx-py3
+git clone https://github.com/NVIDIA/Megatron-LM
+```
+
+You also need to add the vocabulary file and merges table of your tokenizer inside Megatron-LM folder of your container.  These files can be found in the model’s repository with the weights, see this [repository](https://huggingface.co/gpt2/tree/main) for GPT2. You can also train your own tokenizer using `transformers`. For an example, please refer to the [CodeParrot project](https://github.com/huggingface/transformers/tree/main/examples/research_projects/codeparrot).
```
lvwerra, line 69: Can you add the names of the files needed from the GPT2 repo. I think then it is super clear.

lvwerra, line 69:
```suggestion
You also need to add the vocabulary file and merges table of your tokenizer inside Megatron-LM folder of your container.  These files can be found in the model’s repository with the weights, see this [repository](https://huggingface.co/gpt2/tree/main) for GPT2. You can also train your own tokenizer using `transformers`. You can checkout the [CodeParrot project](https://github.com/huggingface/transformers/tree/main/examples/research_projects/codeparrot) for a practical example. 
```

megatron-training.md
```diff
@@ -0,0 +1,194 @@
...
+Now if you want to copy this data from outside the container you can use the following commands:
+```bash
+sudo docker cp vocab.json CONTAINER_ID:/workspace/Megatron-LM
+sudo docker cp merges.txt CONTAINER_ID:/workspace/Megatron-LM
+```
+
+### Data preprocessing
+In the rest of this tutorial we will be using [CodeParrot](https://huggingface.co/codeparrot/codeparrot-small) model and data as an example.
+
+The training data requires preprocessing. First, you need to convert it into a loose json format, with one json containing a text sample per line. If you're using 🤗 [Datasets](https://huggingface.co/docs/datasets/index), here is an example on how to do that (always inside Megatron-LM folder):
```
lvwerra, line 79:
```suggestion
The training data requires some preprocessing. First, you need to convert it into a loose json format, with one json containing a text sample per line. If you're using 🤗 [Datasets](https://huggingface.co/docs/datasets/index), here is an example on how to do that (always inside Megatron-LM folder):
```

megatron-training.md
```diff
@@ -0,0 +1,194 @@
...
+       --vocab vocab.json \
+       --dataset-impl mmap \
+       --tokenizer-type GPT2BPETokenizer \
+       --merge-file merges.txt \
+       --json-keys content \
+       --workers 32 \
+       --chunk-size 25 \
+       --append-eod
+```
+`workers` and `chunk_size` refer to the number of workers used in the preprocessing and the chunk size of data assigned to each one. `dataset-impl` refers to the implementation mode of the indexed datasets from ['lazy', 'cached', 'mmap'].
```
lvwerra, line 103:
```suggestion
The `workers` and `chunk_size` options refer to the number of workers used in the preprocessing and the chunk size of data assigned to each one. `dataset-impl` refers to the implementation mode of the indexed datasets from ['lazy', 'cached', 'mmap'].
```

megatron-training.md
```diff
@@ -0,0 +1,194 @@
...
+This outputs two files `codeparrot_content_document.idx` and `codeparrot_content_document.bin` which are used in the training.
+
+### Training
+You can configure the model architecture and training parameters as shown below, or put it in a bash script that you will run. This runs on 8 GPUs the 110M parameter CodeParrot pre-training. Note that the data is partitioned by default into a 969:30:1 ratio for training/validation/test sets.
```
lvwerra, line 107:
```suggestion
You can configure the model architecture and training parameters as shown below, or put it in a bash script that you will run. This command runs the pretraining on 8 GPUs for a 110M parameter CodeParrot model. Note that the data is partitioned by default into a 969:30:1 ratio for training/validation/test sets.
```

megatron-training.md
```diff
@@ -0,0 +1,194 @@
...
+        --pipeline-model-parallel-size 1 \
+        $GPT_ARGS \
+        --vocab-file $VOCAB_FILE \
+        --merge-file $MERGE_FILE \
+        --save $CHECKPOINT_PATH \
+        --load $CHECKPOINT_PATH \
+        --data-path $DATA_PATH \
+        $TENSORBOARD_ARGS
+```
+The training takes almost 12 hours in this setting.
```
lvwerra, line 153:
```suggestion
With this setup the training takes roughly 12 hours.
```

megatron-training.md
```diff
@@ -0,0 +1,194 @@
...
+
+### Converting the model to 🤗 Transformers
+After training we want to use the model in `transformers` e.g. to evaluate it. You can convert it to `transformers` following this [tutorial](https://huggingface.co/nvidia/megatron-gpt2-345m). For instance, after the training is finished you can copy the weights of the last iteration 150k and convert the `model_optim_rng.pt` file to a `pytorch_model.bin` file that is supported by `transformers`.
```
lvwerra, line 156:
```suggestion
After training we want to use the model in `transformers` e.g. for evaluation or to deploy it to production. You can convert it to a `transformers` model following this [tutorial](https://huggingface.co/nvidia/megatron-gpt2-345m). For instance, after the training is finished you can copy the weights of the last iteration 150k and convert the `model_optim_rng.pt` file to a `pytorch_model.bin` file that is supported by `transformers` with the following commands:.
```

megatron-training.md
```diff
@@ -0,0 +1,194 @@
...
+# to execute outside the container:
+mkdir -p nvidia/megatron-codeparrot-small
+# copy the weights from the container
+sudo docker cp CONTAINER_ID:/workspace/Megatron-LM/experiments/codeparrot-small/iter_0150000/mp_rank_00/model_optim_rng.pt nvidia/megatron-codeparrot-small
+git clone https://github.com/huggingface/transformers.git
+git clone https://github.com/NVIDIA/Megatron-LM.git
+export PYTHONPATH=Megatron-LM
+python transformers/src/transformers/models/megatron_gpt2/convert_megatron_gpt2_checkpoint.py nvidia/megatron-codeparrot-small/model_optim_rng.pt
+```
+Be careful, you will need to replace the generated vocabulary file and merges table after the conversion, with the original ones if you plan to load the tokenizer from there.
```
lvwerra, line 168:
```suggestion
Be careful, you will need to replace the generated vocabulary file and merges table after the conversion, with the original ones we introduced earlier if you plan to load the tokenizer from there.
```

megatron-training.md
```diff
@@ -0,0 +1,194 @@
...
+from transformers import pipeline
+
+pipe = pipeline("text-generation", model="your_username/codeparrot-small")
+outputs = pipe("def hello_world():")
+print(outputs[0]["generated_text"])
+```
+```
+ def hello_world():   
+    print("Hello World!")
+```
```
lvwerra, line 190: @LysandreJik You suggested to maybe show-case `accelerate` here. I am not exactly sure what you had in mind: could you elaborate?

megatron-training.md
```diff
@@ -0,0 +1,194 @@
...
+
+Disclaimer: Megatron-LM framework adds some time overhead because of the extra preprocessing and conversion steps. So it is important that you decide for your case and given your model size which framework is more appropriate. We recommend trying it for pre-training models, but probably not for fine-tuning medium-sized models.
```
lvwerra, line 192:
```suggestion
Disclaimer: We have shown that anybody can use Megatron-LM to train language models. The question is whether you should use it for all use-cases. Megatron-LM framework obviously adds some time overhead because of the extra preprocessing and conversion steps. So it is important that you decide for your case and given your model size which framework is more appropriate. We recommend trying it for pre-training models or extended fine-tuning, but probably not for shorter fine-tuning of medium-sized models.
```

lvwerra (approved):
Nice job! 🎉 I like the tutorial! I left a few suggestions, mainly text nits. Also there is an open question about using `accelerate` with @LysandreJik.

megatron-training.md
LysandreJik: I was thinking that given that the model is likely to be very large given it was trained with `MegatronLM`, we could take this as an opportunity to showcase how `accelerate` can be used to perform big model inference even on smaller hardware.

I think it could be just a single link pointing to accelerate, and a way to showcase how to use it here, which is likely a very simple change:

```suggestion
```python
from transformers import pipeline

pipe = pipeline("text-generation", model="your_username/codeparrot-small", low_cpu_mem_usage=True)
outputs = pipe("def hello_world():")
print(outputs[0]["generated_text"])
```
```
 def hello_world():   
    print("Hello World!")
```
```

cc @sgugger

loubnabnl: Thanks for the remark Lysandre! I think you didn’t correctly tag @sgugger 

We can add your suggestion, or it might also be interesting to showcase Big Model inference with weight dispatching. But I would probably need to add a paragraph about model parallelism because we are only using data parallelism in this example

LysandreJik: Understood! If it's not relevant here, no problem to keep the blogpost as it is

loubnabnl: I actually think it's relevant. I added a few sentences about model parallelism and accelerate big model inference it would be great if you can check them

loubnabnl:
@LysandreJik Can you please take a another look at the blog and merge it if it's ready (I don't have write access to the repo)

LysandreJik (approved):
Thank you, @loubnabnl

PR #498, opened 2022-08-26
Update stable_diffusion.md

perrozzi:
From the blog post is not clear that the command `with autocast("cuda"):` needs to be used with the float16 precision model, otherwise the image generation will fail.
Please consider this or similar change.

Regards and thanks a lot for the awesome model!
Luca

### DIFF
--- a/stable_diffusion.md
+++ b/stable_diffusion.md
@@ -125,12 +125,13 @@ from diffusers import StableDiffusionPipeline
 pipe = StableDiffusionPipeline.from_pretrained("CompVis/stable-diffusion-v1-4", revision="fp16", torch_dtype=torch.float16, use_auth_token=YOUR_TOKEN)
 ```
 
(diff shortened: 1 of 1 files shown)


pcuenca (commented):
Hi @perrozzi, thanks a lot for noticing this! And sorry for the delay, I initially missed the notification.

I like your suggestion, but I think it would be even better if we included a comment in the text itself, just before the example appears. Something along the lines of:

"The use of `torch.autocast("cuda")` is necessary when using half (`float16`) precision, and recommended for full precision as it will make inference faster."

What do you think?

perrozzi:
@pcuenca thanks for the feedback, modifier as suggested (sorry for the spam you might have received, had a major mess with git :-) )
let me know if you have other suggestions

pcuenca (approved):
Looks great, thanks!

PR #499, opened 2022-08-27
spaces_3dmoljs: update to 3.1.7 to make both spaces load

duerrsimon:


osanseviero (approved):
Thank you for the fix!

PR #500, opened 2022-08-29
Replaced the comma with the dot

kingabzpro:
Nothing much. I was reading the tutorial and found this typo.


osanseviero (approved):
Thanks for the fix!

PR #508, opened 2022-09-05
Add (not very useful) note about LaTeX support

julien-c:


osanseviero:
I would be against merging this :sweat_smile: unless you're really excited about merging it hehe

PR #509, opened 2022-09-05
Remove comment about new SD blog post

pcuenca:
A few days have passed, maybe we should remove that comment.


julien-c:
ok for me

natolambert:
Yeah looks good. This "new" blog post is becoming way bigger than the scope of just Stable Diffusion.

osanseviero (approved):
SG!

PR #513, opened 2022-09-08
Fix the blogpost link in the decision transformer notebook

RedTachyon:
The notebook/colab currently points to https://huggingface.co/blog/decision-transformers-train which does not exist. However, https://huggingface.co/blog/train-decision-transformers does exist, and it seems like that's what was intended, so I switched the link so that it works.


osanseviero (approved):
Thanks a lot!

PR #519, opened 2022-09-12
Move diffusers post assets to a dataset

osanseviero:
@julien-c do external urls work well for the thumbnail here?


julien-c:
i _think_ so, so i would still suggest leaving the thumbnail inside the repo for consistency with other posts..

osanseviero:
Alright, I reverted that

PR #507, opened 2022-09-05
Add a documentation on how to write an article

simoninithomas:
Hey there 👋 ,

So last time I replied to Carlos to help him create his first article and I think it's a good idea to have a small tutorial on how to create an article.

I put it on the documentation with 8 actions to do to create your article.

So the explanation is a first draft, I don't know if there is something to improve to be more comprehensible to everyone.

PR #521, opened 2022-09-15
Incredibly Fast BLOOM Inference with DeepSpeed and Accelerate

stas00:
as the server solutions may take some time to complete we agreed to split off the client-side solutions into its own post, so this is an extraction of https://github.com/huggingface/blog/pull/432 with a major update on the benchmarks and scripts.


bloom-inference-pytorch-scripts.md
```diff
@@ -0,0 +1,275 @@
...
+            <span class="fullname">Sylvain Gugger</span>
+        </div>
+    </a>
+</div>
+
+This article shows how to get an incredibly fast per token thoughput when generating with the 176B parameter [BLOOM model](https://huggingface.co/bigscience/bloom).
+
+As the model needs 352GB in bf16 (bfloat16) weights (`176*2`), the most efficient set-up is 8x80GB A100 GPUs. Also 2x8x40GB A100s or 2x8x48GB A6000 can be used. The main reason for using these GPUs is that at the time of this writing they provide the largest GPU memory, but other GPUs can be used as well. For example, 24x32GB V100s can be used.
+
+Using a single node will deliver the fastest througput since PCIe speed is typically much faster than inter-node network.
```
tjruwase, line 37: Do you mean NVLInk speed rather than PCIe speed?

bloom-inference-pytorch-scripts.md
```diff
@@ -0,0 +1,275 @@
...
+
+If you don't have that much hardware, it's still possible to run BLOOM inference on smaller GPUs, by using CPU or NVME offload, but of course, the generation time will be much slower.
+
+We are also going to cover the 8bit quantized solutions which require half the GPU memory at the cost of slightly slower throughput. We will discuss [BitsNBytes](https://github.com/TimDettmers/bitsandbytes) and [Deepspeed-Inference](https://www.deepspeed.ai/tutorials/inference-tutorial/) libraries there.
+
+## Benchmarks
+
+Without much further ado let's show some numbers.
```
philschmid, line 45: ado?

bloom-inference-pytorch-scripts.md
```diff
@@ -0,0 +1,275 @@
...
+
+This article shows how to get an incredibly fast per token thoughput when generating with the 176B parameter [BLOOM model](https://huggingface.co/bigscience/bloom).
+
+As the model needs 352GB in bf16 (bfloat16) weights (`176*2`), the most efficient set-up is 8x80GB A100 GPUs. Also 2x8x40GB A100s or 2x8x48GB A6000 can be used. The main reason for using these GPUs is that at the time of this writing they provide the largest GPU memory, but other GPUs can be used as well. For example, 24x32GB V100s can be used.
+
+Using a single node will deliver the fastest througput since PCIe speed is typically much faster than inter-node network.
+
+If you don't have that much hardware, it's still possible to run BLOOM inference on smaller GPUs, by using CPU or NVME offload, but of course, the generation time will be much slower.
+
+We are also going to cover the 8bit quantized solutions which require half the GPU memory at the cost of slightly slower throughput. We will discuss [BitsNBytes](https://github.com/TimDettmers/bitsandbytes) and [Deepspeed-Inference](https://www.deepspeed.ai/tutorials/inference-tutorial/) libraries there.
```
philschmid, line 41:
```suggestion
We are also going to cover the [8bit quantized solutions](https://huggingface.co/blog/hf-bitsandbytes-integration), which require half the GPU memory at the cost of slightly slower throughput. We will discuss [BitsNBytes](https://github.com/TimDettmers/bitsandbytes) and [Deepspeed-Inference](https://www.deepspeed.ai/tutorials/inference-tutorial/) libraries there.
```

bloom-inference-pytorch-scripts.md
```diff
@@ -0,0 +1,275 @@
...
+
+## Benchmarks
+
+Without much further ado let's show some numbers.
+
+For the sake of consistency, unless stated differently, the benchmarks in this article were all done on the same 8x80GB A100 node w/ 512GB of CPU memory on [Jean Zay HPC](http://www.idris.fr/eng/jean-zay/index.html). The JeanZay HPC users enjoy a very fast IO of about 3GB/s read speed (GPFS). This is important for checkpoint loading time. A slow disc will result in slow loading time. Especially since we are concurrently doing IO in multiple processes.
+
+All benchmarks are doing greedy generation of 100 token outputs:
```
philschmid, line 49:
```suggestion
All benchmarks are doing [greedy generation](https://huggingface.co/blog/how-to-generate#greedy-search) of 100 token outputs:
```

bloom-inference-pytorch-scripts.md
```diff
@@ -0,0 +1,275 @@
...
+```
+Generate args {'max_length': 100, 'do_sample': False}
+```
+The input prompt is comprised of just a few tokens. The previous token caching is on as well, as it'd be quite slow to recalculate them all the time.
+
+First, let's have a quick look at how long did it take to get ready to generate - i.e. how long did it take to load and prepare the data:
```
philschmid, line 55:
```suggestion
First, let's have a quick look at how long did it take to get ready to generate - i.e. how long did it take to load and prepare the model:
```
its the model not the data right?

bloom-inference-pytorch-scripts.md
```diff
@@ -0,0 +1,275 @@
...
+
+| project                 |      |
```
philschmid, line 57:
```suggestion
| project                 |   seconds   |
```

bloom-inference-pytorch-scripts.md
```diff
@@ -0,0 +1,275 @@
...
+| ds-inference shard-fp16 |   60 |
+| ds-inference unsharded  |  662 |
+| ds-zero                 |  462 |
+
+Deepspeed-Inference comes with pre-sharded weight repositories and there the loading takes about 1 minuted. Accelerate's loading time is excellent as well - at just about 2 minutes. The other solutions are much slower here.
+
+The loading time may or may not be of importance, since once loaded you can continually generate tokens again and again without an additional loading overhead.
+
+Next the most important benchmark of token generation thoughput. The throughput metric here is a simple - how long did it take to generate 100 new tokens divided by 100 and the batch size (i.e. divided by the total number of generated tokens).
+
```
philschmid, line 70: We should add a note/warning or explanation of how the numbers are calculated and that 0.69msec at bs 128 does not mean it took in a total of 0.69ms.

philschmid (commented):
Left some comments

bloom-inference-pytorch-scripts.md
stas00: that's if the user has NVLink. 

but you're making a good point - I will mention both.

Also perhaps mention NVSwitch which makes the internode connectivity as fast as intranode? perhaps this is out of scope.

bloom-inference-pytorch-scripts.md
stas00: It's an English idiom: https://www.merriam-webster.com/dictionary/without%20further%20ado

sgugger (approved):
Thanks for moving forward on this!

bloom-inference-pytorch-scripts.md
stas00: I'll change it to normal English ;)

bloom-inference-pytorch-scripts.md
stas00: Great suggestion, I've added:

> Let's revisit again how these numbers were calculated. To generate 100 new tokens for a batch size of 128 took 8.832 secs in real time. So now to calculate the throughput we did: `8832/(128*100) = 0.69`.

stas00: does that work?

philschmid: Awesome! Do you think we should add metrics so the formula? `8832msec/(128bs*100seq_len)`?

philschmid: 👍🏻

bloom-inference-pytorch-scripts.md
tjruwase: Is intranode using PCIe faster than inter-node using infiniband? Or is infiniband not considered for this?

stas00: The latest PCIe gens should be faster, but then there are so many different set-ups so it's super difficult to compare:

https://en.wikipedia.org/wiki/InfiniBand#Performance
https://en.wikipedia.org/wiki/PCI_Express#History_and_revisions

and if one uses NVSwitch shouldn't it make intra-node as fast as internode? I think it can connect up to 256 GPUs at the same speed as NVLInk - but that's Hopper.

Given that we are having this discussion I think it's the best to remove any specific suggesting and I'll just replace it with a very generic:

> Using a single node will typically deliver a fastest throughput since most of the time intra-node GPU linking hardware is faster than inter-node one, but it's not always the case.

_blog.yml
```diff
@@ -1223,4 +1223,15 @@
   date: September 7, 2022
   tags:
     - guide
-    - nlp
\ No newline at end of file
+    - nlp
+
+- local: bloom-inference-pytorch-scripts
```
julien-c, line 1228: i'm late to the party @stas00, but i would have favored a shorter URL, like `bloom-inference-pytorch` or even just `bloom-inference`

Let's keep that way now though! Just something to keep in mind

cc @osanseviero

philschmid: @julien-c AFAIK, there is another blog post in the making focusing on API I/O, I guess that's why they went with `*-pytorch-script`

osanseviero: https://github.com/huggingface/blog/pull/507 is adding some guidelines, and we'll iterate on it cc @simoninithomas

PR #526, opened 2022-09-16
Accelerate big models

sgugger:
This is a draft of a new blog post diving in the internals of Accelerate and how it powers big model inference.


accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+
+We call the checkpoints saved in several files like BLOOM *sharded checkpoints*, and we have standardized their format as such: one file (called `pytorch_model.bin.index.json`) contains some metadata and a map parameter name to file name, indicating where to find each weight, then the other files are standard PyTorch state dicts, they just contain a part of the model instead of the whole one. You can have a look at the content of the index file [here](https://huggingface.co/bigscience/bloom/blob/main/pytorch_model.bin.index.json).
+
+To load such a sharded checkpoint into a model, we just need to loop over the various shards. Accelerate provides a function called `load_checkpoint_in_model` that will do this for you if you have cloned one of the repos of the Hub, or you can directly use the `from_pretrained` method of Transformers, which will handle the downloading and caching for you.
```
muellerzr, line 199:
```suggestion
To load such a sharded checkpoint into a model, we just need to loop over the various shards. Accelerate provides a function called `load_checkpoint_in_model` that will do this for you if you have cloned one of the repos of the Hub, or you can directly use the `from_pretrained` method of Transformers, which will handle the downloading and caching for you:
```

accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+model = AutoModelForCausalLM.from_pretrained(checkpoint, device_map="auto")
+```
+
+If the device map computed automatically requires some weights to be offloaded on disk because you don't have enough GPU and CPU RAM, you will get an error indicating you need to pass an folder where the weights that should be stored on disk will be offloaded.
+
+```python
+from transformers import AutoModelForCausalLM
+
+checkpoint = "facebook/opt-2.7b"
+model = AutoModelForCausalLM.from_pretrained(checkpoint, device_map="auto", offloag_folder="offload")
```
muellerzr, line 214:
```suggestion
model = AutoModelForCausalLM.from_pretrained(checkpoint, device_map="auto", offload_folder="offload")
```

accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+To load such a sharded checkpoint into a model, we just need to loop over the various shards. Accelerate provides a function called `load_checkpoint_in_model` that will do this for you if you have cloned one of the repos of the Hub, or you can directly use the `from_pretrained` method of Transformers, which will handle the downloading and caching for you.
+
+```python
+from transformers import AutoModelForCausalLM
+
+checkpoint = "facebook/opt-2.7b"
+model = AutoModelForCausalLM.from_pretrained(checkpoint, device_map="auto")
+```
+
+If the device map computed automatically requires some weights to be offloaded on disk because you don't have enough GPU and CPU RAM, you will get an error indicating you need to pass an folder where the weights that should be stored on disk will be offloaded.
```
muellerzr, line 208: Perhaps include a quick example of the error message?

accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+
+```python
+from transformers import AutoModelForCausalLM
+
+checkpoint = "facebook/opt-2.7b"
+model = AutoModelForCausalLM.from_pretrained(checkpoint, device_map="auto", offloag_folder="offload")
+```
+
+Note that if you are trying to load a very large model that require some disk offload on top of CPU offload, you might get out of RAM when the last shards of the checkpoint are loaded, since there is the part of the model staying on CPU taking space. If that is the case, use the option `offload_state_dict=True` to temporarily offload the part of the model staying on CPU while the weights are all loaded, and reload it in RAM once all the weights have been processed
```
muellerzr, line 217:
```suggestion
Note that if you are trying to load a very large model that require some disk offload on top of CPU offload, you might run out of RAM when the last shards of the checkpoint are loaded, since there is the part of the model staying on CPU taking space. If that is the case, use the option `offload_state_dict=True` to temporarily offload the part of the model staying on CPU while the weights are all loaded, and reload it in RAM once all the weights have been processed
```

accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+
+checkpoint = "facebook/opt-11b"
+model = AutoModelForCausalLM.from_pretrained(
+    checkpoint, device_map="auto", offload_folder="offload", offload_state_dict = True
+)
+```
+
+## Running a model split on several devices
+
+One last part we haven't touched is how Accelerate enables your model to run with its weight spread across several GPUs, CPU RAM and the disk folder. This is done very simply, using hooks. PyTorch provides an API to add functions executed just before each forward called [hooks](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_forward_hook).
```
muellerzr, line 231:
```suggestion
One last part we haven't touched is how Accelerate enables your model to run with its weight spread across several GPUs, CPU RAM, and the disk folder. This is done very simply using hooks. 

> [hooks](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_forward_hook) are a PyTorch API that adds functions executed just before each forward called 
```

accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+- make sure all the inputs of the module are on the same device as the weights;
+- if the weights have been offloaded to the CPU, move them to GPU 0 before the forward pass and back to the CPU just after;
+- if the weights have been offloaded to disk, load them in RAM then on the GPU 0 before the forward pass and free this memory just after.
+
+The whole process is summarized in the following video:
+
+<iframe width="560" height="315" src="https://www.youtube.com/embed/MWCSGj9jEAo" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+
+This way, your model can be loaded and run even if you don't have enough GPU RAM and CPU RAM. The only thing you need is disk space (and lots of patience!) While this solution is pretty naive if you have multiple GPUs (there is no clever pipeline parallelism involved, just using the GPUs sequentially) it still yield [pretty decent results for BLOOM](https://huggingface.co/blog/bloom-inference-pytorch-scripts). And it allows you to run the model on smaller setups (albeit more slowly).
```
muellerzr, line 242: Towards the end perhaps add links to the Accelerate docs and PyTorch docs over the relevant areas?

muellerzr (approved):
Great work! Added a number of suggestions for improvements and typo fixes (feel free to ignore what you don't like). 

*fantastic* article!

accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+        <div class="bfc">
+            <code>sgugger</code>
+            <span class="fullname">Sylvain Gugger</span>
+        </div>
+    </a>
+</div>
+
+## Load and run large models
+
+Meta AI and BigScience recently open-sourced very large language models. You can use them even if you don't own a supercomputer by just running the following code:
```
LysandreJik, line 29: I'd personally err towards having a slightly longer introduction before diving into the code; something about models becoming larger and larger, about consumer hardware not being enough to run them out of the box, and about tools like accelerate making it possible to try even the largest models without access to a cluster.

accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+
+```python
+import torch
+from transformers import pipeline
+
+# This works on Colab. Pick a larger checkpoint if you have time to wait and enough disk space!
+checkpoint = "facebook/opt-2.7b"
+generator = pipeline("text-generation", model=checkpoint, device_map="auto", torch_dtype=torch.float16)
+generator("More and more large language models are opensourced so Hugging Face has")
```
LysandreJik, line 38: I'd add a line return for readability

LysandreJik:
```suggestion

# Perform inference
generator("More and more large language models are opensourced so Hugging Face has")
```

accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+We'll explain what each of those arguments does in a moment, but first just consider the traditional model loading pipeline in PyTorch: it usually consists of:
+
+1. Create the model
+2. Load in memory its weights (in an object usually called `state_dict`)
+3. Load those weights in the created model
+4. Move the model on the device for inference
+
+While that has worked pretty well in the past years, very large models make this approach challenging. Here the model picked has 2.7 billion parameters. In the default precision, it means that just step 1 (creating the model) will take roughly 10.8GB in RAM (1 parameter in float32 takes 4 bytes in memory). This is enough to fill almost all the RAM available in Colab.
+
+Then step 2 will load in memory a second copy of the model (so another 10.8GB in RAM in default precision) which will get you out of RAM on Colab for the 2.7B model. If you were trying to load BLOOM or OPT-176B, like this, you would need 1.4 **terabytes** of CPU RAM. That is a bit excessive! And all of this to just move the model on one (or several) GPU(s) at step 4.
```
LysandreJik, line 50:
```suggestion
Then step 2 will load in memory a second copy of the model (so another 10.8GB in RAM in default precision) which will get you out of RAM on Colab for the 2.7B model. If you were trying to load the largest models, for example BLOOM or OPT-176B (which both have 176 billion parameters), like this, you would need 1.4 **terabytes** of CPU RAM. That is a bit excessive! And all of this to just move the model on one (or several) GPU(s) at step 4.
```

accelerate-big-models.md
LysandreJik: I'd also specify that what lies below is focused on inference, as "runs" can mean many things

LysandreJik (approved):
Very nice! Only left a few comments. Before the conclusion, I'd likely show an inference example. Also I'd maybe show an example of different model sizes on colab, to show how the disk offload is used when models start to become larger. I'd take that opportunity to mention how this can significantly slow things down, to mention that a very large RAM-capable machine is always going to be faster than a machine with a very large disk.

Great article!

accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+
+Clearly we need something smarter. In this blog post, we'll explain how Accelerate leverages PyTorch features to load and run inference with very large models, even if they don't fit in RAM or one GPU. In a nutshell, it changes the process above like this:
+
+1. Create an empty (e.g. without weights) model
+2. Decide where each layer is going to go (when multiple devices are available)
+3. Load in memory parts of its weights
+4. Load those weights in the empty model
+5. Move the weights on the device for inference
+6. Repeat from step 3 for the next weights until all the weights are loaded
```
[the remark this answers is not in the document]
sgugger: They stay there (device might be GPU/CPU/disk)

accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+1. Create the model
+2. Load in memory its weights (in an object usually called `state_dict`)
+3. Load those weights in the created model
+4. Move the model on the device for inference
+
+While that has worked pretty well in the past years, very large models make this approach challenging. Here the model picked has 2.7 billion parameters. In the default precision, it means that just step 1 (creating the model) will take roughly 10.8GB in RAM (1 parameter in float32 takes 4 bytes in memory). This is enough to fill almost all the RAM available in Colab.
+
+Then step 2 will load in memory a second copy of the model (so another 10.8GB in RAM in default precision) which will get you out of RAM on Colab for the 2.7B model. If you were trying to load BLOOM or OPT-176B, like this, you would need 1.4 **terabytes** of CPU RAM. That is a bit excessive! And all of this to just move the model on one (or several) GPU(s) at step 4.
+
+Clearly we need something smarter. In this blog post, we'll explain how Accelerate leverages PyTorch features to load and run inference with very large models, even if they don't fit in RAM or one GPU. In a nutshell, it changes the process above like this:
```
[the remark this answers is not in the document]
sgugger: Not at this stage IMO.

accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+```python
+from accelerate import init_empty_weights
+from transformers import AutoConfig, AutoModelForCausalLM
+
+config = AutoConfig.from_pretrained("bigscience/bloom")
+with init_empty_weights():
+    model = AutoModelForCausalLM.from_config(config)
+```
+
+This works on any model, but will leave you a shell you can't use directly: some operations are implemented for the meta device, but not all yet. Here for instance, you can use the `large_model` defined above with an input, but not the BLOOM model. Even when using it, the output will be a tensor of the meta device, so you will get the shape of the result, but nothing more.
```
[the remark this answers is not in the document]
sgugger: rephrased differently

accelerate-big-models.md
```diff
@@ -0,0 +1,242 @@
...
+ 'decoder.block.4.layer.1.EncDecAttention.v': 0,
+ 'decoder.block.4.layer.1.EncDecAttention.o': 'cpu',
+ 'decoder.block.4.layer.1.layer_norm': 'cpu',
+ 'decoder.block.4.layer.1.dropout': 'cpu',
+ 'decoder.block.4.layer.2': 'cpu',
+ 'decoder.block.5': 'cpu',
+ ...
+```
+
+Accelerate evaluated that the shared layer, the encoder and the decoder up until the 4th block could all fit on GPU 0, then part of the 4th block needs to be on the CPU, as well as the following weights.
```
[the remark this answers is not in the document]
sgugger: I don't think we care very much ;-p

accelerate-big-models.md
```diff
@@ -0,0 +1,327 @@
+---
+title: "How 🤗 Accelerate runs very large models"
+thumbnail: /blog/assets/20_accelerate_library/accelerate_diff.png
+---
+
+<h1>
+    How 🤗 Accelerate runs very large models
```
julien-c, line 7:
```suggestion
    How 🤗 Accelerate runs very large models thanks to PyTorch
```

_blog.yml
```diff
@@ -1252,3 +1252,12 @@
   tags:
     - research
     - nlp
+
+- local: accelerate-big-models
```
julien-c, line 1256: should it be big or large here?

julien-c (commented):
i'm wondering if we should be consistent with the big and large. Other than that, looks great!

_blog.yml
sgugger: Done

_blog.yml
```diff
@@ -1252,3 +1252,12 @@
   tags:
     - research
     - nlp
+
+- local: accelerate-large-models
+  title: "How 🤗 Accelerate runs very large models thanks to PyTorch"
+  author: sgugger
+  thumbnail: /blog/assets/104_accelerate-large-models/thumbnail.png
+  date: September 26, 2022
```
osanseviero, line 1260: To update

_blog.yml
```diff
@@ -1252,3 +1252,12 @@
...
+  tags:
+    - research
+    - open-source-collab
```
osanseviero, line 1263: Maybe `guide` as well

osanseviero (approved):
I didn't review the content, but the structure/ymls look good!

PR #529, opened 2022-09-20
Add more DS-Inference references

philschmid:
As suggested, I added links to the two blogs I have created.


stas00 (approved):
Thank you, Phlipp!

PR #541, opened 2022-09-26
Fix SetFit publication date

lewtun:


_blog.yml
```diff
@@ -1246,7 +1246,7 @@
     
 - local: setfit
   title: "SetFit: Efficient Few-Shot Learning Without Prompts"
-  author: Unso, lewtun
+  author: Unso, lewtun, luketheduke
```
julien-c, line 1249: multiple authors are not supported here (ie the link links to `https://huggingface.co/Unso,%20lewtun,%20luketheduke` not a valid user page)

julien-c: (fixed in #542)

PR #543, opened 2022-09-27
Fix assets for Accelerate large models

sgugger:
Move the folder in the assets subfolder :grimacing:


osanseviero (approved):
Thanks for the quick fix!

PR #545, opened 2022-09-27
[DO NOT MERGE, BUT DO REVIEW!] HF x PyTorch DDP Blog

muellerzr:
This is a **draft** collaboration blog between Hugging Face and PyTorch on Distributed Data Parallelism. It doesn't 100% fit all of the usual blog styling nits that HF blogs would because we are still looking into whether it will be published only on PyTorch, HF then linked, etc. So for now it doesn't have any of our custom watermarkings etc. 

Sending this off to review with our team before passing it along further to the folks at torch to have their crack at edits etc. Want to make sure it's well and polished before then 😉

### DIFF
--- a/pytorch-ddp-accelerate-transformers.md
+++ b/pytorch-ddp-accelerate-transformers.md
@@ -230,7 +230,7 @@ Next let's talk about how Accelerate can help. There's a few issues with the abo
 1. This is slightly inefficient, given that `n` dataloaders are made based on each device and pushed.
 2. This code will **only** work for multi-GPU, so special care would need to be made for it to be ran on a single node again, or on TPU.
 
-Accelerate helps this through the `Accelerator` class. Through it, the code remains much the same except for three lines of code when comparing a single node to multinode, as shown below:
+Accelerate helps this through the [`Accelerator`](https://huggingface.co/docs/accelerate/v0.12.0/en/package_reference/accelerator#accelerator) class. Through it, the code remains much the same except for three lines of code when comparing a single node to multinode, as shown below:
 
 ```python
 def train_ddp_accelerate():
@@ -279,7 +279,7 @@ def train_ddp_accelerate():
     print(f'Accuracy: {100. * correct / len(test_loader.dataset)}')
 ```
 
-Now any and all custom code needed to launch your code on any distributed setup is simplified to the `Accelerator` object. This code can then still be ran through the `torchrun` CLI, or through Accelerate's own CLI interface, `accelerate launch`.
+Now any and all custom code needed to launch your code on any distributed setup is simplified to the `Accelerator` object. This code can then still be ran through the `torchrun` CLI, or through Accelerate's own CLI interface, [`accelerate launch`](https://huggingface.co/docs/accelerate/v0.12.0/en/basic_tutorials/launch).
 
 As a result its now trivialized to perform distributed training with Accelerate and keeping as much of the barebones PyTorch code the same as possible.
 
@@ -287,7 +287,7 @@ Earlier it was mentioned that Accelerate also makes the DataLoaders more efficie
 
 ### Using the `notebook_launcher`
 
-Earlier it was mentioned you can start distributed code directly out of your Jupyter Notebook. This comes from Accelerate's `notebook_launcher` utility, which allows for starting multi-gpu training based on code inside of a Jupyter Notebook.
+Earlier it was mentioned you can start distributed code directly out of your Jupyter Notebook. This comes from Accelerate's [`notebook_launcher`](https://huggingface.co/docs/accelerate/v0.12.0/en/basic_tutorials/notebook) utility, which allows for starting multi-gpu training based on code inside of a Jupyter Notebook.
 
 To use it is as trivial as importing the launcher:
 
@@ -309,7 +309,7 @@ notebook_launcher(train_accelerate_ddp, args=(), num_processes=2)
 
 ## Using 🤗 Trainer
 
-Finally, we arrive at the highest level of API -- the Hugging Face Trainer.
+Finally, we arrive at the highest level of API -- the Hugging Face [Trainer](https://huggingface.co/docs/transformers/main_classes/trainer).
 
 This wraps as much training as possible while still being able to train on distributed systems without the user needing to do anything at all.
 
--- a/pytorch-ddp-accelerate-transformers.md
+++ b/pytorch-ddp-accelerate-transformers.md
@@ -1,0 +1,427 @@
+# From PyTorch DDP to Accelerate to Trainer, mastery of distributed training with ease
+
+Authors:
+
+- Zachary Mueller (Hugging Face)
+- Less Wright (PyTorch)
+
+## General Overview
+
+This tutorial assumes you have a basic understanding of PyTorch and how to train a very basic model. It will showcase training on multiple GPUs through a process called Distributed Data Parallelism (DDP) through three different layers:
+
+- Native PyTorch DDP through the `pytorch.distributed` module
+- Utilizing 🤗Accelerate's light wrapper around `pytorch.distributed` that also helps ensure the code can be run on a single GPU and TPUs with zero code changes and miminimal code changes to the original code
+- Utilizing 🤗Transformer's trainer, which is a high level wrapping API to perform a similar result to Accelerate
+
+## What is "Distributed" training and why does it matter?
+
+Take some very basic PyTorch training code below, which sets up and trains a model on MNIST based on the [official MNIST example](https://github.com/pytorch/examples/blob/main/mnist/main.py)
+
+
+```python
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+from torchvision import datasets, transforms
+
+class BasicNet(nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.conv1 = nn.Conv2d(1, 32, 3, 1)
+        self.conv2 = nn.Conv2d(32, 64, 3, 1)
+        self.dropout1 = nn.Dropout(0.25)
+        self.dropout2 = nn.Dropout(0.5)
+        self.fc1 = nn.Linear(9216, 128)
+        self.fc2 = nn.Linear(128, 10)
+        self.act = F.relu
+
+    def forward(self, x):
+        x = self.act(self.conv1(x))
+        x = self.act(self.conv2(x))
+        x = F.max_pool2d(x, 2)
+        x = self.dropout1(x)
+        x = torch.flatten(x, 1)
+        x = self.act(self.fc1(x))
+        x = self.dropout2(x)
+        x = self.fc2(x)
+        output = F.log_softmax(x, dim=1)
+        return output
+```
+
+We define the training device (`cuda`):
+
+```python
+device = "cuda"
+```
+
+Build some PyTorch DataLoaders:
+
+```python
+transform = transforms.Compose([
+    transforms.ToTensor(),
+    transforms.Normalize((0.1307), (0.3081))
+])
+
+train_dset = datasets.MNIST('data', train=True, download=True, transform=transform)
+test_dset = datasets.MNIST('data', train=False, transform=transform)
+
+train_loader = torch.utils.data.DataLoader(train_dset, shuffle=True, batch_size=64)
+test_loader = torch.utils.data.DataLoader(test_dset, shuffle=False, batch_size=64)
+```
+
+Move the model to the CUDA device:
+
+```python
+model = BasicNet().to(device)
+```
+
+Build a PyTorch optimizer:
+
+```python
+optimizer = optim.AdamW(model.parameters(), lr=1e-3)
+```
+
+Before finally creating a simplistic training and evaluation loop that performs one full iteration over the dataset and calculates the test accuracy:
+
+```python
+model.train()
+for batch_idx, (data, target) in enumerate(train_loader):
+    data, target = data.to(device), target.to(device)
+    output = model(data)
+    loss = F.nll_loss(output, target)
+    loss.backward()
+    optimizer.step()
+    optimizer.zero_grad()
+model.eval()
+correct = 0
+with torch.no_grad():
+    for data, target in test_loader:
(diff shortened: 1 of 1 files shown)


pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+            output = model(data)
+            pred = output.argmax(dim=1, keepdim=True)
+            correct += pred.eq(target.view_as(pred)).sum().item()
+    print(f'Accuracy: {100. * correct / len(test_loader.dataset)}')
+```
+
+Next let's talk about how Accelerate can help. There's a few issues with the above code:
+
+1. This is slightly inefficient, given that `n` dataloaders are made based on each device and pushed.
+2. This code will **only** work for multi-GPU, so special care would need to be made for it to be ran on a single node again, or on TPU.
```
sgugger, line 231: Ah, it's mentioned here. Might be enough!

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+
+Accelerate helps this through the [`Accelerator`](https://huggingface.co/docs/accelerate/v0.12.0/en/package_reference/accelerator#accelerator) class. Through it, the code remains much the same except for three lines of code when comparing a single node to multinode, as shown below:
+
+```python
+def train_ddp_accelerate():
+    accelerator = Accelerator(device_placement=True)
```
sgugger, line 237:
```suggestion
    accelerator = Accelerator()
```
This is the default.

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+
+This wraps as much training as possible while still being able to train on distributed systems without the user needing to do anything at all.
+
+First we need to import the Trainer:
+
+```python
+from transformers import Trainer
+```
+
+Then we define some TrainingArguments since it will handle the dataloader creation. The trainer also works through dictionaries, so a custom colalte function needs to be made.
```
sgugger, line 322:
```suggestion
Then we define some `TrainingArguments` to control all the usual hyper-parameters. The trainer also works through dictionaries, so a custom collate function needs to be made.
```

sgugger (approved):
Thanks for drafting this! Left a couple of comments.

pytorch-ddp-accelerate-transformers.md
muellerzr: It is, but also in the docs where we assume no prior knowledge (in our docs) we have `device_placement=True`. 

Leave the default and make a note about device placement probably then?

sgugger: Yeah, we can also update the docs since it was originally set like this to show how to remove the automatic device placement but no one ever complained about it.

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+    dist.destroy_process_group()
+```
+
+The last piece of the puzzle is *how do I send my data and model to another GPU?*
+
+This is where the `DistributedDataParallel` module comes into play. It will copy your model into each GPU, and when `loss.backward()` is called the gradients across all these copies of the model will be averaged, backprop is performed, and each of the weights get updated to have the same weights across all devices.
+
+Below is an example of our training setup, refactored as a function, with this capability:
+
+> Note: Here rank is the overall rank of the GPU based on all other devices, meaning they have a rank of `0 -> n`
```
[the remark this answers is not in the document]
muellerzr: Rewording more accurately:

> Note: Here rank is the overall rank of the current GPU compared to all the other GPUs available, meaning they have a rank of `0 -> n-1`

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+
+The optimizer needs to be declared based on the model *on the specific device* for all of the gradients to properly be calculated.
+
+Lastly, to run the script PyTorch has a convient `torchrun` command line module that can help. Just pass in the number of nodes it should use as well as the script to run and you are set:
+
+```bash
+torchrun --nproc_per_nodes=2 --nnodes=1 example_script.py
+```
+
+The above will run the training script on two GPUs that live on a single machine and this is the barebones for performing distributed training with PyTorch.
```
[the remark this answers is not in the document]
muellerzr: Went in the middle with the below clarification, phrasing as:
"The above will run the training script on two GPUs that live on a single machine and this is the barebones for performing only distributed training with PyTorch."

(Aka adding **only**)

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+    dist.init_process_group("gloo", rank=rank, world_size=world_size)
+
+def cleanup():
+    "Cleans up the distributed environment"
+    dist.destroy_process_group()
+```
+
+The last piece of the puzzle is *how do I send my data and model to another GPU?*
+
+This is where the `DistributedDataParallel` module comes into play. It will copy your model into each GPU, and when `loss.backward()` is called the gradients across all these copies of the model will be averaged, backprop is performed, and each of the weights get updated to have the same weights across all devices.
```
pacman100, line 140:
```suggestion
This is where the `DistributedDataParallel` module comes into play. It will copy your model onto each GPU, and when `loss.backward()` is called the backpropagation is performed and the resulting gradients across all these copies of the model will be averaged/reduced. This ensures each device has the same weights post the optimizer step.
```

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+    model.train()
+    for batch_idx, (data, target) in enumerate(train_loader):
+        data, target = data.to(device), target.to(device)
+        output = model(data)
+        loss = F.nll_loss(output, target)
+        loss.backward()
+        optimizer.step()
+        optimizer.zero_grad()
+    cleanup()
+```
```
pacman100, line 164: I believe changes that are necessary for dataloaders are left out. That is a significant detail and quite a bit of a hassle.  Accelerate abstracts and that is great for users. So, good to detail this too. For normal datasets, that would mean using the `DistributedSampler` sampler https://pytorch.org/docs/stable/data.html#torch.utils.data.distributed.DistributedSampler.

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+    train_loader, test_loader, model, optimizer = accelerator.prepare(
+        train_loader, test_loader, model, optimizer
+    )
+
+    # Train for a single epoch
+    model.train()
+    for batch_idx, (data, target) in enumerate(train_loader):
+        output = model(data)
+        loss = F.nll_loss(output, target)
+        loss.backward()
```
pacman100, line 266:
```suggestion
        accelerator.backward()
```

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+        optimizer.step()
+        optimizer.zero_grad()
+    
+    # Evaluate
+    model.eval()
+    correct = 0
+    with torch.no_grad():
+        for data, target in test_loader:
+            data, target = data.to(device), target.to(device)
```
pacman100, line 275:
```suggestion
```

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+    for batch_idx, (data, target) in enumerate(train_loader):
+        data, target = data.to(device), target.to(device)
+        output = model(data)
+        loss = F.nll_loss(output, target)
+        loss.backward()
+        optimizer.step()
+        optimizer.zero_grad()
+    cleanup()
+```
+
```
pacman100, line 165: Also, details on handling outputs during evaluation, i.e, gathering from all the processes is also quite a bit of hassle and an important detail when woking in distributed setting else evaluations won't be reliable hampering the experimentation process. Accelerate abstracts this too making life simple for users.
Without Accelerate one would have code like below:
```
gather_predictions = [predictions.clone() for _ in range(dist.get_world_size())]
torch.distributed.all_gather(gather_predictions, predictions)
batch_predictions =  torch.cat(gather_predictions, dim=0)

gather_labels = [batch["labels"].clone() for _ in range(dist.get_world_size())]
torch.distributed.all_gather(gather_labels, batch["labels"])
batch_labels =  torch.cat(gather_labels, dim=0)

metric.add_batch(
    predictions=batch_predictions.detach().cpu().numpy(),
    references=batch_labels.detach().cpu().numpy(),
)

```

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+torchrun --nproc_per_nodes=2 --nnodes=1 example_script.py
+```
+
+The above will run the training script on two GPUs that live on a single machine and this is the barebones for performing only distributed training with PyTorch.
+
+Now let's talk about Accelerate, a library aimed to make this process more seameless and also help with a few best practices
+
+## 🤗 Accelerate
+
+[Accelerate](https://huggingface.co/docs/accelerate) is a library designed to allow you to perform what we just did above, without needing to modify your code greatly. On top of this, the data pipeline innate to Accelerate can also improve performance to your code as well.
```
pacman100, line 180: I believe mentioning the abstractions for dataloaders and metrics gathering is also important in addition to support for various device types and model related changes in distributed setup.

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+            data, target = data.to(device), target.to(device)
+            output = model(data)
+            pred = output.argmax(dim=1, keepdim=True)
+            correct += pred.eq(target.view_as(pred)).sum().item()
+    print(f'Accuracy: {100. * correct / len(test_loader.dataset)}')
+```
+
+Now any and all custom code needed to launch your code on any distributed setup is simplified to the `Accelerator` object. This code can then still be ran through the `torchrun` CLI, or through Accelerate's own CLI interface, [`accelerate launch`](https://huggingface.co/docs/accelerate/v0.12.0/en/basic_tutorials/launch).
+
+As a result its now trivialized to perform distributed training with Accelerate and keeping as much of the barebones PyTorch code the same as possible.
```
pacman100, line 284: Needs rewording

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+
+Earlier it was mentioned that Accelerate also makes the DataLoaders more efficient. This is through custom Samplers that will send parts of the batches automatically to different devices during training allowing for a single copy of the data to be known at one time, rather than four at once into memory.
```
pacman100, line 286: I think here the important detail is that the user doesn't have to worry about making sure each device has a subset of the original dataset that is exclusive to it. Everything is handled under the hood. Also, single copy vs multiple copies isn't accurate because `DataLoaderDispatcher` (in case of IterableDataset) sends num_processes*batch_size data from rank 0 to all ranks leading to data deduplication/multiple copies wherein each rank then picks the slice exclusive to it.

pacman100 (commented):
Thank you @muellerzr for the insightful blog 🤗. Please find my comments and suggestions.

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,427 @@
...
+        data, target = data.to(device), target.to(device)
+        output = model(data)
+        pred = output.argmax(dim=1, keepdim=True)
+        correct += pred.eq(target.view_as(pred)).sum().item()
+print(f'Accuracy: {100. * correct / len(test_loader.dataset)}')
+```
+
+Typically from here, one could either throw all of this into a python script or run it on a Jupyter Notebook.
+
+However, how would you then get this to run on say two GPUs if these resources available? Just doing `python myscript.py` will only ever run the script using a single GPU. This is where `torch.distributed` comes into play
```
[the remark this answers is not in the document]
muellerzr: Met in the middle on this suggestion and reworded it some

muellerzr:
Going ahead with the blogpost to have it launch by Friday. Still waiting on the torch folks but I think we're set to go from what I can see :) 

We'll split up Sourab's comments into a more advanced "Part 2" to leave this one as intro friendly.

@LysandreJik can you give it a glance and then I think we'll be set :) Thanks!

pytorch-ddp-accelerate-transformers.md
muellerzr: Will do this in a followup as a deep dive, so that this can be surface level :)

pytorch-ddp-accelerate-transformers.md
muellerzr: Fixed :)

pytorch-ddp-accelerate-transformers.md
```diff
@@ -0,0 +1,444 @@
...
+
+> Note: Here rank is the overall rank of the current GPU compared to all the other GPUs available, meaning they have a rank of `0 -> n-1`
+
+```python
+from torch.nn.parallel import DistributedDataParallel as DDP
+
+def train(model, rank, world_size):
+    setup(rank, world_size)
+    model = model.to(rank)
+    ddp_model = DDP(model, device_ids=[rank])
```
mrshenli, line 169: Thanks @muellerzr for putting up the blog post!!

I just wanna mention a few more techniques to further speed up DDP. No need to change version if the goal is not squeezing out the last bit of perf from the cluster. 

1. Setting `gradient_as_bucket_view` to `True`. This can make `param.grad` points to an offset of DDP's bucket, which can help save two copies (i.e., grad to bucket, bucket to grad).
2.  It is possible to overlap optimizer step with DDP's backward communication by using the DDP comm hook. (https://pytorch.org/docs/stable/generated/torch.nn.parallel.DistributedDataParallel.html#torch.nn.parallel.DistributedDataParallel.register_comm_hook) However, this will only make a difference if communication delay is dominating.

muellerzr: Thanks @mrshenli! The goal of this one is not, but I'll definitely include it in the next! (And look at having this happen also in `accelerate`), thanks for the TIL! 🤗

PR #596, opened 2022-10-24
Change install command in habana-getting-started.md

regisss:
Changing `python setup.py install` to `pip install .` when installing `optimum-habana` from source because pip seems to manage dependencies better (I cleaned the deps but at some point `python setup.py install` was failing while `pip install .` was not).


regisss:
cc @juliensimon @julien-c

PR #602, opened 2022-10-28
Update openvino.md

VioletteLepercq:


openvino.md
```diff
@@ -6,7 +6,7 @@ thumbnail: /blog/assets/113_openvino/thumbnail.png
 <h1>Shrink and accelerate your models with 🤗 Optimum Intel and OpenVINO</h1>
 
 <div class="blog-metadata">
-    <small>Published July 20, 2022.</small>
+    <small>Published October 27, 2022.</small>
```
echarlaix, line 9: Good catch @VioletteLepercq ! We can either put today's date (28) or the 31, currently I have kept [31](https://github.com/huggingface/blog/blob/echarlaix/openvino/_blog.yml#L1374) but we might have to wait couple of days to be aligned with Intel on this approval discussion

PR #629, opened 2022-11-11
fix spelling mistakes

cansakirt:


guspan-tanadi (changes requested):
Thanks for the fix.

osanseviero (approved):
Thanks!

PR #651, opened 2022-11-21
Fixed a mistake in specifying inputs to UNet

varuntulsian2001:
The UNet blocks get a batch of timesteps as input as opposed to getting noise as input


patrickvonplaten (approved):
Thank you!

PR #666, opened 2022-11-28
undefined device in train_ddp_accelerate

kiansierra:


kiansierra (commented):
To my understanding in using accelerator for ddp, we would want the accelerator.device

PR #683, opened 2022-12-02
Core ML post: simplify chunked/unchunked

pcuenca:
Simplify description of variants, we no longer have `chunked` and `unchunked` nodes in the tree.
Add link to Apple's `split_einsum` post.


osanseviero (approved):
Thanks!

PR #694, opened 2022-12-05
Add emoji to beta macOS note

pcuenca:
Hopefully it changes the style too.


julien-c:
(What's the issue with the style?)

pcuenca:
Mainly, I think the quotes look slightly weird in this case:

Screenshot 2022-12-05 at 13 52 32

I should create a local environment for testing.

osanseviero:
Note you can use GitHub Codespaces so no need to do a local setup which can be a bit tedious - see https://github.com/huggingface/moon-landing/tree/main#codespace

pcuenca:
Oh right! Thank you!

PR #701, opened 2022-12-08
Typo: Add missing ' ' in  codeparrot post on 'gpt-large'

kargaranamir:
### Description
Based on the [from_pretrained](https://huggingface.co/gpt2-large) method, you need to pass the path of the model which is 'gpt2-large' not the variable of gpt2-large which has no reference.

### Fix
The ' ' was missing.

### Solution
Add missing ' '


kargaranamir:
It'd be great if @patrickvonplaten reviewed it.

PR #734, opened 2022-12-15
[Ethics soc2] A few fixes

julien-c:


yjernite:
Thanks! I should also update the thumbnail with @meg-huggingface 's new version

julien-c:
BTW all the visuals in the post are a bit small/low-res, don't hesitate to use larger ones

PR #737, opened 2022-12-15
suggestions to improve title

clmnt:
The blogpost is amazing and I think it would make it more appealing like this rather than with a title about the newsletter. We can try to make it the blogpost of reference for everyone who talks about biases in ML!

### DIFF
--- a/ethics-soc-2.md
+++ b/ethics-soc-2.md
@@ -1,5 +1,5 @@
 ---
-title: "Let's talk about biases for machine learning" 
+title: "Let's talk about biases in machine learning! Ethics and Society Newsletter #2" 
 thumbnail: /blog/assets/122_ethics_soc_2/thumbnail-solstice.png
 ---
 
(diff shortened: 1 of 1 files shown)


ethics-soc-2.md
```diff
@@ -1,9 +1,9 @@
 ---
-title: "Ethics and Society Newsletter #2" 
+title: "Let's talk about biases for machine learning" 
 thumbnail: /blog/assets/122_ethics_soc_2/thumbnail-solstice.png
 ---
 
-# Ethics and Society Newsletter #2
+# Let's talk about biases for machine learning
```
yjernite, line 6:
```suggestion
# Ethics and Society Newsletter #2: Let's talk about ML biases!
```

yjernite (commented):
Good point! I think we also want to keep the brand/series identity, so suggested an option to combine those

What do you think @meg-huggingface ?

PR #742, opened 2022-12-16
Update ethics-soc-2.md

yjernite:
I promise I'll set up local deployment to try out looks for the next blog post 😅


julien-c:
probably a good idea:)

osanseviero:
Or a bit easier is to use Codespaces for this; there's a section that shows how to get the blog for it :) https://github.com/huggingface/moon-landing#codespace

PR #759, opened 2022-12-23
Fix typo

mrm8488:


regisss (approved):
Thanks for fixing this @mrm8488!

PR #762, opened 2022-12-27
The previous code does not work with the latest version

abhi-11nav:
The previous code doesn't work with the latest version of transformers. It generated random text when we provide the model with deocoder input_id's and attention mask. Seems like the latest version implements shift right on input_ids automatically. Please refer to the link for clarification : https://github.com/huggingface/transformers/issues/18958 

The current version works for me with codebert as input and gpt2 as output. Thank you.


abhi-11nav (commented):
The previous code does not work with the latest version of transformers.

abhi-11nav:
@patrickvonplaten , please take a look into this. Thank you

patrickvonplaten:
Thanks for the fix @abhi-11nav !

PR #761, opened 2022-12-27
AI for Game Development: Creating a Farming Game in 5 Days #1

dylanebert:

PR #763, opened 2022-12-28
Adding Game Development as tags

simoninithomas:
Based on #761 we have a new section in the blog for using ML for game development

I called it game development.

### DIFF
--- a/_tags.yml
+++ b/_tags.yml
@@ -33,3 +33,6 @@
 
 - value: diffusion
   label: Diffusion
+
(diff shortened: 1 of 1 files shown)


_tags.yml
```diff
@@ -33,3 +33,6 @@
 
 - value: diffusion
   label: Diffusion
+
+- value: game
```
julien-c, line 37:
```suggestion
- value: game-dev
```

maybe?

PR #766, opened 2023-01-03
Minor tweak for math rendering

mirzask:
Changed a couple of inline math stuff that was using `$` to `(\` because the `$` wasn't rendering properly. Also deleted a straggling ")".


osanseviero (approved):
Thanks!! :fire:

PR #782, opened 2023-01-16
Added datasets for LLMs internship

TevenLeScao:
Adding the datasets intership


LysandreJik (approved):
Thank you @TevenLeScao!

PR #785, opened 2023-01-17
fix variable used once with wrong name

ViktorStagge:
`mase_metric` created in previous cell
undefined `metric` used in print statement.


osanseviero (approved):
Thanks!

PR #796, opened 2023-01-24
mask2former blog fix for image processor

shivalikasingh95:
This PR fixes the reference regarding image processor for Mask2Former since Mask2Former has a separate image processor now as per this [PR](https://github.com/huggingface/transformers/pull/21256/files).


shivalikasingh95:
@alaradirik / @NielsRogge would be great if you one of you could do a quick review here and merge this.

NielsRogge (approved):
Thanks for fixing!

PR #797, opened 2023-01-24
update readme for asset location

natolambert:
cc @osanseviero @mishig25, noticed this wasn't updated when helping @nazneenrajani.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ This is the official repository of the [Hugging Face Blog](https://hf.co/blog).
 2️⃣ Create a md (markdown) file, **use a short file name**.
 For instance, if your title is "Introduction to Deep Reinforcement Learning", the md file name could be `intro-rl.md`. This is important because the **file name will be the blogpost's URL**.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,7 +7,9 @@ This is the official repository of the [Hugging Face Blog](https://hf.co/blog).
...
 For instance, if your title is "Introduction to Deep Reinforcement Learning", the md file name could be `intro-rl.md`. This is important because the **file name will be the blogpost's URL**.
 
-3️⃣ Create a new folder in `assets`. Use the same name as the name of the md file. Optionally you may add a numerical prefix to that folder, using the number that hasn't been used yet. But this is no longer required. i.e. the asset folder in this example will be `123_intro-rl` or `intro-rl`. This folder will contain **all your illustrations and thumbnail**. The folder number is mostly for (rough) ordering purposes, so it's no big deal if two concurrent articles use the same number.
+3️⃣ Create a new folder in `assets`. Use the same name as the name of the md file. Optionally you may add a numerical prefix to that folder, using the number that hasn't been used yet. But this is no longer required. i.e. the asset folder in this example will be `123_intro-rl` or `intro-rl`. This folder will contain **all your thumbnail only**. The folder number is mostly for (rough) ordering purposes, so it's no big deal if two concurrent articles use the same number.
```
julien-c, line 10:
```suggestion
3️⃣ Create a new folder in `assets`. Use the same name as the name of the md file. Optionally you may add a numerical prefix to that folder, using the number that hasn't been used yet. But this is no longer required. i.e. the asset folder in this example will be `123_intro-rl` or `intro-rl`. This folder will contain **your thumbnail only**. The folder number is mostly for (rough) ordering purposes, so it's no big deal if two concurrent articles use the same number.
```

natolambert: good catch.

PR #813, opened 2023-01-30
Fix thumbnail for direct link

osanseviero:
The thumbnail is still broken for social media


julien-c:
i did it in #816 so closing this one

PR #818, opened 2023-02-03
Update hf-bitsandbytes-integration.md

gante:
Fixes typos :)

Thank you for the cool blog post 💛

### DIFF
--- a/hf-bitsandbytes-integration.md
+++ b/hf-bitsandbytes-integration.md
@@ -62,13 +62,13 @@ Float32 (FP32) stands for the standardized IEEE 32-bit floating point representa
 In the float16 (FP16) data type, 5 bits are reserved for the exponent and 10 bits are reserved for the mantissa. This makes the representable range of FP16 numbers much lower than FP32. This exposes FP16 numbers to the risk of overflowing (trying to represent a number that is very large) and underflowing (representing a number that is very small).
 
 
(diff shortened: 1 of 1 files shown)


hf-bitsandbytes-integration.md
```diff
@@ -62,13 +62,13 @@ Float32 (FP32) stands for the standardized IEEE 32-bit floating point representa
...
+For example, if you do `10k * 10k` you end up with `100M` which is not possible to represent in FP16, as the largest number possible is `64k`. And thus you'd end up with `NaN` (Not a Number) result and if you have sequential computation like in neural networks, all the prior work is destroyed.
 Usually, loss scaling is used to overcome this issue, but it doesn't always work well.
 
 A new format, bfloat16 (BF16), was created to avoid these constraints. In BF16, 8 bits are reserved for the exponent (which is the same as in FP32) and 7 bits are reserved for the fraction.
 
 
-This means that in BF16 we can retain the same dynamic range as FP32. But we lose 3 bits of precision. Now there is absolutely no problem with huge numbers, but the precision is worse than FP16 here.
+This means that in BF16 we can retain the same dynamic range as FP32. But we lose 13 bits of precision. Now there is absolutely no problem with huge numbers, but the precision is worse than FP16 here.
```
gante, line 71:
```suggestion
This means that in BF16 we can retain the same dynamic range as FP32. But we lose 3 bits of precision with respect to FP16. Now there is absolutely no problem with huge numbers, but the precision is worse than FP16 here.
```

gante: @younesbelkada maybe like this?

younesbelkada: Yes LGTM!

younesbelkada (approved):
Thanks for fixing! 💯

PR #823, opened 2023-02-07
Fix paddlepaddle twitter

julien-c:
Looks like twitter account changed to cleaner name!

cc @sijunhe 

also cc @xianbaoqian


sijunhe (approved):
Thanks @julien-c . Yes we were able to get back the real twitter handle. :)

PR #835, opened 2023-02-15
Fix example code in 3dmoljs blog post

duerrsimon:
3dmol.js received a major rewrite that removed JQuery from the bundle. Readding JQuery to make the demo work again.


osanseviero (approved):
Thanks!

PR #862, opened 2023-02-25
tiny tweaks to fast-mac-diffusers

julien-c:
mostly because "models" is a bit confusing in the context of Mac devices (not ML models)


pcuenca (approved):
Good call, thanks a lot!

PR #883, opened 2023-03-02
[rlhf.md] fixes

stas00:
I proposed a few corrections, plus there are these that I wasn't sure about and you may want to fix those:

> the observation space is the possible input token sequences, which is also quite large (size of vocabulary ^ number of input tokens)

the grammar doesn't check out here - needs rewrite.

> When deploying a system using RLHF, gathering the human preference data is quite expensive due to the mandatory and thoughtful human component. 

I don't know what you mean by "thoughtful human component", perhaps "slow human component"

> PPO is a relatively old algorithm, but there are no structural reasons that other algorithms could offer benefits and permutations on the existing RLHF workflow

did you mean to say "could not offer" instead? missing not? - very different meaning

Thanks.


rlhf.md
```diff
@@ -75,7 +75,7 @@ At this point in the RLHF system, we have an initial language model that can be
...
+Training a language model with reinforcement learning was, for a long time, something that people would have thought as impossible both for engineering and algorithmic reasons. What multiple organizations seem to have gotten to work is fine-tuning some or all of the parameters of a **copy of the initial LM** with a policy-gradient RL algorithm, Proximal Policy Optimization (PPO). Parameters of the LM are frozen because fine-tuning an entire 10B or 100B+ parameter model is prohibitively expensive (for more, see Low-Rank Adaptation ([LoRA](https://arxiv.org/abs/2106.09685)) for LMs or the [Sparrow](https://arxiv.org/abs/2209.14375) LM from DeepMind). PPO has been around for a relatively long time – there are [tons](https://spinningup.openai.com/en/latest/algorithms/ppo.html) of [guides](https://huggingface.co/blog/deep-rl-ppo) on how it works. The relative maturity of this method made it a favorable choice for scaling up to the new application of distributed training for RLHF. It turns out that many of the core RL advancements to do RLHF have been figuring out how to update such a large model with a familiar algorithm (more on that later).
```
stas00, line 78: github fails to highlight this change - I changed "is impossible" to "as impossible" in the first line.

PR #909, opened 2023-03-06
👷 Add CI to check YML files

coyotte508:
https://huggingface.slack.com/archives/C035N5UE2JU/p1678135727039249

### DIFF
--- a/.github/workflows/validate-yaml.yml
+++ b/.github/workflows/validate-yaml.yml
@@ -1,0 +1,20 @@
+name: YAML schema validator
+on: [push]
+
+jobs:
+  yaml-schema-validation:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


julien-c (approved):
oh the validation is maybe overkill i was just thinking of "not broken yaml" – but i guess, who can do more, can do less?

coyotte508:
> oh the validation is maybe overkill i was just thinking of "not broken yaml" – but i guess, who can do more, can do less?

eh I didn't use `regex` to valited date, tags and such (and didn't find a way to do a XOR between a blog's `local` / `external`) so not too much overkill xD

coyotte508:
> (and didn't find a way to do a XOR between a blog's `local` / `external`) 

Could have used `any(include('localBlog'), include('externalBlog'))` :thinking: 

Let's hope it won't be needed in the future

PR #923, opened 2023-03-10
[time-series] informer fix yaml

kashif:
cc @NielsRogge


informer.md
```diff
@@ -3,7 +3,7 @@ title: "Multivariate Probabilistic Time Series Forecasting with Informer"
 thumbnail: /blog/assets/134_informer/thumbnail.png
 authors:
 - user: elisim
-   guest: true
+  guest: true
```
julien-c, line 6: maybe overkill but @coyotte508 for exhaustivity we could also validate the yaml in the yaml blocks...

coyotte508: @.@

PR #926, opened 2023-03-12
informer blog fixes

elisim:
@kashif @NielsRogge

### DIFF
--- a/informer.md
+++ b/informer.md
@@ -339,7 +339,7 @@ num_of_variates = len(train_dataset)
 train_grouper = MultivariateGrouper(max_target_dim=num_of_variates)
 test_grouper = MultivariateGrouper(
     max_target_dim=num_of_variates,
-    num_test_dates=len(test_dataset) // len(train_dataset), # number of rolling test windows
+    num_test_dates=len(test_dataset) // num_of_variates, # number of rolling test windows
 )
 
 multi_variate_train_dataset = train_grouper(train_dataset)
--- a/informer.md
+++ b/informer.md
@@ -44,7 +44,7 @@ The main idea of ProbSparse is that the canonical self-attention scores form a l
 
 | ![informer_full_vs_sparse_attention](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/informer/informer_full_vs_sparse_attention.png) |
 |:--:|
-| Vanilla self attention vs ProbSparse attention from [Autoformer (Wu, Haixu, et al., 2021)](https://wuhaixu2016.github.io/pdf/NeurIPS2021_Autoformer.pdf)
+| Vanilla self attention vs ProbSparse attention from [Autoformer (Wu, Haixu, et al., 2021)](https://wuhaixu2016.github.io/pdf/NeurIPS2021_Autoformer.pdf) |
 
 Given the idea of "active" and "lazy" queries, the ProbSparse attention selects the "active" queries, and creates a reduced query matrix \\(Q_{reduced}\\) which is used to calculate the attention weights in \\(O(T \log T)\\). Let's see this more in detail with a code example. 
     
--- a/informer.md
+++ b/informer.md
@@ -672,7 +672,7 @@ def create_train_dataloader(
     )
 
     # the instance splitter will sample a window of
-    # context length + lags + prediction length (from the 862 possible transformed time series)
+    # context length + lags + prediction length (from all the possible transformed time series, 1 in our case)
     # randomly from within the target time series and return an iterator.
     training_instances = instance_splitter.apply(
         Cyclic(transformed_data)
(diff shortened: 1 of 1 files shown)


informer.md
```diff
@@ -672,7 +672,7 @@ def create_train_dataloader(
     )
 
     # the instance splitter will sample a window of
-    # context length + lags + prediction length (from the 366 possible transformed time series)
+    # context length + lags + prediction length (from the 862 possible transformed time series)
```
kashif, line 675:
```suggestion
    # context length + lags + prediction length (from all the possible transformed time series, 1 in our case)
```

informer.md
```diff
@@ -44,7 +44,7 @@ The main idea of ProbSparse is that the canonical self-attention scores form a l
 
 | ![informer_full_vs_sparse_attention](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/informer/informer_full_vs_sparse_attention.png) |
 |:--:|
-| Vanilla self attention vs ProbSparse attention from [Autoformer](https://wuhaixu2016.github.io/pdf/NeurIPS2021_Autoformer.pdf) |
+| Vanilla self attention vs ProbSparse attention from [Autoformer (Wu, Haixu, et al., 2021)](https://wuhaixu2016.github.io/pdf/NeurIPS2021_Autoformer.pdf)
```
kashif, line 47:
```suggestion
| Vanilla self attention vs ProbSparse attention from [Autoformer (Wu, Haixu, et al., 2021)](https://wuhaixu2016.github.io/pdf/NeurIPS2021_Autoformer.pdf) |
```

informer.md
```diff
@@ -60,9 +60,9 @@ $$
 \textrm{ProbSparseAttention}(Q, K, V) = \textrm{softmax}(\frac{Q_{reduce}K^T}{\sqrt{d_k}} )V
 $$
 
-where the \\(Q_{reduce}\\) matrix only selects the Top-\\(u\\) "active" queries. Here, \\(u = c \cdot \log L_Q\\) and \\(c\\) called the _sampling factor_ hyperparameter for the ProbSparse attention. Since \\(Q_{reduce}\\) selects only the Top-\\(u\\) queries, it's size is \\(c\cdot \log L_Q \times d\\), so the multiplication \\(Q_{reduce}K^T\\) takes only \\(O(L_K \log L_Q) = O(T \log T)\\).
+where the \\(Q_{reduce}\\) matrix only selects the Top u "active" queries. Here, \\(u = c \cdot \log L_Q\\) and \\(c\\) called the _sampling factor_ hyperparameter for the ProbSparse attention. Since \\(Q_{reduce}\\) selects only the Top u queries, it's size is \\(c\cdot \log L_Q \times d\\), so the multiplication \\(Q_{reduce}K^T\\) takes only \\(O(L_K \log L_Q) = O(T \log T)\\).
```
kashif, line 63:
```suggestion
where the \\(Q_{reduce}\\) matrix only selects the Top  \\(u)\\ "active" queries. Here, \\(u = c \cdot \log L_Q\\) and \\(c\\) called the _sampling factor_ hyperparameter for the ProbSparse attention. Since \\(Q_{reduce}\\) selects only the Top \\(u\\) queries, its size is \\(c\cdot \log L_Q \times d\\), so the multiplication \\(Q_{reduce}K^T\\) takes only \\(O(L_K \log L_Q) = O(T \log T)\\).
```

informer.md
```diff
@@ -60,9 +60,9 @@ $$
...
 
-This is good! But how can we select the \\(u\\) "active" queries to create \\(Q_{reduce}\\)? Let's define the _Query Sparsity Measurement_.
+This is good! But how can we select the u "active" queries to create \\(Q_{reduce}\\)? Let's define the _Query Sparsity Measurement_.
```
kashif, line 65:
```suggestion
This is good! But how can we select the \\(u\\) "active" queries to create \\(Q_{reduce}\\)? Let's define the _Query Sparsity Measurement_.
```

informer.md
kashif: issue was the `-` so perhaps remove it?

kashif:
LGTM! Thank you!

kashif (approved):

informer.md
```diff
@@ -27,7 +27,7 @@ A few months ago we introduced the [Time Series Transformer](https://huggingface
...
 
-Modeling the full joint conditional distribution of high dimensional data can get computationally expensive and thus methods resort to some approximation of the distribution, the easiest being to model the data as an independent distribution from the same family, or some low-rank approximation to the full covariance, etc. Here we will just resort to the independent (or diagonal) emissions which is supported for the families of distributions we have implemented. 
+Modeling the full joint conditional distribution of high dimensional data can get computationally expensive and thus methods resort to some approximation of the distribution, the easiest being to model the data as an independent distribution from the same family, or some low-rank approximation to the full covariance, etc. Here we will just resort to the independent (or diagonal) emissions which is supported for the families of distributions we have implemented [here](https://huggingface.co/docs/transformers/main/en/internal/time_series_utils.
```
kashif, line 30:
```suggestion
Modeling the full joint conditional distribution of high dimensional data can get computationally expensive and thus methods resort to some approximation of the distribution, the easiest being to model the data as an independent distribution from the same family, or some low-rank approximation to the full covariance, etc. Here we will just resort to the independent (or diagonal) emissions which is supported for the families of distributions we have implemented [here](https://huggingface.co/docs/transformers/main/en/internal/time_series_utils).
```

PR #910, opened 2023-03-07
Add cn translations

mishig25:
Move translations from https://huggingface.co/datasets/HuggingFace-CN-community/translation/tree/main to this repo

Made necessary changes in `_blog.yml` file

- [ ] todo: add equivalent of https://github.com/huggingface/blog/pull/909

PR #930, opened 2023-03-13
YML validation for chinese translation

mishig25:
Follow up to https://github.com/huggingface/blog/pull/910


coyotte508 (approved):
If we add more translations, a Deno script as @julien-c suggested looks good :)

mishig25:
So far, we don't have any other plans for translation. However, will keep that in mind

> If we add more translations, a Deno script as @julien-c suggested looks good :)

xianbaoqian:
Thanks for the PR!

PR #931, opened 2023-03-14
inference fix and parenthesis

elisim:
@kashif @NielsRogge


informer.md
```diff
@@ -902,7 +902,7 @@ model.eval()
 
 forecasts_ = []
 
-for batch in [next(iter(test_dataloader))]:
+for batch in test_dataloader:
```
elisim, line 905: `test_dataloader` is one sample either way, but imho it's more nice the code will iterate all the `test_dataloader` 🙂

kashif:
LGTM thanks! I will update the notebook with these latest changes.

PR #939, opened 2023-03-21
OCD: keep same order of props in YAML as in the EN blog

julien-c:


julien-c:
BTW @xianbaoqian would you be able to open a PR to:
- translate the labels in `cn/_tags.yml`
- replace events with relevant ones in `_events.yml`
- potentially, rename the whole folder to `zh/` (if we do decide to change the language code to better match the one for transformers etc)

Thanks!

mishig25 (approved):
lgtm, thanks for formatting !

julien-c:
(@xianbaoqian feel free to other new issues to track)

PR #963, opened 2023-03-27
Update _blog.yml

xianbaoqian:
cc @chenglu

### DIFF
--- a/zh/_blog.yml
+++ b/zh/_blog.yml
@@ -293,3 +293,13 @@
   date: March 3, 2023
   tags:
     - diffusers
+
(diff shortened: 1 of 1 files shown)


zh/_blog.yml
```diff
@@ -293,3 +293,13 @@
   date: March 3, 2023
   tags:
     - diffusers
+
+- local: bloom-inference-optimization
+  title: 优化故事: BLOOM 模型推理
```
julien-c, line 298: you need to quote the whole title string as `:` makes it invalid YAML unless quoted

PR #971, opened 2023-03-28
add translation for train-your-controlnet

Hoi2022:


xianbaoqian:
Thanks for the contribution! Please also update  https://github.com/huggingface/blog/blob/main/zh/_blog.yml  Thx!

PR #980, opened 2023-03-30
Update fl-with-flower.md

charlesbvll:
I thought it might be good to explicitly call the `load_data` function, I think this was not made clear enough.


charlesbvll:
@osanseviero Sorry to bother you again, but I think this should be nice to add!

osanseviero (approved):
Thanks!

PR #985, opened 2023-03-31
Sync zh social posts back to the main repo

chenglu:
Hi Tiezhen @xianbaoqian & team:

Please take a look at this PR, which include all the previous social posts on WeChat, as these should be regarded as the `zh` localized blog content.

Kudos to @zhongdongy and @inferjay for their outstanding efforts in reformatting (and reviewing) the published Chinese content with the English original's markdown format, which need to transforming the content into markdown, adjusting links, and modifying image references, etc.

Thanks,
Luke


xianbaoqian:
This is awesome and it would unblock our announcement of the Chinese blog site. I know that there are a lot of detailed  work checking URLs for the image and the translated content. This is a true marathon. Huge thanks to @zhongdongy and @inferjay ! And also huge thanks to @chenglu for the great coordination and communication work! Howrray!

julien-c:
also cc @adeenayakup for information!

PR #1007, opened 2023-04-07
Informer - remove install from source, fixed math

elisim:
I updated also in the notebook repo
@kashif @NielsRogge


NielsRogge:
Thanks!

PR #1027, opened 2023-04-19
Add: ZH translation for ethics-soc-3.md

zhongdongy:


xianbaoqian:
Great work. Thanks @chenglu  for trying out the new process!!! Thanks @innovation64 and @zhongdongy  for the hard work!

@adeenayakup You might be interested in this article :-)

PR #1064, opened 2023-04-29
Assisted Generation blog post

gante:
# What's this about?

This PR contains the blog post for assisted generation, an original method by Hugging Face that reduces text generation latency 🚤 

## Review notes:

I've put some energy into the visual side of the blog post, and created 4 GIFs that should help explain what's going on. The quality of the GIFs in the PR is on the low side (Keynote's export functionality is underwhelming). The full-quality GIFs will be the result of a manual screen recording process, meaning that I will only create them after the review process is complete. Nevertheless, feedback on the GIFs is deeply encouraged! Everything is customizable, don't hold back your punches 👊 

## TO DO:
- [x] Benchmarks spaces: https://huggingface.co/spaces/joaogante/assisted_generation_benchmarks
- [x] Demo spaces: https://huggingface.co/spaces/joaogante/assisted_generation_demo
- [x] Cover
- [x] Add citation information?
- [x] Draft Twitter and LI comms to go along this blog post

## After merging:
- [x] Double-check whether the formatting is correct on the blog
- [ ] Spread the word


assisted-generation.md
```diff
@@ -0,0 +1,177 @@
...
+
+First, the requirement – the assistant must have the exact same tokenizer as your model. If this requirement was not in place, expensive token decoding and re-encoding steps would have to be added. Furthermore, these additional steps would have to happen on the CPU, which in turn may need slow inter-device data transfers. Fast usage of the assistant is critical for the benefits of assisted generation to show up.
+
+Finally, the heuristic. By this point, you have probably noticed the similarities between the movie Inception and assisted generation – you are, after all, running text generation inside text generation. There will be one assistant model forward pass per candidate token, and we know that forward passes are expensive. While you can’t know in advance the number of tokens that the assistant model will get right, you can keep track of this information and use it to limit the number of candidate tokens requested to the assiatant – some sections of the output are easier to anticipate than others.
```
Narsil, line 106:
```suggestion
Finally, the heuristic. By this point, you have probably noticed the similarities between the movie Inception and assisted generation – you are, after all, running text generation inside text generation. There will be one assistant model forward pass per candidate token, and we know that forward passes are expensive. While you can’t know in advance the number of tokens that the assistant model will get right, you can keep track of this information and use it to limit the number of candidate tokens requested to the assistant – some sections of the output are easier to anticipate than others.
```

Narsil (commented):
Good blog post I really like ! 

Nice GIFs.

A few nits on formulation.

sgugger (approved):
LGTM, with the comments filled out ;-)
Thanks a lot for drafting this!

assisted-generation.md
```diff
@@ -0,0 +1,177 @@
...
+---
+
+# Assisted Generation: a new direction toward low-latency text generation
+
+<!-- {blog_metadata} -->
+<!-- {authors} -->
+
+Billions of dollars are being poured into large language models and related technologies. We have been finding that larger models show emergent behavior and can solve a wider array of tasks. However, as humans with ever-decreasing attention spawns, we also dislike their slow response times. Latency is critical for a good user experience, and smaller models are often used despite their lower quality (e.g. in [code completion](https://ai.googleblog.com/2022/07/ml-enhanced-code-completion-improves.html)).
+
+Why is text generation so slow? What’s preventing you from deploying low-latency large language models without going bankrupt? In this blog post, we will revisit the bottlenecks for autoregressive text generation and introduce a new decoding method to tackle the latency problem. You’ll see that using our new method, assisted generation, you can reduce latency up to 10x in commodity hardware!
```
lewtun, line 15: nit:
```suggestion
Why is text generation so slow? What’s preventing you from deploying low-latency large language models without going bankrupt? In this blog post, we will revisit the bottlenecks for autoregressive text generation and introduce a new decoding method to tackle the latency problem. You’ll see that by using our new method, **assisted generation**, you can reduce latency up to 10x in commodity hardware!
```

assisted-generation.md
```diff
@@ -0,0 +1,177 @@
...
+
+<!-- [GIF 1 -- FWD PASS] -->
+
+A model forward pass gets you the logits for the next token, which you can freely manipulate (e.g. set the probability of undesirable words or sequences to 0). The following step in text generation is to select the next token from these logits. Common strategies include picking the most likely token, known as greedy decoding, or sampling from this distribution, also called multinomial sampling. Chaining model forward passes with next token selection iteratively gets you text generation. This explanation is the tip of the iceberg, please refer to [our blog post on text generation](https://huggingface.co/blog/how-to-generate) for an in-depth exploration.
```
lewtun, line 23:
```suggestion
A model forward pass gets you the logits for the next token, which you can freely manipulate (e.g. set the probability of undesirable words or sequences to 0). The following step in text generation is to select the next token from these logits. Common strategies include picking the most likely token, known as greedy decoding, or sampling from this distribution, also called multinomial sampling. Chaining model forward passes with next token selection iteratively gets you text generation. This explanation is the tip of the iceberg when it comes to decoding methods; please refer to [our blog post on text generation](https://huggingface.co/blog/how-to-generate) for an in-depth exploration.
```

assisted-generation.md
```diff
@@ -0,0 +1,177 @@
...
+
+<!-- [GIF 2 -- TEXT GENERATION] -->
+
+From the description above, the latency bottleneck in text generation is clear: running a model forward pass for large models is slow, and you may need to do hundreds of them in a sequence. But let’s dive deeper: why are forward passes slow? Forward passes are typically dominated by matrix multiplications and, after a quick visit to the [corresponding wikipedia section](https://en.wikipedia.org/wiki/Matrix_multiplication_algorithm#Communication-avoiding_and_distributed_algorithms), you call tell that memory bandwidth is the limitation in this operation. In other words, *the bottleneck in the forward pass comes from loading the model layer weights into the computation cores of your device, not from performing the computations themselves*.
```
lewtun, line 27: I don't quite understand the punchline here that loading model weights => slow forward pass. Once the weights are loaded, isn't computation the actual latency bottleneck ? Perhaps you mean to say that naively loading the model in full precision on a single device is inefficient?

assisted-generation.md
```diff
@@ -0,0 +1,177 @@
...
+
+Second, when you know you’ll get concurrent text generation requests, you can batch the inputs and massively increase the throughput with a small latency penalty. The model layer weights loaded into the device are now used on several input rows in parallel, which means that you’ll get more tokens out for approximately the same memory bandwidth burden. The catch with batching is that you need additional device memory (or to offload the memory somewhere) – at the end of this spectrum, you can see projects like [FlexGen](https://github.com/FMInference/FlexGen) which optimize throughput at the expense of latency.
+
+```python
+# Example showcasing the impact of batched generation. Measured on a RTX3090
+from transformers import AutoModelForCausalLM, AutoTokenizer
+import time
+
+tok = AutoTokenizer.from_pretrained("distilgpt2")
```
lewtun, line 38: turbo nit for readability:
```suggestion
tokenizer = AutoTokenizer.from_pretrained("distilgpt2")
```

assisted-generation.md
```diff
@@ -0,0 +1,177 @@
...
+
+These three types of improvements can be used in tandem, resulting in [high throughput solutions](https://huggingface.co/blog/bloom-inference-pytorch-scripts). However, after applying hardware-specific optimizations, there are limited options to reduce latency – and the existing options are expensive. Let’s fix that!
```
lewtun, line 65: I think those BLOOM scripts are defunct and `tgi` is the recommended approach now
```suggestion
These three types of improvements can be used in tandem, resulting in [high throughput solutions](https://github.com/huggingface/text-generation-inference). However, after applying hardware-specific optimizations, there are limited options to reduce latency – and the existing options are expensive. Let’s fix that!
```

assisted-generation.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Language decoder forward pass, revisited
+
+You’ve read above that each model forward pass yields the logits for the next token, but that’s actually an incomplete description. During text generation, the typical iteration consists in the model receiving as input the latest generated token plus cached internal computations for all other previous inputs, returning the next token logits. Caching is used to avoid redundant computations, resulting in faster forward passes, but it’s not mandatory (and can be used partially). When caching is disabled, the input contains the entire sequence of tokens generated so far and the output contains the logits corresponding to the next token for *all positions* in the sequence! The logits at position N correspond to the distribution for the next token if the input consisted in the first N tokens, ignoring all subsequent tokens in the sequence. In the particular case of greedy decoding, if you pass the generated sequence as input and apply the argmax operator to the resulting logits, you will obtain the generated sequence back.
```
lewtun, line 69: turbo nit:
```suggestion
You’ve read above that each model forward pass yields the logits for the next token, but that’s actually an incomplete description. During text generation, the typical iteration consists in the model receiving as input the latest generated token, plus cached internal computations for all other previous inputs, returning the next token logits. Caching is used to avoid redundant computations, resulting in faster forward passes, but it’s not mandatory (and can be used partially). When caching is disabled, the input contains the entire sequence of tokens generated so far and the output contains the logits corresponding to the next token for *all positions* in the sequence! The logits at position N correspond to the distribution for the next token if the input consisted in the first N tokens, ignoring all subsequent tokens in the sequence. In the particular case of greedy decoding, if you pass the generated sequence as input and apply the argmax operator to the resulting logits, you will obtain the generated sequence back.
```

assisted-generation.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+
+This means that you can use a model forward pass for a different purpose: in addition to feeding some tokens to predict the next one, you can also pass a sequence to the model and double-check whether the model would generate that same sequence (or part of it).
+
+
+<!-- [GIF 3 -- FWD CONFIRMATION] -->
+
+
+Let’s consider for a second that you have access to a magical latency-free oracle model that generates the same sequence as your model, for any given input. For argument’s sake, it can’t be used directly, it’s limited to being an assistant to your generation procedure. Using the property described above, you could use this assistant model to get candidate output tokens followed by a forward pass with your model to confirm that they are indeed correct. In this utopian scenario, the latency of text generation would be reduced from `O(n)` to `O(1)`, `n` being the number of generated tokens. For long generations, we're talking about several orders of magnitude.
```
lewtun, line 94: nit:
```suggestion
Let’s consider for a second that you have access to a magical latency-free oracle model that generates the same sequence as your model, for any given input. For argument’s sake, it can’t be used directly, it’s limited to being an assistant to your generation procedure. Using the property described above, you could use this assistant model to get candidate output tokens followed by a forward pass with your model to confirm that they are indeed correct. In this utopian scenario, the latency of text generation would be reduced from `O(n)` to `O(1)`, with `n` being the number of generated tokens. For long generations, we're talking about several orders of magnitude.
```

assisted-generation.md
```diff
@@ -0,0 +1,177 @@
...
+
+Walking a step towards reality, the assistant model lost its oracle properties. Now it’s a latency-free model that gets some of the candidate tokens wrong, according to your model. Due to the autoregressive nature of the task, as soon as the assistant gets a token wrong, all subsequent candidates must be invalidated. However, that does not prevent you from querying the assistant again, after correcting the wrong token with your model, and repeating this process iteratively. Even if the assistant fails a few tokens, text generation would have an order of magnitude less latency than in its original form.
```
lewtun, line 96:
```suggestion
Walking a step towards reality, let's assume the assistant model has lost its oracle properties. Now it’s a latency-free model that gets some of the candidate tokens wrong, according to your model. Due to the autoregressive nature of the task, as soon as the assistant gets a token wrong, all subsequent candidates must be invalidated. However, that does not prevent you from querying the assistant again, after correcting the wrong token with your model, and repeating this process iteratively. Even if the assistant fails a few tokens, text generation would have an order of magnitude less latency than in its original form.
```

assisted-generation.md
```diff
@@ -0,0 +1,177 @@
...
+
+Obviously, there are no latency-free assistant models. Nevertheless, it is relatively easy to find a model that approximates some other model’s text generation outputs – smaller versions of the same architecture trained similarly often fit this property. Moreover, when the difference in model sizes becomes significant, the cost of using the smaller model as an assistant becomes an afterthought after factoring in the benefits of skipping a few forward passes! You now understand the core of assisted generation.
```
lewtun, line 98:
```suggestion
Obviously, there are no latency-free assistant models. Nevertheless, it is relatively easy to find a model that approximates some other model’s text generation outputs – smaller versions of the same architecture trained similarly often fit this property. Moreover, when the difference in model sizes becomes significant, the cost of using the smaller model as an assistant becomes an afterthought after factoring in the benefits of skipping a few forward passes! You now understand the core of _assisted generation_.
```

assisted-generation.md
```diff
@@ -0,0 +1,177 @@
...
+
+
+<!-- [DEMO] -->
+
+
+## Future directions
+
+Assisted generation shows that modern text generation strategies are ripe for optimization. Understanding that it is currently a memory-bound problem, not a compute-bound problem, allows us to apply simple heuristics to get the most out of the available memory bandwidth, alleviating the bottleneck. We believe that further refinement of the use of assistant models will get us even bigger latency reductions - for instance, we may be able to skip a few more forward passes if we request the assistant to generate several candidate continuations. Naturally, releasing high-quality small models to be used as assistants will be critical to realizing and amplifying the benefits.
+
+Initially released under our 🤗 transformers library, to be used with the `.generate()` function, we expect to offer it throughout the Hugging Face universe. Its implementation is also completely open-source so, if you’re working on text generation and not using our tools, feel free to use it as a reference.
```
lewtun, line 173:
```suggestion
Initially released under our 🤗 Transformers library, to be used with the `.generate()` function, we expect to offer it throughout the Hugging Face universe. Its implementation is also completely open-source so, if you’re working on text generation and not using our tools, feel free to use it as a reference.
```

lewtun (commented):
Awesome blog post @gante and really well written 🔥 !

I left some suggestions & a question about the observation that text gen is memory bound by model loading. Apart from that it looks great :)

assisted-generation.md
lewtun: Ah yes that makes a lot more sense, thanks @Narsil! Perhaps we could make this point more explicit in the text @gante ?

Pierrci:
Please use https://huggingface.co/datasets/huggingface/documentation-images/tree/main/blog to upload binary files such as videos instead, we want to keep this repo as slim as possible

PR #1069, opened 2023-05-02
Fix grammar / typo

mrm8488:


osanseviero (approved):
Thanks! In the future it would be great if you do a single PR for all typo changes :)

PR #1080, opened 2023-05-06
fix(annotated-diffusion.md): fix image shape desc of PIL and Tensor

csl122:
modifiy the comment after ToTensor with the correct image shape CHW


pcuenca (approved):
LGTM, thanks! /cc @kashif

PR #1081, opened 2023-05-07
Add link to StarCoder paper

thomwolf:


lvwerra (approved):
Thanks, i think we merged the PR with a few local commit missing, will fix a few other things as well.

PR #1091, opened 2023-05-10
Fix sampling algorithm typo for UNet model

mu94-csl:
The UNet model should be epsilon_theta, not z. 
z is noise.


NielsRogge:
cc @kashif

qgallouedec (approved):

PR #1114, opened 2023-05-17
[`RWKV`] Add correct guest tags

younesbelkada:
# What does this PR do?

Adds the correct guest tag to the co-author of the RWKV blogpost

cc @osanseviero


osanseviero (approved):
Thanks!

PR #1147, opened 2023-05-26
Update assisted-generation.md

jc-ryan:
Since forward_confirmation is newly generated, it should be generated[1:].tolist() == forward_confirmation[:-1].tolist()

### DIFF
--- a/assisted-generation.md
+++ b/assisted-generation.md
@@ -95,7 +95,7 @@ forward_confirmation = model(generated).logits.argmax(-1)
 
 # We exclude the opposing tips from each sequence: the forward pass returns
 # the logits for the next token, so it is shifted by one position.
-print(generated[:-1].tolist() == forward_confirmation[1:].tolist())  # True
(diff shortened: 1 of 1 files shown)


assisted-generation.md
```diff
@@ -95,7 +95,7 @@ forward_confirmation = model(generated).logits.argmax(-1)
 
 # We exclude the opposing tips from each sequence: the forward pass returns
 # the logits for the next token, so it is shifted by one position.
-print(generated[:-1].tolist() == forward_confirmation[1:].tolist())  # True
+print(generated[1:].tolist() == forward_confirmation[:-1].tolist())  # True
```
gante, line 98:
```suggestion
print(generated[0, 1:].tolist() == forward_confirmation[0, :-1].tolist())  # True
```

It was printing `True` because it was indexing on the wrong dimension (in addition to the issue you caught!) 😠

gante (approved):
Thank you for spotting it! @jc-ryan

PR #1152, opened 2023-05-29
Fix 8-bit Matrix Multiplication example dimensions

yaroslavpoltoran:
Old code
`input_ = torch.randn(64, dtype=torch.float16)`
produced error. See this [discuss](https://discuss.huggingface.co/t/the-quantization-code-in-the-gentle-introduction-to-8-bit-matrix-multiplication-for-transformers-blog-post-yields-error/38060)

So I fixed dimensions to correct one
`input_ = torch.randn((1, 64), dtype=torch.float16)`


younesbelkada (approved):
Thanks so much!

PR #1162, opened 2023-05-31
Added Speech Recognition in Unity tutorial post

dylanebert:


unity-asr.md
```diff
@@ -0,0 +1,346 @@
...
+---
+
+<h1>AI Speech Recognition in Unity</h1>
+
+<!-- {blog_metadata} -->
+<!-- {authors} -->
+ 
+## Introduction
+
+This tutorial guides you through the process of implementing state-of-the-art Speech Recognition in your Unity game using the Hugging Face Unity API. This feature can be used for giving commands, improving accessibility, or any other functionality where converting spoken words to text may be useful.
```
simoninithomas, line 15: Can we add some simple examples below? Like:
- Give order to a NPC using your voice
etc?
So that people can get an idea

simoninithomas (commented):
I think:
- We need to have a space with the demo so that people can try it directly on their browser.
- I'm doing the tutorial but there's some problem with the microphone I'm going to check if it's a common issue, if yes we should add an element on how to select your microphone on the unity app itself

simoninithomas:
For the update:

What I done:
- Created a dropdown
In start():
```
foreach (var device in Microphone.devices)
{
     devicesDropdown.options.Add(new TMP_Dropdown.OptionData() { text = device });
}
```

Then I take the device selected:

`string device_ = devicesDropdown.options[devicesDropdown.value].text;`

And I pass it to the recorder:
`clip = Microphone.Start(device_, false, 10, 44100);`

simoninithomas:
The Update LGTM just missing the game jam announcement but I know you're going to add a banner for that.

When do you plan to publish it?

PR #1166, opened 2023-06-01
fix to remove on-prem introducing-private-hub.md

clmnt:


lewtun:
OK to merge @julien-c @osanseviero ?

julien-c:
@lewtun i would instead add a big disclaimer on top saying that this is outdated and linking to the new pages

I'll do it if no-one gets around to it

PR #1172, opened 2023-06-04
Update 4bit-transformers-bitsandbytes.md

rekil156:
The value ends up being the same but this maybe clearer for the reader:  mantisaa bits 1010 we have `(2^-1 + 0 + 2^-3 + 0) = (0.5 + 0.125) = 0.625`  instead of 
mantisaa bits 1010 we have `(0 + 2^-1 + 0 +  2^-3 ) = (0.5 + 0.125) = 0.625`  Since the mantissa bits are read from left to right and the first bit is 1 it should correspond to 2^-1 and not 0

### DIFF
--- a/4bit-transformers-bitsandbytes.md
+++ b/4bit-transformers-bitsandbytes.md
@@ -62,7 +62,7 @@ It has been empirically proven that the E4M3 is best suited for the forward pass
 
 ### FP4 precision in a few words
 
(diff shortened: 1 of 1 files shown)


pcuenca (commented):
If I understand it correctly, the exponents are assigned right-to-left, and that's how the original text walked through the sequence. So if mantissa bit positions are represented as `3210` we have the following exponents assigned to each position:

```
3 -> 2^-3
2 -> 2^-2
1 -> 2^-1
0 -> 2^0
```

The original text starts from position 0 to position 3. If we were using the method you propose, results would be different for mantissa bits `1011`, for example.

PR #1176, opened 2023-06-05
Fix relative links in sagemaker-huggingface-llm.md

zachazar:
Having these as relative paths adds a `/blog/` in the url which breaks these links. Changing to absolute paths fixes the problem

/cc @philschmid


philschmid:
@zachazar Thank you for fixing!

zachazar:
Happy to help!

PR #1185, opened 2023-06-06
Update introducing-private-hub.md update note

jeffboudier:
small copy changes indeed can't cc @bassem


julien-c (commented):
lgtm, thanks (we should create or re-add a /platform page and update it to this messaging)

PR #1188, opened 2023-06-07
Sync DuckDB title between thumbnail page and post page

pcuenca:
Just in case it was not intentional to have different titles


stevhliu (approved):
Ah yes, thanks for catching that! 👍

PR #1208, opened 2023-06-10
RLHF KL penalty clarifications

stalkermustang:
Closes https://github.com/huggingface/blog/pull/1207

PR #1212, opened 2023-06-12
RLHF blogpost markdown fixes

stalkermustang:
Hi, following #1208 , it seems I messed up with Markdown (see below, first and last characters in every block, `_` and `*`)
image
image

This PR should fix that


lewtun (approved):
Thanks for the fix!

Issue #1221, opened 2023-06-13
Add DB-GPT


csunny:
DB-GPT: Revolutionizing Database Interactions with Private LLM Technology.  [Source Code](https://github.com/csunny/DB-GPT)

PR #1222, opened 2023-06-13
Add dbgpt introducing. (#1221)

csunny:
DB-GPT is an experimental open-source project that uses localized GPT large models to interact with your data and environment. With this solution, you can be assured that there is no risk of data leakage, and your data is 100% private and secure.

This is the introducing of [DB-GPT](https://github.com/csunny/DB-GPT)


introducing-private-hub.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: "Introducing the Private Hub: A New Way to Build With Machine Learning"
-thumbnail: /blog/assets/92_introducing_private_hub/thumbnail.png
+thumbnail: ../assets/92_introducing_private_hub/thumbnail.png
```
julien-c, line 3: wrong change

csunny: already fix this.

osanseviero (commented):
Hi there! Thanks for the PR!

Did you discuss this with a HF team member before opening this PR? The usual approach is to either discuss the topic and open an issue before opening the PR. I might not have visibility into existing discussions, though :)

julien-c:
where did you publish the blogpost in the end @csunny?

PR #1246, opened 2023-06-20
Update falcon.md

Alkahwaji:
Mistake in the number of tokens in Falcon-7B and 40B.


pcuenca:
Actually the numbers were taken from the [7B](https://huggingface.co/tiiuae/falcon-7b) and [40B](https://huggingface.co/tiiuae/falcon-40b) model cards, as well as [the summary table in this page](https://huggingface.co/tiiuae). Pinging @lewtun anyway for further confirmation.

PR #1268, opened 2023-06-25
Fix section reference in `hf-bitsandbytes-integration.md`

alvarobartt:
Hi to whoever is reading this! 🤗

## What's in this PR?

A simple section reference fix, as Markdown excludes the question mark, so the current reference didn't work.

Thanks!


pcuenca (approved):
Thanks a lot! 🙌

PR #1278, opened 2023-06-29
Mithril Security: Confidential Code Generation Article

lyie28:
Blog article on the deployment of BigCode's Santacoder model deployed in an enclave so devs can query model with privacy guarantees


confidential-code-generation.md
```diff
@@ -0,0 +1,96 @@
...
+  guest: true
+- user: mithril-security
+  guest: true
+---
+
+# AI-assisted code generation with privacy guarantees: Securely deploy SantaCoder with BlindBox on Azure
+
+<!-- {blog_metadata} -->
+<!-- {authors} -->
+
```
julien-c, line 17:
```suggestion

```

from #1554

PR #1282, opened 2023-07-03
Update evaluating-mmlu-leaderboard.md

clefourrier:
Removed huggingface from the table for llama.

### DIFF
--- a/evaluating-mmlu-leaderboard.md
+++ b/evaluating-mmlu-leaderboard.md
@@ -54,7 +54,7 @@ To settle the case, we decided to run these three possible implementations of th
 
 The results are surprising:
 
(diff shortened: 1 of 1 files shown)


pcuenca (approved):
Looks much clearer now, imo!

However, the new image is 4.5MB vs 73 Kb of the previous one, can you run it through an image optimizer? (I use ImageOptim on Mac, there are plenty online). Otherwise I can do it in a bit.

Screenshot 2023-07-03 at 10 46 40

clefourrier:
@pcuenca Should be good now

pcuenca:
Lol, we both did it at the same time (sorry)

Screenshot 2023-07-03 at 10 54 08

PR #1313, opened 2023-07-18
👷 Use simple script to validate yaml

coyotte508:
Context: https://huggingface.slack.com/archives/C02EMARJ65P/p1689613020560539


julien-c (commented):
i would still have plugged a jsonschema validator in the Deno code

but i guess this works too

PR #1317, opened 2023-07-18
Do not refer to Llama 2 as open source

aphedges:
The Llama 2 license (<https://ai.meta.com/llama/license/>) is not open source, so the models should not be referred to as such. I used wording from earlier in the paragraph that is more appropriate.

A blog post from Hugging Face yesterday (<https://huggingface.co/blog/os-llms>) states that Llama isn't open source, so this change makes the Llama 2 blog post consistent with it.


osanseviero (approved):
Thank you!

PR #1322, opened 2023-07-19
Add link to paper page

osanseviero:


pcuenca (approved):

chenrq2005:
could you please talk about  what  are the key difference between models with -hf suffix and without -hf?

osanseviero:
The `hf` are supported in `transformers`

PR #1359, opened 2023-08-01
3d assets pr

dylanebert:


julien-c:
at some point we could have a way to embed a 3D scene (gltf/glb) in the blog, given we already have that component in the website (https://huggingface.co/datasets/allenai/objaverse/blob/main/glbs/000-004/0031ba19d3e042c4bcf79eba40ccc812.glb for instance)

PR #1361, opened 2023-08-02
New case studies

VioletteLepercq:


julien-c (commented):
IMO it'd be better to cross-post the content from the original posts here.

Redirecting to external URLs was more like a legacy historical thing, we don't really do it anymore

WDYT @osanseviero @Pierrci ?

osanseviero:
Agree!

_blog.yml
```diff
@@ -2571,3 +2571,37 @@
...
+
+- local: databricks-case-study
+  title: "Databricks ❤️ Hugging Face: up to 40% faster training and tuning of Large Language Models"
+  author: alighodsi
+  guest: true
+  author: maddiedawson
+  guest: true
+  date: April 26, 2023
+  tags:
+    - case-studies
```
Pierrci, line 2607: this means the blog posts will all appear on top of hf.co/blog at once, not sure that's what we want - the `date` doesn't matter in this regard, only the order does

julien-c: yes let's order them by date

VioletteLepercq: I ordered them :)

_blog.yml
```diff
@@ -2426,6 +2452,13 @@
     - cv
     - hardware
 
+- local: writer-case-study
```
Pierrci, line 2455: I think we need to add `thumbnail` too for each of them (in addition to each markdown file as you already did), otherwise, they won't have a thumbnail when appearing on hf.co/blog

_blog.yml
```diff
@@ -2034,6 +2050,16 @@
     - partnerships
     - community
 
+- local: databricks-case-study
+  title: "Databricks ❤️ Hugging Face: up to 40% faster training and tuning of Large Language Models"
+  author: alighodsi
+  guest: true
+  author: maddiedawson
```
Pierrci, line 2057: there can only be one author here (contrary to the markdown file itself, where we can add as many as we want)

VioletteLepercq: changes done

_blog.yml
VioletteLepercq: changes done

VioletteLepercq:
all checks have passed and there is no conflicts with the base branch, does that mean I can squash and merge?

_blog.yml
```diff
@@ -1792,6 +1792,15 @@
     - fine-tuning
     - community
     - dreambooth
+    
+- local: mantis-case-study
+  title: "Why we’re switching to Hugging Face Inference Endpoints, and maybe you should too"
+  author: mattupson
+  guest: true
+  thumbnail: /blog/assets/78_ml_director_insights/mantis1.png
```
julien-c, line 1800: the mantis.png one doesn't seem to be used, can you remove it?

snorkel-case-study.md
```diff
@@ -0,0 +1,52 @@
+---
+title: "Snorkel AI x Hugging Face: unlock foundation models for enterprises"
+thumbnail: /blog/assets/78_ml_director_insights/snorkel.png
+authors:
+- user: Friea Berg
```
julien-c, line 5: ^needs to be a Hub user account

VioletteLepercq: updated

_blog.yml
VioletteLepercq: removed :)

Pierrci: I think what @julien-c meant is that there are two PNGs for mantis in this PR: `mantis.png` and `mantis1.png`, we should only keep one (and set back the one we keep as `thumbnail` here)

VioletteLepercq: 🙈 deleted the mantis.png file and added back the thumbnails linking to mantis1.png

snorkel-case-study.md
Pierrci: it needs to be changed both in `_blog.yml` (which you did) and in the markdown

genomicsengland-case-study.md
Pierrci, line 1: this one seems to be empty?

VioletteLepercq: edited the markdown

VioletteLepercq: deleted the file

Pierrci: ok then we can delete the PNG associated with it too, and we should be good to go :)

PR #1362, opened 2023-08-03
DPO blog

lvwerra:
The blog highlights how to use the new `DPOTrainer` to do RLHF without RL on the example of training StackLlama with Llama 2.

cc @kashif @younesbelkada


dpo-trl.md
```diff
@@ -0,0 +1,182 @@
...
+---
+
+# Fine-tune Llama 2 with DPO
+
+<!-- {blog_metadata} -->
+<!-- {authors} -->
+
+## Introduction
+
+Reinforcement Learning from Human Feedback (RLHF) has become the de facto last training step of LLMs such as GPT-4 or Claude to ensure that the language model's outputs are aligned with human expectations such as chattiness or safety features. However, it brings some of the complexity of RL into NLP: we need to build a good reward function, train the model to estimate the value of a state, and at the same time be careful not to strive to far from the original model and produce gibberish instead of sensible text. Such a process is quite involved requiring a number of complex moving parts where it is not always easy to get things right.
```
younesbelkada, line 17:
```suggestion
Reinforcement Learning from Human Feedback (RLHF) has become the de facto last training step of LLMs such as GPT-4 or Claude to ensure that the language model's outputs are aligned with human expectations such as chattiness or safety features. However, it brings some of the complexity of RL into NLP: we need to build a good reward function, train the model to estimate the value of a state, and at the same time be careful not to strive too far from the original model and produce gibberish instead of sensible text. Such a process is quite involved requiring a number of complex moving parts where it is not always easy to get things right.
```

dpo-trl.md
```diff
@@ -0,0 +1,182 @@
...
+dpo_trainer.train()
+```
+
+## Experiment with Llama v2
+
+The benefit of implementing the DPO trainer in TRL is that one can take advantage of all the extra bells and whistles of training large LLMs which come with TRL and its dependent libraries like Peft and Accelerate. With these libraries we are even able to train a Llama v2 model using the QLoRA technique provided by the bitsandbytes library. 
+
+### Supervised Fine Tuning
+
+The process as introduced above involves the supervised fine-tuning step using QLoRA on the 7B Llama v2 model on the SFT split of the data via TRL’s `SFTTrainer`:
```
younesbelkada, line 100:
```suggestion
The process as introduced above involves the supervised fine-tuning step using [QLoRA](https://arxiv.org/abs/2305.14314) on the 7B Llama v2 model on the SFT split of the data via TRL’s `SFTTrainer`:
```

dpo-trl.md
```diff
@@ -0,0 +1,182 @@
...
+
+So as can be seen we load the model in the 4-bit configuration and then train it via the QLora method via the `peft_config` arguments. The trainer will also evaluate the progress during training with respect to the evaluation dataset and report back a number of key metrics like the implicit reward which can be recorded and displayed via WandB for example. We can then push the final trained model to the HuggingFace Hub.
+
+## Demo/Example/Conclusion
+
+The full source code of the training scripts for the SFT and DPO are available.
+
+What is the catch then? We do lose the ability to sample from the model when training the RL objective which is a hallmark of RL training in the presence of distributional shifts.
+
+TODO: fill here
```
younesbelkada, line 182:
```suggestion
TODO: fill here
```
👀

younesbelkada (approved):
Very cool! 🔥

dpo-trl.md
```diff
@@ -0,0 +1,187 @@
...
+dpo_trainer.save_model()
+```
+
+So as can be seen we load the model in the 4-bit configuration and then train it via the QLora method via the `peft_config` arguments. The trainer will also evaluate the progress during training with respect to the evaluation dataset and report back a number of key metrics like the implicit reward which can be recorded and displayed via WandB for example. We can then push the final trained model to the HuggingFace Hub.
+
+## Demo/Example/Conclusion
+
+The full source code of the training scripts for the SFT and DPO are available in the following [examples/stack_llama_2](https://github.com/lvwerra/trl/tree/main/examples/research_projects/stack_llama_2) directory for the keen readers to try out themselves.
+
+The WandB logs for the DPO training run can be found [here](https://wandb.ai/krasul/huggingface/runs/c54lmder).
```
vwxyzjn, line 181: One thing maybe helpful is to have an explanation of logged metrics. For example, I don't know what `train/rewards/margins` mean. Maybe we can add a section to DPO docs and link it here. They should look something like 


* `charts/episodic_return`: episodic return of the game
* `charts/episodic_length`: episodic length of the game
* `charts/SPS`: number of steps per second
* `charts/learning_rate`: the current learning rate
* `losses/value_loss`: the mean value loss across all data points
* `losses/policy_loss`: the mean policy loss across all data points
* `losses/entropy`: the mean entropy value across all data points
* ...

dpo-trl.md
```diff
@@ -0,0 +1,187 @@
...
+)
+dpo_trainer.train()
+dpo_trainer.save_model()
+```
+
+So as can be seen we load the model in the 4-bit configuration and then train it via the QLora method via the `peft_config` arguments. The trainer will also evaluate the progress during training with respect to the evaluation dataset and report back a number of key metrics like the implicit reward which can be recorded and displayed via WandB for example. We can then push the final trained model to the HuggingFace Hub.
+
+## Demo/Example/Conclusion
+
+The full source code of the training scripts for the SFT and DPO are available in the following [examples/stack_llama_2](https://github.com/lvwerra/trl/tree/main/examples/research_projects/stack_llama_2) directory for the keen readers to try out themselves.
```
vwxyzjn, line 179: Were there experiments replicating the results in the original setup with hh in https://wandb.ai/eric_anthony_mitchell/dpo-demos/runs/og8q3euz.

kashif: so in theory yes but here we are using llama-v2 on the stack-exchange dataset...

dpo-trl.md
kashif: ah yes good idea... let me add that

dpo-trl.md
```diff
@@ -0,0 +1,187 @@
...
+
+The DPO formulation bypasses the reward modeling step and directly optimises the language model on preference data via a key insight: namely an analytical mapping from the reward function to the optimal policy that enables the authors to transform the RL loss over the reward  and reference models to a loss over the reference model directly!
+
+Thus this direct likelihood object can be optimized without the need for a reward model or the need to perform the potentially fiddly RL based optimisation.
```
pacman100, line 29:
```suggestion
Thus this direct likelihood objective can be optimized without the need for a reward model or the need to perform the potentially fiddly RL based optimisation.
```

dpo-trl.md
```diff
@@ -0,0 +1,187 @@
...
+
+```python
+model = AutoPeftModelForCausalLM.from_pretrained(
+    script_args.model_name_or_path, # location of saved SFT model
+    low_cpu_mem_usage=True,
+    torch_dtype=torch.float16,
+    load_in_4bit=True,
+    is_trainable=True,
+)
+model_ref = AutoPeftModelForCausalLM.from_pretrained(
```
pacman100, line 154: why not just use `model.disable_adapter` context manager for the reference model?
It will save a lot of GPU memory.

dpo-trl.md
```diff
@@ -0,0 +1,187 @@
...
+
+## DPO vs PPO
+
+In the traditional model of optimising human derived preferences via RL, the goto method has been to use an auxiliary reward model and fine-tune the model of interest so that it maximizes this given reward via the machinery of RL. Intuitively we use the reward model to provide feedback to the model we are optimising so that it generates high-reward samples more often and low-reward samples less often. At the same time we use a frozen reference model to make sure that whatever is generated does not deviate too much and continues to maintain generation diversity. This is typically done by adding a KL penalty to the full reward maximisation objective via a reference model, which serves to prevent the model from learning to cheat or exploit the reward model.
+
+The DPO formulation bypasses the reward modeling step and directly optimises the language model on preference data via a key insight: namely an analytical mapping from the reward function to the optimal policy that enables the authors to transform the RL loss over the reward  and reference models to a loss over the reference model directly!
```
pacman100, line 27: Maybe a little more detail into how to arrive at the mapping and how it ties back with optimal policy that the RLHF learns.

dpo-trl.md
vwxyzjn: Added some examples at https://huggingface.co/docs/trl/main/en/logging#ppo-logging

kashif: added to the code PR

PR #1378, opened 2023-08-10
Fix grammar / typos

mrm8488:


younesbelkada (approved):
Thanks a lot @mrm8488 !

PR #1425, opened 2023-08-28
Add some info on dtypes

osanseviero:


codellama.md
```diff
@@ -94,6 +94,15 @@ Until `transformers` 4.33 is released, please install it from the main branch.
 ```bash
 !pip install git+https://github.com/huggingface/transformers.git@main accelerate
 ```
+#### A Note on dtypes
+
+When using models like Code Llama, it's important to take a look at the data types of the models. Code Llama was trained in `bfloat16
```
pcuenca, line 99:
```suggestion
When using models like Code Llama, it's important to take a look at the data types of the models. Code Llama was trained in `bfloat16`
```

pcuenca: Was this sentence inadvertently truncated? I feel the bullet points afterwards are a bit abrupt, not sure if you wanted to add a longer intro.

codellama.md
```diff
@@ -94,6 +94,15 @@ Until `transformers` 4.33 is released, please install it from the main branch.
...
+
+* 32-bit floating point (`float32`): PyTorch convention on model initialization is to load models in `float32`. `transformers` also follows this convention.
```
pcuenca, line 101:
```suggestion
* 32-bit floating point (`float32`): PyTorch convention on model initialization is to load models in `float32`, no matter what the weight files contain. `transformers` also follows this convention for consistency with PyTorch.
```

codellama.md
```diff
@@ -94,6 +94,15 @@ Until `transformers` 4.33 is released, please install it from the main branch.
...
+* 16-bit Brain floating point (`bfloat16`): Code Llama was trained with this precision.
```
pcuenca, line 102:
```suggestion
* 16-bit Brain floating point (`bfloat16`): Code Llama was trained with this precision, so we recommend to use it for further training or fine-tuning.
```

codellama.md
```diff
@@ -94,6 +94,15 @@ Until `transformers` 4.33 is released, please install it from the main branch.
...
+* 16-bit floating point (`float16`): Code Llama is supposed to run inference with this precision.
```
pcuenca, line 103:
```suggestion
* 16-bit floating point (`float16`): We recommend to run inference using this precision, as it's usually faster than `bfloat16` and evaluation metrics show no discernible degradation with respect to `bfloat16`. You can also run inference using `bfloat16`, and we recommend you check inference results with both `float16` and `bfloat16` after fine-tuning.
```

codellama.md
```diff
@@ -94,6 +94,15 @@ Until `transformers` 4.33 is released, please install it from the main branch.
...
+
+As mentioned above, `transformers` loads using `float32` (no matter with which precision the models are stored), so it's important to specify the dtype when loading the models. If you want to fine-tune Code Llama, it's recommended to use `bfloat16`. Using `float16` can lead to overflows and NaNs. Similarly, if you run inference, it's recommended to use `float16`. Using `bfloat16` can lead to a slower inference.
```
pcuenca, line 105:
```suggestion
As mentioned above, `transformers` loads weights using `float32` (no matter with which precision the models are stored), so it's important to specify the desired `dtype` when loading the models. If you want to fine-tune Code Llama, it's recommended to use `bfloat16`, as using `float16` can lead to overflows and NaNs. If you run inference, we recommended to use `float16` because `bfloat16` can be slower. 
```

pcuenca: One additional question I have is: why do we convert the original weights to `float16`? If they were produced in `bfloat16`, shouldn't we keep them like that and still recommend to run inference in `float16`? I understand the difference is small in this case, but what's the drawback in keeping the original precision?

pcuenca (commented):
Thanks a lot, I like the idea to introduce this section! My suggestions are totally optional, but I wouldn't state that `float16` is the _expected_ way to run the model.

codellama.md
osanseviero: Maybe @ArthurZucker can shed some light on conventions for this

pcuenca (approved):
Thank you!

codellama.md
ArthurZucker: Actually the convention is more historic (we kind of always uploaded the checkpoints of Llama1 to fp16, same with other models) and there are no real degradation if you are training again the model. Though uploading to original dtype is better I guess, we did not really care since what is important is to respect the dtype of training. And given how people were saying `"Inference is slow"` I'd rather have fast inference for people not used to code, and training issues for people who are able to train and just need to cast the model to the correct dtype 😉 

Would be great to be able to store both versions on the hub @osanseviero, with a revision or tag this way people could choose from either 👍🏻

pcuenca: @ArthurZucker that makes a lot of sense. We use `variant`s in diffusers, i.e.: https://huggingface.co/stabilityai/stable-diffusion-xl-base-1.0/tree/main/unet (see the `fp16` vs normal safetensors), and the same mechanism exists in transformers: https://github.com/huggingface/transformers/pull/21332

PR #1439, opened 2023-08-31
Update classification-use-cases.md

VioletteLepercq:


VioletteLepercq:
updated only the blog post thumbnail

PR #1456, opened 2023-09-06
Add correct hardware requirements number for falcon 180B

younesbelkada:
I am not sure the numbers for hardware requirements are correct as for half precision models you need 2x the number of parameters thus 340GB in the case of falcon 180B, with respect to GPTQ it should be around 90GB I believe (180 / 2 + 5-10% overhead for the quantization statistics - maybe I am wrong here but it is not 320GB for sure) 

Spotted in: https://twitter.com/Yampeleg/status/1699417031185981673?s=20

cc @pacman100 @philschmid @osanseviero


philschmid:
I added a sentence to make it more clear. That those are the setups we tested.  

FYI 90GB is not enough @pcuenca had issues running it on the DGX which has 8x 40GB.

younesbelkada:
Thanks! closing this PR as bf61c7611f265eb9b9cfe646198702541a83afb9 being pushed

PR #1471, opened 2023-09-08
Update bertopic.md: replaced all instances of RLFH to RLHF

tripathiaakash:
Reinforcement Learning with Human Feedback is RLHF rather than RLFH. Corrected the typo


pcuenca (approved):
Thanks a lot!

PR #1507, opened 2023-09-15
Make optimize LLM public

patrickvonplaten:


vincentmin:
@patrickvonplaten Thanks for the blog, it's very informative. I wanted to point out a small error in the blog:
In this snippet of code:
```
past_key_values = None # past_key_values is the key-value cache
generated_tokens = []
next_token_id = tokenizer(prompt, return_tensors="pt")["input_ids"].to("cuda")

for _ in range(5):
  next_logits, past_key_values = model(next_token_id, past_key_values=past_key_values, use_cache=True).to_tuple()
  next_logits = next_logits[:, -1:]
  next_token_id = torch.argmax(next_logits, dim=-1)

  print("shape of input_ids", input_ids.shape)
  print("length of key-value cache", len(past_key_values[0][0]))  # past_key_values are of shape [num_layers, 0 for k, 1 for v, batch_size, length, hidden_dim]
  generated_tokens.append(next_token_id.item())

generated_text = tokenizer.batch_decode(generated_tokens)
generated_text
```
you print `input_ids.shape`, where I think you meant to print `next_token_id.shape`. `input_ids` is computed in the previous code block and is no longer relevant here.

patrickvonplaten:
You're perfectly right! Thanks for pointing this on out

PR #1510, opened 2023-09-16
Replace feature extractor by image processor

NielsRogge:
I saw a lot of our blogs are using the deprecated `FeatureExtractor` class of Transformers for vision models, which is now called `ImageProcessor`.

Hence this PR updates and replaces all blogs to the up-to-date class.

cc @amyeroberts @rafaelpadilla


osanseviero (approved):
Thanks!

PR #1569, opened 2023-10-10
Remove broken embedded space

dylanebert:
Removing the embedded remote viewer space since the page still gets a lot of traffic, and the remote viewer space will not be indefinitely maintained


julien-c (commented):
would keeping a hyperlink to it make sense?

dylanebert:
I considered it, but removed the link there since it's still present lower in the post, where it's more forgivable if it's broken

pcuenca (approved):
Accepting since it's already broken :( Would be nice to find a longer-term solution if we can.

PR #1598, opened 2023-10-19
Attempt to fix rendering, remove metadata

pcuenca:
There's a 500 error now (missing quotes in the title, I think).


abidlabs:
Ah thank you so much @pcuenca!

abidlabs (approved):

PR #1602, opened 2023-10-23
[Whisper] Tip for ST

sanchit-gandhi:
The ASR fine-tuning script can also be used for ST with one argument change in the tokenizer. Here, we provide a tip on how to do so.

### DIFF
--- a/fine-tune-whisper.md
+++ b/fine-tune-whisper.md
@@ -357,7 +357,7 @@ from transformers import WhisperTokenizer
 tokenizer = WhisperTokenizer.from_pretrained("openai/whisper-small", language="Hindi", task="transcribe")
 ```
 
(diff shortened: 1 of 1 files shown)


fine-tune-whisper.md
```diff
@@ -357,6 +357,8 @@ from transformers import WhisperTokenizer
 tokenizer = WhisperTokenizer.from_pretrained("openai/whisper-small", language="Hindi", task="transcribe")
 ```
 
+> **Tip:** the blog post can be adapted for *speech translation* by setting the task to `"translate"` and the language to the target text language in the above line. This will prepend the relevant task and language tokens for speech translation when the dataset is pre-processed.
```
sanchit-gandhi, line 360: This is how it renders: 
Screenshot 2023-10-23 at 15 58 29

Vaibhavs10 (approved):
Dope! Thanks for the quick PR.

PR #1618, opened 2023-10-27
Add Zephyr 7B beta to Inference for PROs

pcuenca:


merveenoyan (approved):

pcuenca:
Merging for now, can tweak later cc @osanseviero

inference-pro.md
```diff
@@ -34,10 +34,11 @@ Hugging Face PRO users now have access to exclusive API endpoints for a curated
 
 ## Supported Models
 
-In addition to thousands of public models available in the Hub, PRO users get free access to the following state-of-the-art models:
+In addition to thousands of public models available in the Hub, PRO users get free access and higher rate limits to the following state-of-the-art models:
```
julien-c, line 37: well non-PRO don't have access at all so the higher rate limits part is a bit superfluous

pcuenca: I thought Zephyr 7B is [available to all users](https://github.com/huggingface/api-inference/blob/88951dd693c61ccd113a22342a5aa6c9fbbae75a/master-rs/custom_config.yml#L646), isn't it? That was the reason I included the higher rate limits part, because it's not exclusive to PRO users. But we can remove it, no problem.

julien-c: ah yes ok i see

PR #1641, opened 2023-11-09
Swap thumbnail on LCM blog

apolinario:


pcuenca (approved):

julien-c:
you need to change the filename to bypass CDN caching

pcuenca:
Makes sense, thought about it after the fact.

PR #1650, opened 2023-11-17
Remove the incorrect MFU computation

pacman100:
### What does this PR do?
1. Remove the incorrect MFU computation. Detailed explanations and experiments with respect to this are done in https://github.com/huggingface/blog/issues/1649#issuecomment-1815918168.


sayakpaul (approved):
Thank you!

PR #1686, opened 2023-12-06
PR requested changes

laikhtewari:


pcuenca (approved):
Looks good to me, cc @mfuntowicz, @jeffboudier

optimum-nvidia.md
```diff
@@ -30,11 +30,13 @@ If you already set up a pipeline from Hugging Face’s transformers library to r
...
 ```
-You can also enable FP8 quantization with a single flag, which allows you to run a bigger model on a single GPU, at faster speeds, and without sacrificing accuracy. 
+You can also enable FP8 quantization with a single flag, which allows you to run a bigger model on a single GPU at faster speeds and without sacrificing accuracy. 
 The flag shown in this example uses a predefined calibration strategy by default, though you can provide your own calibration dataset and customized tokenization to tailor the quantization to your use case.
 
+The pipeline interface is great for getting up and running quickly, but power users who want fine-grained control over setting sampling parameters can use the Model API. 
+
 ```diff
-- from transformers import LlamaForCausalLM
+- from transformers import AutoModelForCausalLM
```
jeffboudier, line 39: @mfuntowicz can you double check this change makes sense?

mfuntowicz: Yep, makes sense

mfuntowicz (approved):
LGTM - Thanks @laikhtewari

PR #1711, opened 2023-12-15
Mixtral: Quant fixes and improvements

pcuenca:
- Updated the TOC and disclaimers.
- Some tweaks and fixes.
- Change memory requirements, I could not reproduce :(

### DIFF
--- a/mixtral.md
+++ b/mixtral.md
@@ -249,7 +249,7 @@ output = model.generate(**inputs, max_new_tokens=50)
 print(tokenizer.decode(output[0], skip_special_tokens=True))
 ```
 
-Read more about QLoRA in the bitsandbytes 4-bit quantization section of [the documentation](https://huggingface.co/docs/transformers/quantization#4-bit).
+This 4-bit quantization technique was introduced in the [QLoRA paper](https://huggingface.co/papers/2305.14314), you can read more about it in the corresponding section of [the documentation](https://huggingface.co/docs/transformers/quantization#4-bit) or in [this post](https://huggingface.co/blog/4bit-transformers-bitsandbytes).
 
 ### Load Mixtral with GPTQ
 
--- a/mixtral.md
+++ b/mixtral.md
@@ -34,6 +34,9 @@ Among the features and integrations being released today, we have:
   - [Using 🤗 Transformers](#using-🤗-transformers)
   - [Using Text Generation Inference](#using-text-generation-inference)
 - [Fine-tuning with 🤗 TRL](#fine-tuning-with-🤗-trl)
+- [Quantizing Mixtral](#quantizing-mixtral)
+  - [Load Mixtral with 4-bit quantization](#load-mixtral-with-4-bit-quantization)
+  - [Load Mixtral with GPTQ](#load-mixtral-with-gptq)
 - [Disclaimers and ongoing work](#disclaimers-and-ongoing-work)
 - [Additional Resources](#additional-resources)
 - [Conclusion](#conclusion)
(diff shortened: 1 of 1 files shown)


mixtral.md
```diff
@@ -34,6 +34,9 @@ Among the features and integrations being released today, we have:
   - [Using 🤗 Transformers](#using-🤗-transformers)
   - [Using Text Generation Inference](#using-text-generation-inference)
 - [Fine-tuning with 🤗 TRL](#fine-tuning-with-🤗-trl)
+- [Quantizing Mixtral](#quantizing-mixtral)
+  - [Load Mixtral with 4-bit quantization](#load-mixtral-with-4-bit-quantization)
+  - [Load Mixtral with GPTQ](#load-mixtral-with-gptq)
```
pcuenca, line 39: Added to the TOC here :)

mixtral.md
```diff
@@ -216,15 +219,15 @@ accelerate launch --config_file examples/accelerate_configs/multi_gpu.yaml --num
 
 This takes about 48 hours to train on a single A100, but can be easily parallelised by tweaking `--num_processes` to the number of GPUs you have available.
 
-## Quantizing Mixtral-7B
+## Quantizing Mixtral
```
pcuenca, line 222: I think it's less confusing without the 7B, as this is the MoE model.

mixtral.md
```diff
@@ -216,15 +219,15 @@ accelerate launch --config_file examples/accelerate_configs/multi_gpu.yaml --num
...
 
 As seen above, the challenge for this model is to make it run on consumer-type hardware for anyone to use it, as the model requires ~90GB just to be loaded in half-precision (`torch.float16`).
 
 With the 🤗 transformers library, we support out-of-the-box inference with state-of-the-art quantization methods such as QLoRA and GPTQ. You can read more about the quantization methods we support in the [appropriate documentation section](https://huggingface.co/docs/transformers/quantization). 
 
-### Load Mixtral with QLoRA
+### Load Mixtral with 4-bit quantization
```
pcuenca, line 228: I may be confused, but I thought QLoRA is the name of the fine-tuning technique? The 4-bit quantization and nf4 format are introduced in the paper, but it sounds strange to me to use the term `QLoRA` when we are not doing fine-tuning. Happy to revert if I'm wrong or this is an accepted term in the community :)

mixtral.md
```diff
@@ -216,15 +219,15 @@ accelerate launch --config_file examples/accelerate_configs/multi_gpu.yaml --num
...
 
-As demonstrated in the inference section, you can load Mixtral with 4-bit quantization by installing `bitsandbytes` library `pip install -U bitsandbytes` and passing the flag `load_in_4bit=True` to the `from_pretrained` method. For better performance, we advise users to load the model with `bnb_4bit_compute_dtype=torch.float16`. Note you need a GPU device with at least 24GB VRAM to properly run the snippet below.
+As demonstrated in the inference section, you can load Mixtral with 4-bit quantization by installing the `bitsandbytes` library (`pip install -U bitsandbytes`) and passing the flag `load_in_4bit=True` to the `from_pretrained` method. For better performance, we advise users to load the model with `bnb_4bit_compute_dtype=torch.float16`. Note you need a GPU device with at least 30GB VRAM to properly run the snippet below.
```
pcuenca, line 230: I could not reproduce with less than ~28 GB of GPU RAM, unfortunately. I updated to transformers @ `main`, latest bitsandbytes and optimum :(

mixtral.md
```diff
@@ -275,19 +278,19 @@ tokenizer = AutoTokenizer.from_pretrained(model_id)
...
-prompt = "[/INST] Explain what a Mixture of Experts is in less than 100 words. [/INST]"
+prompt = "[INST] Explain what a Mixture of Experts is in less than 100 words. [/INST]"
 inputs = tokenizer(prompt, return_tensors="pt").to(0)
 
 output = model.generate(**inputs, max_new_tokens=50)
 print(tokenizer.decode(output[0], skip_special_tokens=True))
 ```
 
-Note both for QLoRA and GPTQ you need to allocate at least 24GB GPU VRAM to fit the model and make sure inference runs properly.
+Note that for both QLoRA and GPTQ you need at least 30 GB of GPU VRAM to fit the model. You can make it work with 24 GB if you use `device_map="auto"`, like in the example above, so some layers are offloaded to CPU.
```
pcuenca, line 288: Same when using GPTQ. In my case, using `device_map="auto"` with just a single 24GB GPU enabled in `CUDA_VISIBLE_DEVICES` offloaded part of the model to CPU. Very happy to revert if there's a way to fix this!

mixtral.md
younesbelkada: Makes sense, 24GB was a ball-park calculation given the number of params, let's put 30GB then!

mixtral.md
younesbelkada: QLoRA is the name of the paper that intorduces the 4-bit quant method, I agree we should leave it to 4-bit then - but we can maybe just mention the paper below: https://arxiv.org/abs/2305.14314

younesbelkada (approved):
Thanks a lot @pcuenca !!
I just left one suggestion to mention the QLoRA paper: https://arxiv.org/abs/2305.14314

pcuenca:
Thanks for the suggestion @younesbelkada, and for updating the post with your findings in the first place :) Merging now.

younesbelkada:
Awesome, thanks a lot again @pcuenca !

PR #1713, opened 2023-12-15
Correct typo; image caption staying opposite of ST-MoE paper

pdtgct:
The paragraph preceding the image from the ST-MoE paper quotes the paper verbatim as saying that sparse models fine tune better with smaller batch sizes and higher learning rates but the caption states the opposite.


osanseviero (approved):
Thanks!

PR #1718, opened 2023-12-20
[Whisper Spec Dec] Bump video width

sanchit-gandhi:
The video's on the Whisper speculative decoding blog post are rendering well, but are fractionally too small. This PR bumps the video width to 90% (which matches that used by @gante in the assisted generation blog post)


pcuenca (approved):

pcuenca:
Thank you!

PR #1742, opened 2024-01-14
fix: translation in moe.md

alephpi:
chinese translation in figure caption is not consistent with the paragraph above


pcuenca:
cc @zhongdongy

PR #1748, opened 2024-01-16
Fix link in HTML div environment

meg-huggingface:
Really thoughtful addition to the blog post. Want to make sure people can link to it ok.


clefourrier (approved):
LGTM! cc @pcuenca , apparently markdown is not rendered properly in the block (which makes sense but I had not realized)

clefourrier:
Thanks a lot @meg-huggingface  :)

PR #1777, opened 2024-01-29
Changing the thumbnail of leaderboards on the hub

clefourrier:
At least it will be the correct size/shape ^^


osanseviero (approved):

osanseviero:
Should we close this?

clefourrier:
Yep, let's!

PR #1785, opened 2024-01-31
Point to PR with IPO fix

lewtun:
Adds a clarifying remark about the root cause of the bug and a pointer to the TRL PR so the community understands better what changed.

### DIFF
--- a/pref-tuning.md
+++ b/pref-tuning.md
@@ -13,7 +13,7 @@ authors:
 
 **Addendum**
 
-After consulting with the authors of the IPO paper, we have discovered that the implementation of IPO in TRL was incorrect. We have since fixed this and re-run the experiments. The results are now consistent with the paper, with IPO on par with DPO and performing better than KTO in the paired preference setting. We have updated the post to reflect these new results.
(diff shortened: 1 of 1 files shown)


edbeeching (approved):
Thanks @lewtun , LGTM

kashif (approved):

lifengjin:
Just curious (and also slightly confused): where in paper does it mention that the loss needs to be averaged across tokens?

JubilantJerry:
I'm also confused about this claim that the loss needs to be averaged. Page 7 of https://arxiv.org/pdf/2310.12036.pdf clearly shows that the IPO loss is defined based on $\pi(y \mid x)$, matching their notation on page 3 for their recap on the DPO loss. I didn't see them formally define $\pi(y|x)$, but I'm pretty sure it needs to be interpreted as a probability of a sequence. The log of the probability of a sequence can be found by summing the log probability of each token, but not by taking the average of those token probabilities, the average has no meaningful interpretation. Unless I missed something from the paper?

lewtun:
> @lifengjin The summed operation makes the constant regularization / beta depend on the length of the response. Averaged can fix it.

Indeed this is what the DeepMind authors pointed out to us in the initial version of the blog post. I don't believe it is explicitly mentioned in the paper, which is why we overlooked this detail

PR #1796, opened 2024-02-01
Add optimum code snippet to load IPEX models

echarlaix:


intel-starcoder-quantization.md
```diff
@@ -142,3 +142,28 @@ These quantization choices are backed up by the following observations:
...
+```diff
+- from transformers import AutoModelForCausalLM
++ from optimum.intel import IPEXModelForCausalLM
+  from transformers import AutoTokenizer, pipeline
+
+  model_id = "Intel/q8_starcoder"
+- model = AutoModelForCausalLM.from_pretrained(model_id)
++ model = IPEXModelForCausalLM.from_pretrained(model_id)
+  tokenizer = AutoTokenizer.from_pretrained(model_id)
+  pipe = pipeline("text-generation", model=model, tokenizer=tokenizer)
```
ofirzaf, line 167: Do the pipelines support AG? We should show an example with it no? We might have a problem there also because AG needs transformers v4.37 while I saw IPEX depends on 4.31. Is there a reason for this dependency on an old transformers?

echarlaix: Yes it can support AG as well, here it's just to show how to load the IPEX model, we can add a second code snippet for AG

echarlaix: v4.37 is supported as well

ofirzaf: I think that it will be good to show how to use AG. I will let you know when we can release the models

pcuenca (approved):

intel-starcoder-quantization.md
echarlaix: we can add it after the models release

PR #1801, opened 2024-02-02
Display equations below the text, centered

pcuenca:
Otherwise the preceding line is centered vertically with the equation and looks a bit weird.


clefourrier (approved):
Thanks a lot :)

PR #1818, opened 2024-02-16
Reorder yaml

pcuenca:
Not sure if this will fix the ordering in the main page


MoritzLaurer (approved):
thanks for this pull request! yeah, it's possible that the ordering caused the wrong ordering on the website, my bad

PR #1822, opened 2024-02-18
[Optimize LLM blog] Fix typos

NielsRogge:
Was reading the awesome blog by @patrickvonplaten, but noticed that there was a typo, and the past_key_values seem to have a different shape.

This PR fixes that :)


patrickvonplaten (approved):
Thanks!

PR #1829, opened 2024-02-19
add links for whisper large v2 and v3 to the finetune article

iitsg:
As I was reading through your finetune article for Whisper, I saw that both `large-v2` and `large-v3` are missing in the table showcase.
I have added these, so new readers are able to find both versions way easier.

This is especially helpful, as `large-v3` is still missing from both `large` and `large-v2`'s model cards.


sanchit-gandhi:
Thanks for the PR @iitsg, and sorry I'm late coming to this. The changes have been implemented as part of #1944. Feel free to take a look!

PR #1840, opened 2024-02-21
Remove copypasta from Mixtral blog post

milocress:
https://huggingface.co/blog/mixtral

> Note that unlike dense transformers, one should not target the MLP layers as they are sparse and don’t interact well with PEFT.

This applies to mixtral, which is a mixture of experts, but not to gemma, which is a dense model.


pcuenca (approved):
Looks good to me!

pcuenca:
cc @philschmid for verification

osanseviero (approved):

PR #1850, opened 2024-02-22
Fix some typos

mrm8488:


synthetic-data-save-costs.md
```diff
@@ -356,9 +356,9 @@ Note that in the [full reproduction script](https://github.com/MoritzLaurer/synt
...
+The main advantage of this data created with the open-source Mixtral model is that the data is fully commercially usable without legal uncertainty. For example, data created with the OpenAI API is subject to the [OpenAI Business Terms](https://openai.com/policies/business-terms), which explicitly prohibits using model outputs for training models that compete with their products and services. The legal value and meaning of these Terms are unclear, but they introduce legal uncertainty for the commercial use of models trained on synthetic data from OpenAI models. Any smaller, efficient model trained on synthetic data could be considered as competing, as it reduces dependency on the API service.
```
MoritzLaurer, line 359: I'd keep "prohibit" here, because it refers to the plural "Terms"

MoritzLaurer (approved):
Thanks @mrm8488 for spotting and correcting the typos! I would just revert back to "prohibit" because it refers to the plural "Terms"

synthetic-data-save-costs.md
pcuenca, line 359:
```suggestion
The main advantage of this data created with the open-source Mixtral model is that the data is fully commercially usable without legal uncertainty. For example, data created with the OpenAI API is subject to the [OpenAI Business Terms](https://openai.com/policies/business-terms), which explicitly prohibit using model outputs for training models that compete with their products and services. The legal value and meaning of these Terms are unclear, but they introduce legal uncertainty for the commercial use of models trained on synthetic data from OpenAI models. Any smaller, efficient model trained on synthetic data could be considered as competing, as it reduces dependency on the API service. 
```

pcuenca (approved):
Thanks @mrm8488 !

PR #1869, opened 2024-02-26
Watermarking: fig. 3

pcuenca:


sashavor (approved):

pcuenca:
cc @mishig25 this change (Figure 3 includes an HTML link in the caption) is not visible in the blog post after a few minutes, I checked in Codespaces and it seemed to render there. Do you think there may be something off?

PR #1872, opened 2024-02-27
Update author's hf handle.

chenglu:
@adeenayakup FYI


adeenayakup (approved):
Thanks for the update!!

PR #1881, opened 2024-02-28
Update the blog url of starcoder2.md in zh/

chenglu:
@adeenayakup


adeenayakup (approved):
Thank you @chenglu!!🙌

PR #1887, opened 2024-03-01
Delete zh/StarCoder2

pcuenca:
Note the capitalization. There are two different files `StarCoder2.md` and `starcoder2.md` that are causing problems on case insensitive but case preserving filesystems (default apfs on Mac).

- `StarCoder2` was created in 435833e
- Updates were applied to `starcoder2` instead in 3779a8d and 66cc5d7.
- The blog looks correct in https://huggingface.co/blog/zh/starcoder2. Note the translator and banner image are visible, but not in https://huggingface.co/blog/zh/StarCoder2.

Therefore, I think `StarCoder2` is outdated and has to be deleted.

cc @chenglu @adeenayakup for content verification


chenglu:
Yes, please delete the file StarCoder2.md, thanks!

chenglu (approved):

xianbaoqian (approved):
Thanks for catching the issue!

PR #1895, opened 2024-03-11
Update Llama table with new evals

osanseviero:
We added Winogrande and GSM8K, so the LLM leaderboard score was outdated


pcuenca:
Should we add a tip about the scores being updated?

osanseviero:
Done, good idea!

PR #1940, opened 2024-03-26
Update mixtral.md

saahil1801:
Exllama kernels using GPTQConfig for faster inference and production load. @davanstrien  @younesbelkada @pcuenca

### DIFF
--- a/mixtral.md
+++ b/mixtral.md
@@ -295,8 +295,11 @@ model_id = "TheBloke/Mixtral-8x7B-v0.1-GPTQ"
 tokenizer = AutoTokenizer.from_pretrained(model_id)
 
 gptq_config = GPTQConfig(bits=4, use_exllama=True)
-model = AutoModelForCausalLM.from_pretrained(model_id,quantization_config=gptq_config,
-                                             device_map="auto")
+model = AutoModelForCausalLM.from_pretrained(
+    model_id,
+    quantization_config=gptq_config,
+    device_map="auto"
+)
 
 prompt = "[INST] Explain what a Mixture of Experts is in less than 100 words. [/INST]"
 inputs = tokenizer(prompt, return_tensors="pt").to(0)


mixtral.md
```diff
@@ -285,8 +285,30 @@ output = model.generate(**inputs, max_new_tokens=50)
 print(tokenizer.decode(output[0], skip_special_tokens=True))
 ```
 
-Note that for both QLoRA and GPTQ you need at least 30 GB of GPU VRAM to fit the model. You can make it work with 24 GB if you use `device_map="auto"`, like in the example above, so some layers are offloaded to CPU.
+You could also just load the model using a GPTQ configuration setting the desired parameters , as usual when working with transformers .
+For faster inference and production load we want to leverage the [exllama kernels](https://github.com/turboderp/exllama) ( Achieving the same latency as fp16 model, but 4x less memory usage ) .
```
younesbelkada, line 289:
```suggestion
For faster inference and production load we want to leverage the [exllama kernels](https://github.com/turboderp/exllama) (Achieving the same latency as fp16 model, but 4x less memory usage) .
```

mixtral.md
```diff
@@ -285,8 +285,30 @@ output = model.generate(**inputs, max_new_tokens=50)
...
+                                             device_map="auto")
+
+prompt = "[INST] Explain what a Mixture of Experts is in less than 100 words. [/INST]"
+inputs = tokenizer(prompt, return_tensors="pt").to(0)
+
+output = model.generate(**inputs, max_new_tokens=50)
+print(tokenizer.decode(output[0], skip_special_tokens=True))
+```
+
+If left unset , the "use_exllama" parameter defaults to True , enabling the exllama backend functionality, specifically designed to work with the "bits" value of 4 .
```
younesbelkada, line 309:
```suggestion
If left unset , the "use_exllama" parameter defaults to True , enabling the exllama backend functionality, specifically designed to work with the "bits" value of 4.
```

younesbelkada (approved):
LGTM !

mixtral.md
```diff
@@ -285,8 +285,30 @@ output = model.generate(**inputs, max_new_tokens=50)
 print(tokenizer.decode(output[0], skip_special_tokens=True))
 ```
 
-Note that for both QLoRA and GPTQ you need at least 30 GB of GPU VRAM to fit the model. You can make it work with 24 GB if you use `device_map="auto"`, like in the example above, so some layers are offloaded to CPU.
+You could also just load the model using a GPTQ configuration setting the desired parameters , as usual when working with transformers .
+For faster inference and production load we want to leverage the [exllama kernels](https://github.com/turboderp/exllama) (Achieving the same latency as fp16 model, but 4x less memory usage) .
```
pcuenca, line 289:
```suggestion
If you have [exllama kernels installed](https://github.com/turboderp/exllama), you can leverage them to run the GPTQ model. To do so, load the model with a custom GPTQ configuration where you set the desired parameters:
```

mixtral.md
```diff
@@ -285,8 +285,30 @@ output = model.generate(**inputs, max_new_tokens=50)
...
+```python
+import torch
+from transformers
+
+model_id = "TheBloke/Mixtral-8x7B-v0.1-GPTQ"
+tokenizer = AutoTokenizer.from_pretrained(model_id)
+
+gptq_config = GPTQConfig(bits=4, use_exllama=True)
+model = AutoModelForCausalLM.from_pretrained(model_id,quantization_config=gptq_config,
+                                             device_map="auto")
```
pcuenca, line 300:
```suggestion
model = AutoModelForCausalLM.from_pretrained(
    model_id,
    quantization_config=gptq_config,
    device_map="auto"
)
```

This is the style usually used in transformers and docs.

mixtral.md
```diff
@@ -285,8 +285,30 @@ output = model.generate(**inputs, max_new_tokens=50)
...
+
+prompt = "[INST] Explain what a Mixture of Experts is in less than 100 words. [/INST]"
+inputs = tokenizer(prompt, return_tensors="pt").to(0)
+
+output = model.generate(**inputs, max_new_tokens=50)
+print(tokenizer.decode(output[0], skip_special_tokens=True))
+```
+
+If left unset , the "use_exllama" parameter defaults to True , enabling the exllama backend functionality, specifically designed to work with the "bits" value of 4.
```
pcuenca, line 309:
```suggestion
If left unset, `use_exllama` defaults to `True` when kernels are installed.
```

I don't fully follow, sorry. If the backend is designed for 4-bits and `use_exllama` is `True` by default, then it means:
- We can't use any other value (4 bits) in the GPTQConfig.
- Exllama would be enabled anyway if we don't provide the configuration object.

Is that correct? If it is, then I'd simply mention in a paragraph that exllama will be used when installed, and wouldn't provide a code example that might confuse readers.

mixtral.md
```diff
@@ -285,8 +285,30 @@ output = model.generate(**inputs, max_new_tokens=50)
...
+
+Note that for both QLoRA and GPTQ you need at least 30 GB of GPU VRAM to fit the model. You can make it work with 24 GB if you use `device_map="auto"`, like in the example above, so some layers are offloaded to CPU.
```
pcuenca, line 311: Is this also true when exllama is enabled?

pcuenca (commented):
Thank you! I have some questions / doubts about how it works, so I'd suggest we try to dispel them for readers too.

mixtral.md
saahil1801: A better way to rephrase the above , thanks

mixtral.md
saahil1801: Yes, I'll remember that.

mixtral.md
saahil1801: The exllama kernels are passed through the GPTQConfig object.Simply passing the GPTQConfig would do the trick for LLama Based LLMS.

saahil1801: I created the GPTQConfig with other parameters defined 
```` python gptq_config = GPTQConfig(bits=4, use_exllama=True)
````
to help educate readers about some basic parameters in GPTQConfig object.

mixtral.md
saahil1801: Using exllama kernels would significantly reduce only the inferencing speed of the fitted model as it uses 4-bit GPTQ weights for faster computation

PR #1944, opened 2024-03-28
[Whisper Fine-Tuning] Updates and Fixes

sanchit-gandhi:
Updates the Whisper fine-tuning blog post to use the latest API for generate. This should un-block cases where multiple languages were automatically being inferred at inference (e.g. [this issue](https://huggingface.co/distil-whisper/distil-medium.en/discussions/14#6604a6e94b0ad9bc40a72d33)). It also removes references to the now defunct `hf-speech-bench`, and instead compares the performance of the fine-tuned model to the pre-trained one.

TODO:
* [ ] Add pre-trained Whisper small eval results (on-going [here](https://wandb.ai/sanchit-gandhi/distil-whisper-speed-benchmark/runs/lkkc7n5c))

### DIFF
--- a/fine-tune-whisper.md
+++ b/fine-tune-whisper.md
@@ -106,17 +106,19 @@ vocabulary of text tokens.
 
 The Whisper checkpoints come in five configurations of varying model sizes.
 The smallest four are trained on either English-only or multilingual data.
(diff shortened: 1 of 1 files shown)


pcuenca (approved):
Thank you Sanchit! 🙌

PR #1949, opened 2024-04-01
[Whisper Fine-Tuning] Fix decoder start token id

sanchit-gandhi:
Fixes https://github.com/huggingface/transformers/issues/29973, by correctly setting the decoder start token id in the data collator.


pcuenca (approved):
Nice!

PR #2000, opened 2024-04-16
guest articles: link guest authors to their org with a nice UI

julien-c:
i.e. look at the small org logos here:

image


tomaarsen:
Excellent!

One small comment, the borders look a bit awkward with the dark mode background:
image

(That goes both for the org logos and the author pictures)

- Tom Aarsen

PR #2008, opened 2024-04-18
Leaderboard CoT - publication Apr22 (Monday)

clefourrier:
Ready for a light review! :)


leaderboard-cot.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+## How are chain-of-thought traces generated? 
+[#cot-generation]
+
+The literature on chain-of-thought prompting has exploded in the last year, and the wealth of prompting strategies for CoT (involving, e.g., decoding, prompt wording, prompt chains, decomposition, aggregation and revision of reasoning traces) has become [sheerly mindblowing](https://github.com/logikon-ai/awesome-deliberative-prompting/#readme).
+
+To cope with this variety, the evaluation pipeline behind the Open CoT Leaderboard implements CoT generation through modular and easily extensible prompt chains. So far, we have implemented two fairly simple prompting strategies: 
+- Classic: Problem presentation followed by “Let’s think step by step.” instruction.
+- Reflect: Problem presentation followed by instruction to reflect upon the problem in general terms before solving it step by step.
```
pcuenca, line 79: Are these triggered by appropriate changes to the `Reasoning` section of the prompt above?

leaderboard-cot.md
```diff
@@ -0,0 +1,179 @@
...
+- user: clefourrier
+- user: yakazimir
+  guest: true
+  org: allenai
+---
+
+# Introducing the Open Chain of Thought Leaderboard
+[Chain-of-thought prompting](https://huggingface.co/docs/transformers/main/en/tasks/prompting#chain-of-thought) is emerging as a powerful and effective design pattern for LLM-based apps and agents. The basic idea of chain-of-thought prompting is to let a model generate a step-by-step solution (“reasoning trace”) before answering a question or taking a decision. With the Open CoT Leaderboard we’re tracking LLMs’ ability to generate effective chain-of-thought traces for challenging reasoning tasks. 
+
+Unlike most performance based leaderboards, we’re not scoring the absolute accuracy a model achieves on a given task, but the difference between the accuracy with and without chain-of-thought prompting:
```
pcuenca, line 20: But the more powerful the "base" model is, the more difficult it would be for it to improve, wouldn't it? In addition (and this is also mentioned in the post), some models have a natural tendency to follow CoT-like responses. Is this taken into account somehow?

leaderboard-cot.md
```diff
@@ -0,0 +1,179 @@
...
+
+## What are the main take-aways so far?
+
+We’re gradually extending the Open CoT Leaderboard by evaluating more and more models, but current results (model count=30) already suggest some interesting insights.
+
+- Mighty dwarfs: We have been very pleased to see that relatively small (7B) open LLMs are capable of effective, i.e. accuracy-improving chain-of-thought reasoning. 🎉
+- Instruction- and chat-finetuning helps: Finetuned models score much better than their corresponding base models. More specifically, finetuning may improve both the baseline accuracy without CoT and the marginal accuracy gains achieved through CoT.
+- Vindication of *textbooks-is-all-you-need*: The Phi-2 model clearly outperforms the Mixtral MoE base model on the Open CoT Leaderboard.
```
pcuenca, line 156: I'd be careful with this phrasing. This does not mean that Phi-2 is better (or worse), just that it improves more when applying explicit CoT prompting.

leaderboard-cot.md
```diff
@@ -0,0 +1,179 @@
...
+
+Over time, static accuracy-based benchmarks risk to become less informative: Does a model score well because of its superior skill, because it has seen the correct answers during training, or because it has been developed in a competitive context that is governed by this very benchmark? These widely acknowledged issues are addressed by recent eval approaches such as ChatBot arenas, the use of LLMs as judges, or dynamic benchmarks with programmatically generated tasks. The Open CoT Leaderboard contributes to these efforts. It seems more robust to training data contamination because knowing the answer to a question doesn’t ensure that one can reason effectively about it. Plus, the leaderboard’s metric is inversely related to the default metric of other accuracy-based leaderboards (higher accuracy w/o CoT means, all else being equal, lower marginal accuracy gain due to CoT).
```
pcuenca, line 35: Why is this inverse relationship beneficial?

leaderboard-cot.md
```diff
@@ -0,0 +1,179 @@
...
+
+It is now common to distinguish fast and implicit (system-1) from slow and explicit (system-2) reasoning. These two modes are equally important dimensions of human cognition. To understand—for example in debates about AGI—what LLMs are capable of, it seems pivotal to assess LLMs comprehensively along all dimensions of human cognition. In accuracy-based benchmarks without CoT, LLMs arguably solve tasks fast and implicitly. To assess the system-2 reasoning skill of LLMs, we need chain-of-thought based benchmarks. The Open CoT Leaderboard may therefore also inform broader debates about the significance of LLMs and AGI.
```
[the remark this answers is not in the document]
clefourrier: Ouch completely missed this sentence, I agree!

leaderboard-cot.md
```diff
@@ -0,0 +1,179 @@
...
+
+Passage: <passage>
+    
+Question: <question>
+
+A. <choice1>
+B. <choice2>
+…
+
+Reasoning: <reasoning>
```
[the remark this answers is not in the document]
clefourrier: Not sure how italics plays with code typing, I'll do some tests

ggbetz:
Thanks for the great review and feedback!

_blog.yml
```diff
@@ -3822,3 +3822,16 @@
     - open-source
     - guide
     - demo
+
+- local: leaderboard-cot
+  title: "Introducing the Open CoT Leaderboard"
```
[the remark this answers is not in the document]
ggbetz: Good point. I'd be happy to go with "Introducing the Open Chain of Thought Leaderboard" for blog post.

leaderboard-cot.md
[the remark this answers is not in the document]
ggbetz: OK, so you doubt the analogy:
|AI | Human|
|--|--|
|Inference w/o CoT|System 1 reasoning|
|Inference with CoT| System 2 reasoning|

While this is a more and more widely used analogy (e.g., https://arxiv.org/ftp/arxiv/papers/2212/2212.05206.pdf),  I see that this might be debated.

I see two options:

* "It has been argued that in accuracy-based benchmarks without CoT, LLMs solve tasks fast and implicitly; whereas we need chain-of-thought based benchmarks to assess the system-2 reasoning skill of LLMs [REF1, REF"]."
* Drop this bullet point.

I'd be fine with both.

leaderboard-cot.md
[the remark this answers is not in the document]
pcuenca: You are right, it won't work. We can just update the description in the previous paragraph to refer to this entry.

leaderboard-cot.md
ggbetz: Not precisely, these are two different steps / prompts for generating the reasoning traces. Reasoning traces generation is separated from answering. We use our own [prompt chains](https://github.com/logikon-ai/cot-eval/tree/main/src/cot_eval/chains) to generate reasoning traces, and lm eval harness to assess model on tasks.

leaderboard-cot.md
ggbetz: > But the more powerful the "base" model is, the more difficult it would be for it to improve, wouldn't it?

Absolutely. So, when chosing model, you'd still want to consider Open LLM Leaderboard or YALL beside Open CoT leaderboard.

> In addition (and this is also mentioned in the post), some models have a natural tendency to follow CoT-like responses.

Not sure I fully understand this, but its probably not taken into account.

leaderboard-cot.md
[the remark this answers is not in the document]
clefourrier: I would maybe rephrase as
```
It is now common to distinguish fast and implicit (system-1) from slow and explicit (system-2) reasoning for human cognition. Though models do not "reason" like humans do, accuracy-based benchmarks without CoT are closest to the system-1 paradigm, whereas chain-of-thought based benchmarks are closer to system-2, an under-explored area of LLM performance. 
```
but even this is anthropomorphizing models a bit too much.

leaderboard-cot.md
ggbetz: I thought: Because this means that a model which has high absolute accuracy w/o CoT because it has seen the test examples during training doesn't automatically score high on the marginal CoT metric.

We're really not trying to substitute or replace existing metrics, but just adding something more (novel and informative).

leaderboard-cot.md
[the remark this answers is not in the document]
ggbetz: I'd say, let's drop that paragraph ... and discuss the philosophy of LLMs over a beer / glass of vine, if the opportunity arises. ;-)

leaderboard-cot.md
```diff
@@ -0,0 +1,179 @@
...
+Therefore, the assumption on which the editorial's argument depends is D) A school calendar made up of periods of study without long breaks would serve the interests of children more than a school calendar with long breaks.</td>
+  </tr>
+</table> 
+
+Each of our two prompting strategies—Classic and Reflect—is combined and run with multiple decoding parameters:
+- Greedy.
+- Beam search (n=2).
+- Sampling (T=.3).
+
+This gives us, in combination, six “CoT generation regimes.” When we evaluate a model, we let it generate, for every example in the test datasets, one chain-of-thought trace for each regime. We obtain, accordingly, six different numerical scores for accuracy with chain-of-thought (namely one for each regime). In the Open CoT Leaderboard, we report (for every model/task) the best marginal accuracy gain achieved under any regime.
```
ggbetz, line 148: I suggest to add another sentence that explains how the generated reasoning traces are used to score the models:

When we evaluate a model, we let it generate, for every example in the test datasets, one chain-of-thought trace for each regime. [START_INSERT]**The generated traces are then plugged into the prompt template shown above, which we use to score the models.</red>**[END_INSERT] We obtain, accordingly, six different numerical scores for accuracy with chain-of-thought (namely one for each regime). In the Open CoT Leaderboard, we report (for every model/task) the best marginal accuracy gain achieved under any regime.

leaderboard-cot.md
ggbetz: ... have suggested to add another sentence below.

leaderboard-cot.md
clefourrier: But it's not that beneficial - it will just allow to say "the performance delta is not there for this model", without pointing to a cause - I'll rephrase this paragraph.

leaderboard-cot.md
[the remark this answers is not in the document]
clefourrier: Works for me!

PR #2016, opened 2024-04-19
Update llama3.md

Hemanthkumar2112:
Encountering a NameError on 'tokenizer' within the terminator ID list element. Retrieve IDs from the pipeline tokenizer attribute instead.


pcuenca (approved):
Thank you!

PR #2024, opened 2024-04-22
Update Matryoshka blogpost to use new Sentence Transformers v2.7.0 functionality

tomaarsen:
Hello!

## Pull Request overview
* Use new `truncate_dim` parameter in inference snippet.

## Details
This should help share the news for `truncate_dim`, which simplifies the usage of Matryoshka models, especially in third party applications where you can't manually post-process embeddings before they're used.

- Tom Aarsen

### DIFF
--- a/matryoshka.md
+++ b/matryoshka.md
@@ -127,7 +127,7 @@ Keep in mind that although processing smaller embeddings for downstream tasks (r
 
 ### In Sentence Transformers
 
-In Sentence Transformers, you can load a Matryoshka Embedding model like normal, and run inference with it using [`SentenceTransformers.encode`](https://sbert.net/docs/package_reference/SentenceTransformer.html#sentence_transformers.SentenceTransformer.encode). After getting the embeddings, we can truncate them to our desired size, and we can normalize them if we want.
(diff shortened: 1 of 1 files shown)


matryoshka.md
```diff
@@ -135,17 +135,16 @@ Let's try to use a model that I trained using [`matryoshka_nli.py`](https://gith
 from sentence_transformers import SentenceTransformer
 from sentence_transformers.util import cos_sim
 
-model = SentenceTransformer("tomaarsen/mpnet-base-nli-matryoshka")
-
 matryoshka_dim = 64
+model = SentenceTransformer("tomaarsen/mpnet-base-nli-matryoshka", truncate_dim=matryoshka_dim)
```
pcuenca, line 139: Maybe add a comment in line 130 about `truncate_dim` doing the truncation for us? It's ok, imo, that another comment about doing it manually appears later in L170.

matryoshka.md
```diff
@@ -168,6 +167,8 @@ References:
 
 <details><summary><b>Click here to see how to use the Nomic v1.5 Matryoshka Model</b></summary>
 
+Note: Nomic specifically requires an `F.layer_norm` before the embedding truncation. As a result, the following snippet uses manual truncation to the desired dimension. For all other models, you can use the `truncate_dim` option in the constructor.
```
pcuenca, line 170:
```suggestion
Note: Nomic specifically requires an `F.layer_norm` before the embedding truncation. As a result, the following snippet uses manual truncation to the desired dimension. For all other models, you can use the `truncate_dim` option in the constructor, as shown in the previous example.
```

pcuenca (approved):

matryoshka.md
tomaarsen: I mentioned it in the running text above it, instead

PR #2030, opened 2024-04-24
Update moe.md

ThomasLWang:
Fix a translation error. The translation here is completely opposite in meaning. The original text is: Sparse models fine-tuned quality improves with higher learning rates and smaller batch sizes.


xianbaoqian:
Ah thanks for catching that!

xianbaoqian (approved):

PR #2037, opened 2024-04-29
Fix unnecessary memory usage when cpu offloading

standardAI:
https://huggingface.co/docs/diffusers/main/en/optimization/memory#cpu-offloading
@sayakpaul @stevhliu


stevhliu (approved):
Thanks for the fix!

standardAI:
Thanks for merging!

PR #2045, opened 2024-05-07
Blog post on KV cache quantization

zucchini-nlp:
Add the blog post for kv cache quantization. Mostly the finalized version, but I could not find a way to group figures in a row to save space. Let me know if you have any idea on how it's done :) 

cc @younesbelkada @SunMarc @gante


kv_cache_quantization.md
```diff
@@ -0,0 +1,147 @@
+---
+title: "Unlocking Longer Generation with Key-Value Cache Quantization" 
+thumbnail: /blog/assets/kv_cache_quantization/thumbnail.gif
+authors:
+- user: RaushanTurganbay
+- user: your_coauthor
```
younesbelkada, line 6:
```suggestion
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,147 @@
...
+---
+
+
+# Introduction
+
+I'm excited to share with you a new feature that's going to take your language models to the next level: kv cache quantization.
```
younesbelkada, line 12:
```suggestion
At Hugging Face, we are excited to share with you a new feature that's going to take your language models to the next level: *kv cache quantization*. 
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,147 @@
...
+---
+
+
+# Introduction
+
+I'm excited to share with you a new feature that's going to take your language models to the next level: kv cache quantization. 
+
+Have you ever tried generating a lengthy piece of text with your language model, only to hit a wall because of pesky memory limitations? As language models continue to grow in size and capabilities, supporting longer generations can start to really eat up memory. It's a common frustration, especially when you're dealing with limited resources. That's where kv cache quantization swoops in to save the day.
+
+So, what exactly is kv cache quantization? If you're not familiar with the term, don't sweat it! Let's break it down into two pieces: kv cache and quantization.
```
younesbelkada, line 16:
```suggestion
So, what exactly is kv cache quantization? If you're not familiar with the term, don't sweat it! Let's break it down into two pieces: *kv cache* and *quantization*. 
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,147 @@
...
+
+Another key part of the KIVI method is retaining a residual cache to store keys and values in their original precision, and when the residucal cache reaches its maximum capacity the keys and values are quantized and the cache content is discarded. This small trick allows to rpeserve accuracy since some part of the most recent keys and values are always stored in their original precision. Also, it allows to run seamlessly the per-channel key quantization which requires the number of tokens to be divisible by the group size. The main consideration is the memory-efficiency trade-off when setting the residual cache length. While residual cache stores keys and values in their original precision, that may result in overall memory usage increase. So, the general idea of KIVI is to maintain two separate caches for quantized and residual parts of key and value matrices.
```
younesbelkada, line 41:
```suggestion
Another key part of the KIVI method is retaining a residual cache to store keys and values in their original precision, and when the residual cache reaches its maximum capacity the keys and values are quantized and the cache content is discarded. This small trick allows to preserve accuracy since some part of the most recent keys and values are always stored in their original precision. Also, it allows to run seamlessly the per-channel key quantization which requires the number of tokens to be divisible by the group size. The main consideration is the memory-efficiency trade-off when setting the residual cache length. While residual cache stores keys and values in their original precision, that may result in overall memory usage increase. So, the general idea of KIVI is to maintain two separate caches for quantized and residual parts of key and value matrices.
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,147 @@
...
+`X_Q = (X / S) - Z`
+
+where, 
+- X_Q is the quantized tensor
+- S is the scale calculated as `(maxX - minX) / (max_val_for_precision - min_val_for_precision)`
+- Z is zeropoin calculated as `round(-minX / S)`
+
+
+To integrate kv quantization seamlessly in Transformers, we rely on [quanto](https://github.com/huggingface/quanto) library. Quanto is a toolkit for easy implementation and combination of different quantization tecniques. Currently available quantization precisions for the key-value cache are `int2` and `int4`.
+
```
younesbelkada, line 54: You can also add a link to the recent quanto blogpost: https://huggingface.co/blog/quanto-introduction

kv_cache_quantization.md
```diff
@@ -0,0 +1,147 @@
...
+1. **Memory vs Speed trade-off**: By quantizing the kv cache into lower precision formats, memory usage is significantly reduced, allowing for longer text generations without encountering memory constraints. But users have to decide on whether giving up a tiny bit of generation speed suits their use-case.
+
+2. **Maintained Accuracy**: Despite the reduction in precision, kv cache quantization in `int4` preserves model accuracy to a satisfactory extent, ensuring that generated text remains contextually relevant and coherent.
+
+3. **Flexibility**: Users have the flexibility to choose between different precision formats based on their specific requirements, allowing for customization to suit varying use cases and priorities.
+
+4. **Potential for Further Optimization**: While kv cache quantization provides significant benefits on its own, it can also be combined with other optimization techniques, such as weight quantization, to further enhance memory efficiency and computational speed.
+  
+
+# Acknowledgment
```
younesbelkada, line 138: This section is empty

younesbelkada (commented):
Great work @zucchini-nlp ! Looking forward to 🚢 this in the HF ecosystem ! 🚀 Left few comments

kv_cache_quantization.md
zucchini-nlp: i will add later

kv_cache_quantization.md
```diff
@@ -0,0 +1,146 @@
...
+
+The implementation of kv cache in Transformers is partially based on KIVI, specifically we adopted the same strategy of retaining `k` tokens in the residual cache for quality maintenance. In contrast to the KIVI, Transformers quantizes both keys and value per-channel by grouping on the last dimension. We found that this type of per-channel quantization does not hurt accuracy of the generations. So given a key or value of shape `batch size, num of heads, num of tokens, head dim` we group it to `num of groups, group size` and perform affine quantization as follows:
+
+`X_Q = (X / S) - Z`
+
+where, 
+- X_Q is the quantized tensor
+- S is the scale calculated as `(maxX - minX) / (max_val_for_precision - min_val_for_precision)`
+- Z is zeropoin calculated as `round(-minX / S)`
```
SunMarc, line 49:
```suggestion
where, 
- X_Q is the quantized tensor
- S is the scale calculated as `(maxX - minX) / (max_val_for_precision - min_val_for_precision)`
- Z is zeropoint calculated as `round(-minX / S)`
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,146 @@
...
+
+The implementation of kv cache in Transformers is partially based on KIVI, specifically we adopted the same strategy of retaining `k` tokens in the residual cache for quality maintenance. In contrast to the KIVI, Transformers quantizes both keys and value per-channel by grouping on the last dimension. We found that this type of per-channel quantization does not hurt accuracy of the generations. So given a key or value of shape `batch size, num of heads, num of tokens, head dim` we group it to `num of groups, group size` and perform affine quantization as follows:
+
+`X_Q = (X / S) - Z`
```
SunMarc, line 44:
```suggestion
`X_Q = round(X / S) - Z`
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,146 @@
...
+| TREC                  | 63.0        | 67.5         | 63.0                | 63.0    | 55.0    |
+| SAMSum                | 41.12       | 42.18        | 41.12               | 41.3    | 14.04   |
+| TriviaQA              | NA          | NA           | 84.28               | 84.76   | 63.64   |
+| HotPotQA              | NA          | NA           | 30.08               | 30.04   | 17.3    |
+| Passage_retrieval_en  | NA          | NA           | 8.5                 | 9.5     | 4.82    |
+
+
+Now, let's talk about the trade-off between memory savings and speed. When we quantize the kv cache in models, we're making them less memory hungry, but sometimes that comes at a tiny cost to generation speed. While quantiziing the cache to `int4` can offer roughly an x2 memory saving, the generation speed starts to decrease with higher batch sizes. One has to decide whether using quantized kv cache and potentially sacrificing a bit of speed is worth the trade-off for the significant gains in memory efficiency. It's all about finding the approach that best suits your specific use case and priorities. 
+
```
SunMarc, line 76: Why do we only get a x2, I was thinking we might get around a x4 or x3 ? Is it because of the quantization parameters or the residuals ? Also, @dacorvo is working on better kernels for int4 (awq kernels which are the one of the fastest kernels). We might get better results soon !

```suggestion
Now, let's talk about the trade-off between memory savings and speed. When we quantize the kv cache in models, we're making them less memory hungry, but sometimes that comes at a tiny cost to generation speed. While quantizing the cache to `int4` can offer roughly an x2 memory saving, the generation speed starts to decrease with higher batch sizes. One has to decide whether using quantized kv cache and potentially sacrificing a bit of speed is worth the trade-off for the significant gains in memory efficiency. It's all about finding the approach that best suits your specific use case and priorities. 

```

kv_cache_quantization.md
```diff
@@ -0,0 +1,146 @@
...
+To use kv cache quantization in 🤗 Tranformers we have to install external dependencies first by running `pip install quanto`. To activate quantization on kv cache, we have to pass in `cache_implementation="quantized"` and indicate quantization parameters in a cache config in dictionary format. And that's all we need to start using kv cache quantization. Additionally, since quanto is device agnostic, you can quantize and run your model regardless if you are on CPU/GPU/ MPS (Apple Silicon). 
+
+
+```python
+>>> import torch
+>>> from transformers import AutoTokenizer, AutoModelForCausalLM
+
+>>> tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-2-7b-chat-hf")
+>>> model = AutoModelForCausalLM.from_pretrained("meta-llama/Llama-2-7b-chat-hf", torch_dtype=torch.float16).to("cuda:0")
```
SunMarc, line 110: You can load the model faster by doing: 
```suggestion
>>> model = AutoModelForCausalLM.from_pretrained("meta-llama/Llama-2-7b-chat-hf", torch_dtype=torch.float16, device_map="cuda:0")
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,146 @@
...
+>>> out = model.generate(**inputs, do_sample=False, max_new_tokens=20)
+>>> print(tokenizer.batch_decode(out, skip_special_tokens=True)[0])
+I like rock music because it's loud and energetic. I like to listen to it when I'm feeling
+```
+
+
+
+# Conclusion
+
+There are many more different methods to reduce memory usage by key-value cache, including [MultiQueryAttention](https://arxiv.org/abs/1911.02150), [GroupedQueryAttention](https://arxiv.org/abs/2305.13245) or recent [kv cache retrieval](https://arxiv.org/abs/2403.09054) methods. While some of these methods are bound to the model architecture choices, others can be applied post-training. Quantization is one of such post-training optimization tecniques and we can draw the following conclusion from our short blogpost:
```
SunMarc, line 126:
```suggestion
There are many more different methods to reduce memory usage by key-value cache, including [MultiQueryAttention](https://arxiv.org/abs/1911.02150), [GroupedQueryAttention](https://arxiv.org/abs/2305.13245) or recent [kv cache retrieval](https://arxiv.org/abs/2403.09054) methods. While some of these methods are bound to the model architecture choices, others can be applied post-training. Quantization is one of such post-training optimization techniques and we can draw the following conclusion from our short blogpost:
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+# How to use quantized kv cache in 🤗 Transformers?
+
+To use kv cache quantization in 🤗 Tranformers we have to install external dependencies first by running `pip install quanto`. To activate quantization on kv cache, we have to pass in `cache_implementation="quantized"` and indicate quantization parameters in a cache config in dictionary format. And that's all we need to start using kv cache quantization. Additionally, since quanto is device agnostic, you can quantize and run your model regardless if you are on CPU/GPU/ MPS (Apple Silicon). 
+
```
SunMarc, line 103: It would be great to include a colab notebook, so that the reader can test the code without having the think about how to install the libraries !

SunMarc (approved):
Very nice blogpost @zucchini-nlp ! 🔥I left a few comments

kv_cache_quantization.md
zucchini-nlp: Yes, prob should be more than x2, I will check with lower residual length just to make sure it is working as expected.

kv_cache_quantization.md
```diff
@@ -0,0 +1,146 @@
...
+<figure class="image text-center m-0">
+  <img class="center" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/kv_cache_quantization/perplexity.png" alt="Log Perplexity Comparison"/>
+</figure>
+
+
+Same holds true when calculating performance on the [LongBench](https://huggingface.co/datasets/THUDM/LongBench) benchmark comparing it to results from the KIVI paper. `Int4` precision is comparable and even outperforms slighly the `fp16` in all of the datasets.
```
gante, line 64: I would add a note somewhere to say that higher is better in the benchmarks in the table below

kv_cache_quantization.md
```diff
@@ -0,0 +1,146 @@
...
+| TriviaQA              | NA          | NA           | 84.28               | 84.76   | 63.64   |
+| HotPotQA              | NA          | NA           | 30.08               | 30.04   | 17.3    |
+| Passage_retrieval_en  | NA          | NA           | 8.5                 | 9.5     | 4.82    |
+
+
+Now, let's talk about the trade-off between memory savings and speed. When we quantize the kv cache in models, we're making them less memory hungry, but sometimes that comes at a tiny cost to generation speed. While quantizing the cache to `int4` can offer roughly an x2 memory saving, the generation speed starts to decrease with higher batch sizes. One has to decide whether using quantized kv cache and potentially sacrificing a bit of speed is worth the trade-off for the significant gains in memory efficiency. It's all about finding the approach that best suits your specific use case and priorities. 
+
+Below are the performance metrics for kv cache in original precision and quantized format. Script to obtain the following figures is available [here](https://gist.github.com/zucchini-nlp/56ce57276d7b1ee666e957912d8d36ca).
```
gante, line 77: reproducibility 💛 

I would add a comment with the `transformers` commit hash used for the blog post (in the gist)!

kv_cache_quantization.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+# How to use quantized kv cache in 🤗 Transformers?
+
+To use kv cache quantization in 🤗 Tranformers we have to install external dependencies first by running `pip install quanto`. To activate quantization on kv cache, we have to pass in `cache_implementation="quantized"` and indicate quantization parameters in a cache config in dictionary format. And that's all we need to start using kv cache quantization. Additionally, since quanto is device agnostic, you can quantize and run your model regardless if you are on CPU/GPU/ MPS (Apple Silicon).
```
gante, line 102:
```suggestion
To use kv cache quantization in 🤗 Tranformers we have to install external dependencies first by running `pip install quanto`. To activate quantization on kv cache, we have to pass in `cache_implementation="quantized"` and indicate quantization parameters in a cache config in dictionary format. And that's all we need to start using kv cache quantization. Additionally, since quanto is device agnostic, you can quantize and run your model regardless if you are on CPU/GPU/MPS (Apple Silicon). 
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+Now, let's talk about the trade-off between memory savings and speed. When we quantize the kv cache in models, we're making them less memory hungry, but sometimes that comes at a tiny cost to generation speed. While quantizing the cache to `int4` can offer roughly an x2 memory saving, the generation speed starts to decrease with higher batch sizes. One has to decide whether using quantized kv cache and potentially sacrificing a bit of speed is worth the trade-off for the significant gains in memory efficiency. It's all about finding the approach that best suits your specific use case and priorities. 
+
+Below are the performance metrics for kv cache in original precision and quantized format. Script to obtain the following figures is available [here](https://gist.github.com/zucchini-nlp/56ce57276d7b1ee666e957912d8d36ca).
+
+
+<div style="display: flex;">
+  <figure class="image m-0" style="width: 20%;">
+    <img class="center" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/kv_cache_quantization/gpu_mem_max_new_tokens.png" alt="GPU memory consumption as max new tokens increase"/>
```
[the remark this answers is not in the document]
zucchini-nlp: Hmm I will look into it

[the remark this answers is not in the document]
zucchini-nlp: I found what is the reason, the plot shows metrics in a cumulative way, so I will update them. I was accidenteally subtracting the prev_max every time we got new max_memory. Anyway, the general trend of x2.5 memroy savings will remain same

kv_cache_quantization.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+<figure class="image text-center m-0">
+  <img class="center" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/kv_cache_quantization/kv-cache-optimization.png" alt="kv cache visual"/>
```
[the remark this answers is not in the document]
gante: No need to bother then :) Most users won't notice the indexing errors 😉

kv_cache_quantization.md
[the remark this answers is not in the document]
gante: Perfect, thank you for the update!

kv_cache_quantization.md
```diff
@@ -0,0 +1,146 @@
+---
+title: "Unlocking Longer Generation with Key-Value Cache Quantization" 
+thumbnail: /blog/assets/kv_cache_quantization/thumbnail.gif
```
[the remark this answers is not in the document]
zucchini-nlp: done!

kv_cache_quantization.md
```diff
@@ -0,0 +1,152 @@
...
+  </figure>
+</div>
+
+
+<figure class="image text-center m-0">
+  <img class="center" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/kv_cache_quantization/latency.png" alt="Latency as batch size increases"/>
+</figure>
+
+
+Wondering what happens when we throw weight quantization into the mix? Sure, combining these techniques can further slim down your model's memory footprint, but there's a catch – it might slow things down even more. In fact, our experiments show that weigt quantization together with kv cache quantization can lead to a threefold decrease in speed. But we're constantly tinkering away to find ways to make this combo work seamlessly. And while we don't currently have optimized kernels in the `quanto` library, we're open to community contributions that could help improve computational efficiency. Our goal is to ensure your model runs smoothly while maintaining high latency and accuracy.
```
dacorvo, line 95: ### 
```suggestion
Wondering what happens when we throw weight quantization into the mix? Sure, combining these techniques can further slim down your model's memory footprint, but there's a catch – it might slow things down even more. In fact, our experiments show that weight quantization together with kv cache quantization can lead to a threefold decrease in speed. But we're constantly tinkering away to find ways to make this combo work seamlessly. And while we don't currently have optimized kernels in the `quanto` library, we're open to community contributions that could help improve computational efficiency. Our goal is to ensure your model runs smoothly while maintaining high latency and accuracy.
```

dacorvo (commented):
Nice ! Wasn't aware of that work. Cool to see that quanto is versatile enough to allow for lower bitwidth quantization of activations (I only tested it with 8-bit activations so far). A very interesting derivative work for this would be to take advantage of integer matmul when weights are also 4-bit quantized. This would require changes in quanto, but not so much.
Finally, I found a small nit.

pcuenca (changes requested):
This is a new post, right? In that case, we need to add an entry to `_blog.yml` before this can be merged.

For example: https://github.com/huggingface/blog/blob/main/_blog.yml#L3999-L4007

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
+---
+title: "Unlocking Longer Generation with Key-Value Cache Quantization" 
+thumbnail: /blog/assets/kv_cache_quantization/thumbnail.png
+authors:
+- user: RaushanTurganbay
+---
+
+
+# Introduction
```
pcuenca, line 9:
```suggestion
# Unlocking Longer Generation with Key-Value Cache Quantization
```

The blog title is not rendered automatically :)

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
...ediction is dependant on the previous context. That is where key-value cache is used to optimize the sequential generation process by storing previously calculated values in a cache. Key-Value cache acts as a memory bank for autoregressive generative models, where the model stores key-value pairs derived from self-attention layers for previously processed tokens. In the transformer architecture self-attention layers calculate attention scores by multiplying queries with keys, producing weighted sums of value vectors as outputs. By storing this information, the model can avoid redundant computations and instead retrieve keys and values of previous tokens from the cache. Basically, key value cache enables the model to use information from earlier parts of the input text during generation without performing extra computations. For a visual explanation of this concept, take a look at how key-value cache functions in the image below. When calculating the attentions scores for `K+1`th token we do not need to recompute all of the previous keys and values, but rather take it from cache and concatenate to the current vector. This usually results in faster and more efficient text generation.
```
pcuenca, line 17:
```suggestion
Key-value cache, or kv cache, is needed to optimize the generation in autoregressive models, where the model predicts text token by token. This process can be slow since the model can generate only one token at a time, and each new prediction is dependent on the previous context. That is where key-value cache is used to optimize the sequential generation process by storing previously calculated values in a cache. Key-Value cache acts as a memory bank for autoregressive generative models, where the model stores key-value pairs derived from self-attention layers for previously processed tokens. In the transformer architecture, self-attention layers calculate attention scores by multiplying queries with keys, producing weighted sums of value vectors as outputs. By storing this information, the model can avoid redundant computations and instead retrieve keys and values of previous tokens from the cache. Basically, key value cache enables the model to use information from earlier parts of the input text during generation without performing extra computations. For a visual explanation of this concept, take a look at how key-value cache functions in the image below. When calculating the attentions scores for the `K+1`th token we do not need to recompute all of the previous keys and values, but rather take it from cache and concatenate to the current vector. This usually results in faster and more efficient text generation.
```

Fixed a couple of typos above (dependant -> dependent, for example), not sure why the preview doesn't show them.

I also found this description a bit hard to follow. Just as a suggestion (that you can of course ignore), I tried to explain it more intuitively first and simplify the technical description, but not sure if I dumbed it down a lot:

> Key-value cache, or kv cache, is needed to optimize the generation in autoregressive models, where the model predicts text token by token. This process can be slow since the model can generate only one token at a time, and each new prediction is dependent on the previous context. That means, to predict token number 1000 in the generation, you need information from the previous 999 tokens, which comes in the form of some matrix multiplications across the representations of those tokens. But to predict token number 1001, you also need the same information from the first 999 tokens, plus additional information from token number 1000. Key-value caching stores the components of previous calculations that can be reused in subsequent tokens, so they don't need to be computed again.
> 
> More concretely, the information stored for models based on the transformer architecture consists of two essential components used in all of the self-attention layers: the input _keys_, and the attention scores, or _values_, calculated after multiplying queries with keys. For a visual explanation of this concept, take a look at how key-value cache works in the image below. When calculating the attentions scores for the `K+1`th token we do not need to recompute all of the previous keys and values, but rather take it from cache and concatenate to the current vector. This usually results in faster and more efficient text generation.

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+<figure class="image text-center m-0">
+  <img class="center" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/kv_cache_quantization/kv-cache-optimization.png" alt="kv cache visual"/>
+</figure>
+
+
+Moving on to the second term, quantization is just a fancy word for reducing the precision of numerical values in a model which helps save memory. During quantization each numerical value is rounded or truncated to fit within the reduced precision format, which may result in a loss of information. However, careful selection of quantization parameters and techniques can minimize this loss while still achieving satisfactory performance. There are different quantization methods existing and if you're curious to learn more, be sure to check out our [previous blog post](https://huggingface.co/blog/4bit-transformers-bitsandbytes) for a deeper dive into the world of quantization.
```
pcuenca, line 25:
```suggestion
Moving on to the second term, quantization is just a fancy word for reducing the precision of numerical values to save memory. During quantization, each numerical value is rounded or truncated to fit within the reduced precision format, which may result in a loss of information. However, careful selection of quantization parameters and techniques can minimize this loss while still achieving satisfactory performance. There are different quantization methods, so if you're curious to learn more be sure to check out our [previous blog post](https://huggingface.co/blog/4bit-transformers-bitsandbytes) for a deeper dive into the world of quantization. 
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+`2 * 2 * 32 * 32 * 128 * 10000 ≈ 5GB`
+
+of memory only to store the previous key-value cache, which is almost one third of the memory required to store model parameters in half-precision.
+
+Therefore by compressing kv cache into a more compact form we can save up a lot of memory and run long context generation on consumer GPUs. In our experiments, we were able to significantly reduce the memory footprint without sacrificing too much quality by quantizing the kv cache into lower precision formats. With this new quantization feature, we can now support longer generations without running out of memory, which means you can expand your model's context length without worrying about hitting a memory constraint.
```
pcuenca, line 33:
```suggestion
Therefore, by compressing kv cache into a more compact form we can save up a lot of memory and run longer context generation on consumer GPUs. In our experiments, we were able to significantly reduce the memory footprint without sacrificing too much quality by quantizing the kv cache into lower precision formats. With this new quantization feature, we can now support longer generations without running out of memory, which means you can expand your model's context length without worrying about hitting a memory constraint.
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+# Implementation Details
```
pcuenca, line 36:
```suggestion
## Implementation Details
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+Key-value cache quantization in Transformers was largely inspired by the [KIVI: A Tuning-Free Asymmetric 2bit Quantization for kv Cache](https://arxiv.org/abs/2402.02750) paper. The paper introduced a 2bit asymmetrical quantization for large language models without quality degradation. KIVI quantizes the key cache per-channel and the value cache per-token, because keys have higher magnitudes of outliers in some channels while value cache does not such a pattern. That is why the relative error between quantized and original precision is much smaller when keys are quantized per-channel and the values per-token.
```
pcuenca, line 38:
```suggestion
Key-value cache quantization in Transformers was largely inspired by the [KIVI: A Tuning-Free Asymmetric 2bit Quantization for kv Cache](https://arxiv.org/abs/2402.02750) paper. The paper introduced a 2bit asymmetrical quantization for large language models without quality degradation. KIVI quantizes the key cache per-channel and the value cache per-token, because keys have higher magnitudes of outliers in some channels while values don't show such a pattern. Therefore, the relative error between quantized and original precision is much smaller when keys are quantized per-channel and the values per-token.
```

zucchini-nlp:
@pcuenca thanks, added it to the yaml file

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+The implementation of kv cache in Transformers is partially based on KIVI, specifically we adopted the same strategy of retaining `k` tokens in the residual cache for quality maintenance. In contrast to the KIVI, Transformers quantizes both keys and value per-token. We found that this type of per-token quantization does not hurt accuracy of the generations for `int4` precision. So given a key or value of shape `batch size, num of heads, num of tokens, head dim` we group it to `num of groups, group size` and perform affine quantization as follows:
```
pcuenca, line 42: Nice. It's interesting that we explain the KIVI process in detail and then don't use some of the learnings. Just as an idea, perhaps we could simply introduce the paper and the general idea, then explain the transformers implementation, then discuss the details of how it differs from KIVI. But probably not worth it at this stage.

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+`X_Q = round(X / S) - Z`
+
+where, 
+- X_Q is the quantized tensor
+- S is the scale calculated as `(maxX - minX) / (max_val_for_precision - min_val_for_precision)`
+- Z is zeropoint calculated as `round(-minX / S)`
```
pcuenca, line 49:
```suggestion
- Z is the zeropoint, calculated as `round(-minX / S)`
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+Currently the kv quantization works on [quanto](https://github.com/huggingface/quanto) backend and supports precisions `int2` and `int4`. Although we don't currently support other quantization backends, we are open to community contributions that could help integrate them. Specifically, quantization methods that do not need calibration data and can dynamically calculate lower-bit tensors on-the-fly can be easily integrated.
```
pcuenca, line 51:
```suggestion
Currently, the kv quantization works on [quanto](https://github.com/huggingface/quanto) backend and supports `int2` and `int4` precisions. Although we don't currently support other quantization backends, we are open to community contributions that could help integrate them. Specifically, quantization methods that do not need calibration data and can dynamically calculate lower-bit tensors on-the-fly can be easily integrated.
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+# Comparing performance of fp16 and quantized cache
```
pcuenca, line 54:
```suggestion
## Comparing performance of fp16 and quantized cache
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+We know visuals speak louder than words, so we've prepared up some comparison plots to give you a snapshot of how quantization stacks up against FP16 precision. These plots will show you at a glance how the model's generation hold up in terms of quality when we tweak the precision settings for kv cache. We calculated the perplexity of [Llama2-7b-chat](https://huggingface.co/meta-llama/Llama-2-7b-chat-hf) model on the [`PG-19`](https://huggingface.co/datasets/emozilla/pg19-test) dataset. We can see that `int4` cache performs almost same as the original `fp16` precision, while the quality degrades when using `int2`. The script to reproduce the resuts is available [here](https://gist.github.com/zucchini-nlp/a7b19ec32f8c402761d48f3736eac808).
```
pcuenca, line 56:
```suggestion
We know visuals speak louder than words, so we've prepared some comparison plots to give you a snapshot of how quantization stacks up against FP16 precision. These plots show you at a glance how the model's generation holds up in terms of quality when we tweak the precision settings for kv cache. We calculated the perplexity of [Llama2-7b-chat](https://huggingface.co/meta-llama/Llama-2-7b-chat-hf) model on the [`PG-19`](https://huggingface.co/datasets/emozilla/pg19-test) dataset. We can see that `int4` cache performs almost the same as the original `fp16` precision, while the quality degrades when using `int2`. The script to reproduce the results is available [here](https://gist.github.com/zucchini-nlp/a7b19ec32f8c402761d48f3736eac808).
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+<figure class="image text-center m-0">
+  <img class="center" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/kv_cache_quantization/perplexity.png" alt="Log Perplexity Comparison"/>
+</figure>
+
+
+Same holds true when calculating performance on the [LongBench](https://huggingface.co/datasets/THUDM/LongBench) benchmark comparing it to results from the KIVI paper. `Int4` precision is comparable and even outperforms slighly the `fp16` in all of the datasets in the table below (higher is better).
```
pcuenca, line 63:
```suggestion
The same conclusion holds when calculating performance on the [LongBench](https://huggingface.co/datasets/THUDM/LongBench) benchmark comparing it to results from the KIVI paper. `Int4` precision is comparable and even outperforms slightly the `fp16` in all of the datasets in the table below (higher is better).
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+Although it is worth noting that processing input prompt tokens (aka pre-fill stage), unlike subsequent generated tokens, still require computing the entire key-value matrices in one go for the whole input that may be another memory bottleneck for long contexts. Respectively, the latency associated with generating the first token tends to be higher compared to subsequent tokens. There are other different strategies to decrease memory burden for the pre-fill stage by optimizing the attention computation stage, such like [Local Windowed Attention](https://arxiv.org/abs/2004.05150) or [Flash-Attention](https://arxiv.org/abs/2307.08691). If you are out of memory for the pre-fill stage, you can use `FlashAttention` in 🤗 Transformers along with the kv cache quantization to decrease memory usage even more for long input prompts. See [docs](https://huggingface.co/docs/transformers/main/en/perf_infer_gpu_one#flashattention-2) for more information on that.
```
pcuenca, line 95:
```suggestion
It's also worth noting that initial processing of the input prompt (aka pre-fill stage) still requires computing the entire key-value matrices in one go for the whole input, which may be another memory bottleneck for long contexts. This is the reason why the latency associated with generating the first token tends to be higher compared to subsequent tokens. There are other different strategies to decrease the memory burden of the pre-fill stage by optimizing the attention computation stage, such like [Local Windowed Attention](https://arxiv.org/abs/2004.05150) or [Flash-Attention](https://arxiv.org/abs/2307.08691). If you are out of memory for the pre-fill stage, you can use `FlashAttention` in 🤗 Transformers along with the kv cache quantization to decrease memory usage even more for long input prompts. See [the docs](https://huggingface.co/docs/transformers/main/en/perf_infer_gpu_one#flashattention-2) for more information on that.
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+If you are interested how many tokens we can fit in the context if we were to push the memory usage to its limits, quantized kv cache can support up to 128k tokens with Flash Attention enabled in an 80GB A100. For the cache in half precision, the maximum capacity is 40k tokens.
```
pcuenca, line 97:
```suggestion
If you are interested to know how many tokens we can fit in the context when we push memory usage to its limits, quantized kv cache can support up to 128k tokens with Flash Attention enabled in an 80GB A100. For the cache in half precision, the maximum capacity is 40k tokens.
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+# How to use quantized kv cache in 🤗 Transformers?
```
pcuenca, line 99:
```suggestion
## How to use quantized kv cache in 🤗 Transformers?
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+I like rock music because it's loud and energetic. It's a great way to express myself and rel
+
+>>> out = model.generate(**inputs, do_sample=False, max_new_tokens=20)
+>>> print(tokenizer.batch_decode(out, skip_special_tokens=True)[0])
+I like rock music because it's loud and energetic. I like to listen to it when I'm feeling
+```
+
+
+
+# Conclusion
```
pcuenca, line 124:
```suggestion
## Conclusion
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+1. **Memory vs Speed trade-off**: By quantizing the kv cache into lower precision formats, memory usage is significantly reduced, allowing for longer text generations without encountering memory constraints. But users have to decide on whether giving up a tiny bit of generation speed suits their use-case.
+
+2. **Maintained Accuracy**: Despite the reduction in precision, kv cache quantization in `int4` preserves model accuracy to a satisfactory extent, ensuring that generated text remains contextually relevant and coherent.
+
+3. **Flexibility**: Users have the flexibility to choose between different precision formats based on their specific requirements, allowing for customization to suit varying use cases and priorities.
+
+4. **Potential for Further Optimization**: While kv cache quantization provides significant benefits on its own, it can also be combined with other optimization techniques, such as weight quantization, to further enhance memory efficiency and computational speed.
+  
+
+# Acknowledgment
```
pcuenca, line 137:
```suggestion
## Acknowledgment
```

kv_cache_quantization.md
```diff
@@ -0,0 +1,150 @@
...
+
+Special thanks to [Younes](https://huggingface.co/ybelkada) and [Marc](https://huggingface.co/marcsun13) for their assistance and advice on quantization techniques. Their expertise greatly contributed to the development of this feature.
+
+Additionally, I would like to thank [Joao](https://huggingface.co/joaogante) for his invaluable support.
+
+
+# Further Resources
```
pcuenca, line 144:
```suggestion
## Additional Resources
```

pcuenca (commented):
Very cool, great work! 🔥

_blog.yml
```diff
@@ -4005,3 +4005,13 @@
     - multimodal
     - LLM
     - vision
+
+- local: kv-cache-quantization
```
pcuenca, line 4009: The file name uses underscores. I'd rename the file instead of modifying this entry.

pcuenca (approved):
Approving to unblock, please make sure the name in the yaml matches the filename :)

kv_cache_quantization.md
zucchini-nlp: Yes, and the `transformers` version is mostly same as KIVI. Do you think we can leave the almost same explanation, but do not mention the KIVI paper more than the first time, in the beginning? 

So, it will be something like `Another key part our method is retaining a residual cache ...`

pcuenca: Sounds good!

gante (approved):
🔥 (ready to merge on my end)

zucchini-nlp:
I don't have merge rights on this repo, who can merge?

gante:
@zucchini-nlp merged

PR #2063, opened 2024-05-14
Update stable_diffusion.md

tolgacangoz:
Shouldn't the compression ratio be 48?
@stevhliu @sayakpaul


stable_diffusion.md
```diff
@@ -269,7 +269,7 @@ Inspired by [Imagen](https://imagen.research.google/), Stable Diffusion does **n
 
 **Why is latent diffusion fast and efficient?**
 
-Since latent diffusion operates on a low dimensional space, it greatly reduces the memory and compute requirements compared to pixel-space diffusion models. For example, the autoencoder used in Stable Diffusion has a reduction factor of 8. This means that an image of shape `(3, 512, 512)` becomes `(3, 64, 64)` in latent space, which requires `8 × 8 = 64` times less memory.
+Since latent diffusion operates on a low dimensional space, it greatly reduces the memory and compute requirements compared to pixel-space diffusion models. For example, the autoencoder used in Stable Diffusion has a reduction factor of 8. This means that an image of shape `(3, 512, 512)` becomes `(4, 64, 64)` in latent space, which requires `8 × 8 × 3 ÷ 4 = 48` times less memory.
```
sayakpaul, line 272: We only count the resolution. So, I think it's fine.

standardAI: I want to understand well. It says "becomes `(3, 64, 64)` in latent space"; but is this latent space? Isn't latent space 4, 64, 64?

standardAI: I want/need to understand well. It says "becomes `(3, 64, 64)` in latent space"; but is this latent space? Isn't latent space `(4, 64, 64)`? AFAIU, 64 is the spatial compression factor, but 48 is the overall compression factor. At the end of the day, isn't the overall one necessary to be reported? It is argued that much times less memory.

sayakpaul: Let’s go with spatial compression. That should be fine.

standardAI: Saying 48 times compression seemed reasonable to me here: https://github.com/fastai/diffusion-nbs/blob/master/Stable%20Diffusion%20Deep%20Dive.ipynb

tolgacangoz:
@sayakpaul Gently pinging here.

tolgacangoz:
So, what to do here?

tolgacangoz:
So, what to do here?
@sayakpaul Gently pinging here.

pcuenca (approved):
Let's go with the latest wording proposed, which no longer mentions a 64-fold memory reduction.

pcuenca:
Thanks for the contribution! I agree that using `64` can be confusing, especially when trying to understand how everything works :)

PR #2070, opened 2024-05-15
Update agents.md

spacewaterbear:
fix bad typing in RetrieverTool class

### DIFF
--- a/agents.md
+++ b/agents.md
@@ -184,7 +184,7 @@ class RetrieverTool(Tool):
     }
     output_type = "text"
     
-    def __init__(self, vectordb: VectorStore, all_sources: str, **kwargs):
+    def __init__(self, vectordb: VectorStore, all_sources: list[str], **kwargs):
         super().__init__(**kwargs)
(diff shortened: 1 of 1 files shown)


agents.md
```diff
@@ -184,7 +184,7 @@ class RetrieverTool(Tool):
     }
     output_type = "text"
     
-    def __init__(self, vectordb: VectorStore, all_sources: str, **kwargs):
+    def __init__(self, vectordb: VectorStore, all_sources: list[str], **kwargs):
```
pcuenca, line 187: Actually, the comment in [L191](https://github.com/huggingface/blog/pull/2070/files#diff-b951327c37d8a1ed4d5ed18908681f6aca718ecdfe174aa39cf2237547f92fe7R191) says it's a string representation of a list, but both that and a list should work. cc @aymeric-roucher to use whichever one he thinks it's best.

aymeric-roucher: Both `List[str]` and `str` should work, so this line could be `Union[str, List[str]]`!

PR #2073, opened 2024-05-16
Fix GPU instance parameters

juliensimon:
From:

instance_type="p4de",
instance_size="2xlarge"

To:

instance_type="nvidia-a100",
instance_size="x2",


pcuenca (approved):
Not familiar with this, approving to unblock. cc @philschmid for confirmation

philschmid (approved):
Thank you!

PR #2111, opened 2024-05-30
Update pipeline inference snippet to use chat messages

pcuenca:


pcuenca:
Setting as draft to check out locally and update other posts at the same time

gemma.md
```diff
@@ -119,33 +121,29 @@ pip install -U "transformers==4.38.1" --upgrade
 The following snippet shows how to use `gemma-7b-it` with transformers. It requires about 18 GB of RAM, which includes consumer GPUs such as 3090 or 4090.
 
 ```python
-from transformers import AutoTokenizer
-import transformers
```
osanseviero, line 124: I think you can change the pip install above as well, no need to pin to specific version

osanseviero (approved):
Thanks!

gemma.md
pcuenca: Great call, fixed.

Rocketknight1 (approved):
This looks great, thanks!

PR #2145, opened 2024-06-13
Fix a missing Stas + broken links

muellerzr:
Sorry @stas00, got your HF handle wrong 😢


stas00:
aren't those HF hub usernames? it's just `stas` then

PR #2171, opened 2024-06-27
add zh translation for gemma2

chenglu:
@adeenayakup


adeenayakup (approved):
Yay!! Let's merge it. Thanks!

PR #2207, opened 2024-07-10
set better h1 title

julien-c:
cc @ariG23498


ariG23498:
Ah! Thanks for doing this @julien-c

PR #2220, opened 2024-07-16
FIX TYPO: Simple variable change in LoraConfig

Rishav-hub:
When loading get_peft_model the config passed was config but defined as lora_config. Just a typo fix!!


osanseviero (approved):
Thanks!

PR #2228, opened 2024-07-18
fix image for multi-lora-serving

nbroad1881:
had the same image twice

@datavistics


nbroad1881:
@osanseviero, please review when you can

datavistics:
Great catch!

datavistics (approved):
Thanks for catching this

Issue #2247, opened 2024-07-25
Validate `local` in _blog.yml matches a local file


coyotte508:
In https://github.com/huggingface/blog/blob/main/scripts/validate-yaml.ts add a check that `local` matches a file that exists locally

PR #2249, opened 2024-07-25
🦺 Check that _blog.yml only contains valid file paths

coyotte508:
Fix https://github.com/huggingface/blog/issues/2247


julien-c (commented):
lgtm, thanks @coyotte508

PR #2251, opened 2024-07-25
Video encoding for robotics datasets

aliberts:
Blog post on video encoding for robotics datasets.

Related PR: https://huggingface.co/datasets/huggingface/documentation-images/discussions/349


video-encoding.md
mfarre, line 1: - in Motivation & Contribution there is a detail to fix: "This means that with a ratio compression ratio"
- "The frames in-between are then compressed temporally as "differences" (P-frames or B-frames)." I think it would be great to quickly define P and B frames (if I am not mistaken, the first place where you can read about this is in the article that you link) because it is quite easy to explain the difference.
- In Criteria -> "decoding time" -> I wonder if it is really faster to load from video versus load a tar file with N images inside. Just asking because this is a common practice and some people might wonder how those compare. If loading from video with the proper library is even faster than loading a tar, I think it is an additional argument to go your way :-)
- In Metrics -> I think the nature of the video content of your policies will have an impact on the encoder results: it is not the same to encode cartoons than Star Wars :-) In your case, it would be great to open the door to explore this: maybe for synthetic policies you can be more aggressive on the encoder parameters than on real videos... or maybe is a function of  optical flow info  + light info + synthetic vs real to decide the proper codec. Just FYI some tech around this content-based encoding: https://bitmovin.com/encoding-service/per-title-encoding/ [Correction: I see it is covered in the next section. I keep the comment because maybe you find value on the bitmovin link]
- In Metrics -> just for future explorations, maybe is useful to take a look at how other people explored depth map compression using h264 or other video codecs. They were in a similar situation than you in the sense that they are using a codec optimized for human eye balls to compress data to be consumed by computers after all. Maybe there are some learnings that you can take from that.
- In Variables -> Encoding Parameters -> Video Codec -> I would define CRF (Constant Rate Factor). CRF is a quality-based variable rate encoding mode. This means that it is constantly calculating a trade-off between video-quality and size. CRF of 0 -> lossless [ah I see that you actually cover this, great! It is also super interesting the size of the GOPs that you explore, very cool]
- Future work -> I think it would be very interesting to explore 2 pass encoding, it usually helps when a video has complex scenes as well as helping to optimize how bitrate is distributed and get consistent quality across the video. In your case given the size of the GOPs maybe the impact is more limited but as it is such a small change in the commandline, I think is worth trying.

aliberts: Thank you @mfarre!
My anwers to your points:

> I think it would be great to quickly define P and B frames

I didn't want to get into more details here and keep it short. I've added a sentence to clarify that these are explained in the linked article

> I wonder if it is really faster to load from video versus load a tar file with N images inside.

Indeed, some numbers on that would be interesting as well. Intuitively I'd say video encoding still wins because it's essentially compression optimized for video (whereas tar has no such optimization) but having numbers to back this up on this would be best.

> In Metrics -> just for future explorations, maybe is useful to take a look at how other people explored depth map compression using h264 or other video codecs. They were in a similar situation than you in the sense that they are using a codec optimized for human eye balls to compress data to be consumed by computers

Good point! We do indeed have datasets that use depth maps but we're not using it (yet). I'm adding a remark on that.

> I think it would be very interesting to explore 2 pass encoding

I thought about it but I was afraid of really increasing encoding time (running this benchmark took a long time already). But it would be interesting to see what are the gains. Adding a note on that

Cadene (approved):
LGTM

pcuenca:
Sorry, I'm late for this review. There needs to be an entry in `_blog.yml` for the post to be linked from the blog page. I'll create a PR now.

PR #2254, opened 2024-07-26
Update quanto introduction blog charts

dacorvo:
The blog used to reference charts directly from the optimum-quanto repo that are now obsolete.
This now uses static assets in the blog documentation repo.


sayakpaul (approved):
Yeah nice and better example!

PR #2258, opened 2024-07-28
Fix typo in the Llama3.1 prompt template

wizzard0:
The llama 3.1 tool calling template says "orginal user question" instead of the "original user question". 

Unless it's the actual prompt pattern the model is trained with, this should affect the tool use performance; and I see this typo spreading to other tutorials and tools using Llama 3.1

### DIFF
--- a/llama31.md
+++ b/llama31.md
@@ -656,7 +656,7 @@ Llama 3.1 Instruct supports custom function calls from a single user message. Th
 ```
 <|begin_of_text|><|start_header_id|>system<|end_header_id|>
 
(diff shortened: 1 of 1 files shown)


pcuenca:
Hi @wizzard0! Actually, the typo is also present in the official documentation (see Step 3 in the [JSON-based tool calling section](https://llama.meta.com/docs/model-cards-and-prompt-formats/llama3_1/#json-based-tool-calling)). I doubt that the system prompt has to be supplied verbatim like that and I'd prefer to have the typo fixed, but I haven't really tested it comprehensively. Have you observed differences one way or another?

PR #2261, opened 2024-07-29
remove zephyr 141B

nbroad1881:
@osanseviero 

I removed zephyr 141B from supported models as it was removed from inf api


pcuenca (approved):
Thank you!

PR #2275, opened 2024-08-01
Online DPO

qgallouedec:


qgallouedec:
We've decided to opt for an in-depth documentation in TRL rather than a blog post.

PR #2303, opened 2024-08-15
Update infini-attention.md

thomwolf:
Proof reading and corrections


xrsrke (approved):
LGTM! Thanks

infini-attention.md
```diff
@@ -16,18 +16,18 @@ TLDR: Infini-attention's performance gets worse as we increase the number of tim
...
-> In standard attention, every token attends to every other token in the sequence, resulting in an attention matrix of size [seq_len, seq_len]. For each pair of tokens, we compute an attention score, and as the sequence length (seq_len) increases, the memory and computation requirements grow quadratically: Memory for the attention matrix is O(seq_len^2). For instance, a 10x increase in sequence length results in a 100x increase in memory requirements.
+> In standard attention (not-flash-attention), every token attends to every other token in the sequence, resulting in an attention matrix of size [seq_len, seq_len]. For each pair of tokens, we compute an attention score, and as the sequence length (seq_len) increases, the memory and computation requirements grow quadratically: Memory for the attention matrix is O(seq_len^2). For instance, a 10x increase in sequence length results in a 100x increase in memory requirements. Even memory efficient attention like Flash Attention still increase linearly with context length and are bottlenecked by single GPU memory, leading to a typical max context far lower than 1M tokens on today's GPUs.
```
pcuenca, line 24:
```suggestion
> In standard attention (not-flash-attention), every token attends to every other token in the sequence, resulting in an attention matrix of size [seq_len, seq_len]. For each pair of tokens, we compute an attention score, and as the sequence length (seq_len) increases, the memory and computation requirements grow quadratically: Memory for the attention matrix is O(seq_len^2). For instance, a 10x increase in sequence length results in a 100x increase in memory requirements. Even memory efficient attention methods like Flash Attention still increase linearly with context length and are bottlenecked by single GPU memory, leading to a typical max context far lower than 1M tokens on today's GPUs.
```

infini-attention.md
```diff
@@ -16,18 +16,18 @@ TLDR: Infini-attention's performance gets worse as we increase the number of tim
...
+Motivated by this, we explore an alternative approach to standard attention: infini-attention. The paper was released by researchers from Google in April 2024 [[link]](https://arxiv.org/abs/2404.07143). Instead of computing attention scores between every word, Infini-attention divides the sequence into segments, compresses earlier segments into a fixed buffer, and allows the next segment to retrieve memory from the earlier segments while limiting attention scores to words within the current segment.  A key advantage is it's fixed buffer size upper bound the total memory usage and it uses the same query within a segment to access information from both its own segment and the compressed memory which enables us to cheaply extend the context length for a pretrained model. In theory we can achieve infinite context length, as it only keeps a single buffer for all the memory of earlier segments. However, in reality compression limits the amount of information which can effectively been stored and the question is thus: how usably is the memory such compressed?
```
pcuenca, line 26:
```suggestion
Motivated by this, we explore an alternative approach to standard attention: infini-attention. The paper was released by researchers from Google in April 2024 [[link]](https://arxiv.org/abs/2404.07143). Instead of computing attention scores between every word, Infini-attention divides the sequence into segments, compresses earlier segments into a fixed buffer, and allows the next segment to retrieve memory from the earlier segments while limiting attention scores to words within the current segment.  A key advantage is its fixed buffer size upper bounds the total memory usage. It also uses the same query within a segment to access information from both its own segment and the compressed memory, which enables us to cheaply extend the context length for a pretrained model. In theory, we can achieve infinite context length, as it only keeps a single buffer for all the memory of earlier segments. However, in reality compression limits the amount of information which can effectively been stored and the question is thus: how usably is the memory such compressed?
```

infini-attention.md
```diff
@@ -36,14 +36,14 @@ We found the following rules helpful when implementing a new method and use it a
 
 + **Principle 1:** Start with the smallest model size that provides good signals, and scale up the experiments once you get good signals.
 + **Principle 2.** Always train a solid baseline to measure progress.
-+ **Principle 3.** To determine if a factor improves performance, train two identical models except for the difference in the factor being tested.
++ **Principle 3.** To determine if a modification improves performance, train two models fully identical excepted for the modification being tested.
```
pcuenca, line 39:
```suggestion
+ **Principle 3.** To determine if a modification improves performance, train two models identically except for the modification being tested.
```

infini-attention.md
```diff
@@ -188,26 +188,28 @@ def forward(...):
...
 
-**The evaluation**
+**Evaluating using the passkey retrieval task**
 
-The passkey retrieval task was first introduced by researchers from EPFL [[link]](https://arxiv.org/abs/2305.16300). This task is designed to evaluate a model's ability to retrieve information from long contexts where the location of the information is controllable. The input format for prompting a model is structured as follows:
+The passkey retrieval task was first introduced by researchers from EPFL [[link]](https://arxiv.org/abs/2305.16300). It's a task is designed to evaluate a model's ability to retrieve information from long contexts where the location of the information is controllable. The input format for prompting a model is structured as follows:
```
pcuenca, line 200:
```suggestion
The passkey retrieval task was first introduced by researchers from EPFL [[link]](https://arxiv.org/abs/2305.16300). It's a task designed to evaluate a model's ability to retrieve information from long contexts where the location of the information is controllable. The input format for prompting a model is structured as follows:
```

infini-attention.md
```diff
@@ -218,49 +220,53 @@ Graduate Education
 The Department of Physics and Astronomy offers a program leading to the Master of Science degree in physics. The program is designed to provide students with a broad background in
 ```
 
-Based on these results, it suggests that the model does use compressed memory, so we decided to scale up our experiments by continually pretraining a Llama 3 8B. Unfortunately, the model failed to pass the needle evaluation when the needle was placed in an earlier segment. This led us to inspect the balance factors across all layers, and based on Figure 3a, and Figure 3b, we found that 95% of the weights are centered around 0.5. Recall that for a weight to converge to an ideal range, it depends on two general factors: the step size and the magnitude of the gradients. However, since Adam normalizes the gradients to a magnitude of 1, the question now is: are we setting up the training to allow it to converge?
+Based on these results, the model appear to in fact use the compressed memory. We decided to scale up our experiments by continually pretraining a Llama 3 8B. Unfortunately, the model failed to pass the needle evaluation when the needle was placed in an earlier segment.
```
pcuenca, line 223:
```suggestion
Based on these results, the model appears to in fact use the compressed memory. We decided to scale up our experiments by continually pretraining a Llama 3 8B. Unfortunately, the model failed to pass the needle evaluation when the needle was placed in an earlier segment.
```

infini-attention.md
```diff
@@ -218,49 +220,53 @@ Graduate Education
...
+
+We decided to inspect the balance factors (factor balancing the amont of compressed and not-compressed memory) across all layers. Based on Figure 3a and Figure 3b, we found that about 95% of the weights are centered around 0.5. Recall that for a weight to converge to an ideal range, it depends on two general factors: the step size and the magnitude of the gradients. However, Adam normalizes the gradients to a magnitude of 1 so the question became: are the training hyper-parameters the right ones to allow the finetuning to converge?
```
pcuenca, line 225:
```suggestion
We decided to inspect the balance factors (factor balancing the amount of compressed and not-compressed memory) across all layers. Based on Figure 3a and Figure 3b, we found that about 95% of the weights are centered around 0.5. Recall that for a weight to converge to an ideal range, it depends on two general factors: the step size and the magnitude of the gradients. However, Adam normalizes the gradients to a magnitude of 1 so the question became: are the training hyper-parameters the right ones to allow the finetuning to converge?
```

infini-attention.md
```diff
@@ -218,49 +220,53 @@ Graduate Education
...
+> Conclusion: we decided to go with a global learning rate of  3.0x10^-4 and a gating learning rate of 0.01 which should allows the gating function to converge.
 
-Now the balance factors in Infini-attention are trainable, but 200m llama's loss went NaN after 20B tokens (we tried learning rates from 0.001 to 1.0e-6). So, we ran some generation at the 20B tokens checkpoint (10k training steps), as you can see in the Figure 4a, now it can continue the exact content and recall identities (if knock out the memory, it generated trash). But it isn't able to recall the needle in the 1st segment (it does so reliably within the segment).
+With these hyper-parameters the balance factors in Infini-attention are trainable, but we observed that the 200M llama's loss went NaN after 20B tokens (we tried learning rates from 0.001 to 1.0e-6). We investigated a few generation at the 20B tokens checkpoint (10k training steps) which you can see in the Figure 4a. The model now continue the exact content and recall identities (if knock out the memory, it generated trash).
```
pcuenca, line 244:
```suggestion
With these hyper-parameters the balance factors in Infini-attention are trainable, but we observed that the 200M llama's loss went NaN after 20B tokens (we tried learning rates from 0.001 to 1.0e-6). We investigated a few generations at the 20B tokens checkpoint (10k training steps) which you can see in Figure 4a. The model now continue the exact content and recall identities (if the memory is knocked out, it generates trash). 
```

I don't understand the "continue" sentence.

infini-attention.md
```diff
@@ -218,49 +220,53 @@ Graduate Education
...
 
 ![Figure 4c: global weights across training steps](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/185_infini_attention/exp_51_global_weights_across_training_steps.png)
 
-But needle evaluation fails completely when the needle is placed in the 1st segment (100% success when placed in the 2nd segment, out of 2 segments total). And based on Figure 4b, after 5,000 steps, the balance factors stopped changing.  While we made some progress, we weren't out of the woods yet. The balance factors were still not behaving as we hoped, so we had to dig deeper and make some more adjustments.
+But it is still not able to recall the needle from one segment to the other (it does so reliably within the segment). Needle evaluation fails completely when the needle is placed in the 1st segment (100% success when placed in the 2nd segment, out of 2 segments total). As showed in Figure 4b, we also observed that the balance factors stopped changing after 5,000 steps.  While we made some progress, we weren't such not yet out of the woods. The balance factors were still not behaving as we hoped. We decided to dig deeper and make more adjustments.
```
pcuenca, line 250:
```suggestion
But it is still not able to recall the needle from one segment to the other (it does so reliably within the segment). Needle evaluation fails completely when the needle is placed in the 1st segment (100% success when placed in the 2nd segment, out of 2 segments total). As showed in Figure 4b, we also observed that the balance factors stopped changing after 5,000 steps.  While we made some progress, we were not yet out of the woods. The balance factors were still not behaving as we hoped. We decided to dig deeper and make more adjustments.
```

infini-attention.md
```diff
@@ -218,49 +220,53 @@ Graduate Education
...
 
 
 ## Section 5: No weight decay on balance factors
 
-So we inspected the balance factor once again and saw some progress: approximately 95% of the heads now show a global weight ranging from 0.4 to 0.5, and none of the heads have a global weight greater than 0.6, but the weights still aren't in the ideal range. Another reason could be weight decay, which encourages a small L2 norm of balance factors, leading sigmoid values close to zero to center around 0.5.
+Inspecting in details the balance factor once again, we saw some progress: approximately 95% of the heads now show a global weight ranging from 0.4 to 0.5, and none of the heads have a global weight greater than 0.6. But the weights still aren't in the ideal range.
```
pcuenca, line 255:
```suggestion
Inspecting in detail the balance factor once again, we saw some progress: approximately 95% of the heads now show a global weight ranging from 0.4 to 0.5, and none of the heads have a global weight greater than 0.6. But the weights still aren't in the ideal range.
```

infini-attention.md
```diff
@@ -218,49 +220,53 @@ Graduate Education
...
 
-Another reason could be that we used too small a rollout. In the 200m experiment, we only used 4 rollouts *, and in the 8b experiment, we only used 2 rollouts (8192**2). It makes sense that using a larger rollout would pressure the model to learn how to compress and use the memory well. So we decided to increase the number of rollouts to 16 and use no weight decay.  So we decided to scale down the context length to 1024 context length, with 16 rollouts, so the segment length is 64.
+We thought of another potential reason: weight decay, which encourages a small L2 norm of balance factors, leading sigmoid values to converge close to zero and factor to center around 0.5.
+
+Yet another potential reason could be that we used too small a rollout. In the 200m experiment, we only used 4 rollouts, and in the 8b experiment, we only used 2 rollouts (8192**2). Using a larger rollout should incentives the model to compress and use the memory well. So we decided to increase the number of rollouts to 16 and use no weight decay. We scaled down the context length to 1024 context length, with 16 rollouts, getting segment lengths of 64.
```
pcuenca, line 259:
```suggestion
Yet another potential reason could be that we used too small a rollout. In the 200m experiment, we only used 4 rollouts, and in the 8b experiment, we only used 2 rollouts (8192**2). Using a larger rollout should incentive the model to compress and use the memory well. So we decided to increase the number of rollouts to 16 and use no weight decay. We scaled down the context length to 1024 context length, with 16 rollouts, getting segment lengths of 64.
```

infini-attention.md
```diff
@@ -218,49 +220,53 @@ Graduate Education
...
 
 ![Figure 5c: global weights across training](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/185_infini_attention/exp57_global_weights_across_training.png)
 
-As you see, the global weights are now distributed across the range from 0 to 1, with 10% of heads having a global weight between 0.9 and 1.0, even though after 18k steps, most heads stopped changing their global weights. But now we are quite confident in saying that we have set up the experiments to allow convergence if gradient descent wants to. The only question now is whether Infini-attention works well enough.
+As you can see, global weights are now distributed across the range from 0 to 1, with 10% of heads having a global weight between 0.9 and 1.0, even though after 18k steps, most heads stopped changing their global weights. We were then quite confident that the experiments were setup to allow convergence if the spirits of gradient descents are with us. The only question remaining was whether the general approach of Infini-attention could works well enough.
```
pcuenca, line 267:
```suggestion
As you can see, global weights are now distributed across the range from 0 to 1, with 10% of heads having a global weight between 0.9 and 1.0, even though after 18k steps, most heads stopped changing their global weights. We were then quite confident that the experiments were setup to allow convergence if the spirits of gradient descent are with us. The only question remaining was whether the general approach of Infini-attention could works well enough.
```

infini-attention.md
```diff
@@ -271,12 +277,14 @@ The following valuations were run at 1.5B tokens.
...
+Unfortunately, despite these progress, we found that Infini-attention was not convincing enough in our experiments and in particular not reliable enough. At this stage of our reproduction we are still of the opinion that Ring Attention [[link]](https://x.com/Haojun_Zhao14/status/1815419356408336738), YaRN [[link]](https://arxiv.org/abs/2309.00071) and rope scaling [[link]](https://arxiv.org/abs/2309.16039) are better options for extending a pretrained model to longer context length. 
+
+These later technics still come with large resource requirements for very large model sizes (e.g., 400B and beyond). we thus till think that exploring compression techniques or continuing to push the series of experiments we've bee describing in this blog post is of great interest for the community and are are excited to follow and try new techniques that may be developped and overcome some of the limitation of the present work.
```
pcuenca, line 287:
```suggestion
These later techniques still come with large resource requirements for very large model sizes (e.g., 400B and beyond). We thus still think that exploring compression techniques or continuing to push the series of experiments we've been describing in this blog post is of great interest for the community and are excited to follow and try new techniques that may be developed and overcome some of the limitations of the present work.
```

pcuenca (approved):

pcuenca:
Merging now, we can keep iterating if we need to. Thanks a lot!

PR #2338, opened 2024-09-07
Pytorch DDP tutorial. Update pytorch-ddp-accelerate-transformers.md

danilyef:
Correction in DDP Pytorch:
- Rank instead of device
- ddp_model instead of model


muellerzr (approved):
Thanks for correcting this bug!

PR #2339, opened 2024-09-09
Mohamed/1.58 llm new era of extreme quantization

MekkCyber:
Adding the following blogpost : **1.58 Bit LLM a new era of extreme quantization**


MekkCyber:
cc @SunMarc

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+In this blog, we'll dive into the process of 1.58 bit quantization. First, we'll review the key research paper that explains the theory behind this quantization method. Next, we'll address the challenges of fine-tuning quantized bitnet models and how we managed to train the best-performing 7/8B BitNet on downstream tasks. Finally, we'll discuss optimizing performance for fast inference, including writing specialized kernels to speed up processing. This comprehensive look will help you understand how to get the best out of LLMs while reducing their computational costs.
+
+BitNet b1.58 offers a unique approach to the challenges of extreme quantization. By representing each parameter with just 1.58 bits using a ternary system (-1, 0, 1). However, a key limitation is that BitNet b1.58 goes one step further and pretrains a model in 1.58 bits. While this is cool, not everybody has budget to pretrain such models. Thus we explore in this blog post that if we add a few tricks we can simply fine-tune existing models to 1.58 bits. !! Keep reading to find out !
```
osanseviero, line 17:
```suggestion
BitNet b1.58 offers a unique approach to the challenges of extreme quantization. It represents each parameter with just 1.58 bits using a ternary system (-1, 0, 1). However, a key limitation is that BitNet b1.58 goes one step further and trains a model in 1.58 bits. While this is cool, not everybody has the budget to pre-train such models. Thus, in this blog post, we explore that if we add a few tricks, we can simply fine-tune existing models to 1.58 bits! Keep reading to find out!
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+- [Results in the paper](#results-in-the-paper)
+- [Pretraining results in 1.58b](#pretraining-results-in-158b)
+- [Finetuning in 1.58bit](#finetuning-in-158bit)
+- [Kernels used & Benchmarks](#kernels-used--benchmarks)
+- [Additional Resources](#additional-resources)
+- [Integration in Transformers & how to use](#integration-in-transformers--how-to-use)
+- [Conclusion](#conclusion)
+- [Additional Resources](#additional-resources)
+
+## What is BitNet ?
```
osanseviero, line 31:
```suggestion
## What is BitNet?
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+BitNet replaces traditional Linear layers in Multi-Head Attention and Feed-Forward Networks with specialized layers called BitLinear with ternary precision. The main obstacle of training in ternary precision is that the weight values are discretized (via the `round()` function) and thus non-differentiable. BitLinear solves this with a nice trick: [STE (Straight Through Estimator)](https://arxiv.org/pdf/1903.05662). The STE allows gradients to flow through the non-differentiable rounding operation by approximating its gradient as 1 (treating `round()` as equivalent to the identity function). Another way to view it is that, instead of stopping the gradient at the rounding step, the STE lets the gradient pass through as if the rounding never occurred, enabling weight updates using standard gradient-based optimization techniques.
```
osanseviero, line 33:
```suggestion
BitNet replaces traditional Linear layers in Multi-Head Attention and Feed-Forward Networks with specialized layers called BitLinear with ternary precision. The main obstacle to training in ternary precision is that the weight values are discretized (via the `round()` function) and thus non-differentiable. BitLinear solves this with a nice trick: [STE (Straight Through Estimator)](https://arxiv.org/abs/1903.05662). The STE allows gradients to flow through the non-differentiable rounding operation by approximating its gradient as 1 (treating `round()` as equivalent to the identity function). Another way to view it is that, instead of stopping the gradient at the rounding step, the STE lets the gradient pass through as if the rounding never occurred, enabling weight updates using standard gradient-based optimization techniques.
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+ The BitLinear layers quantize the weights using ternary precision (with values of -1, 0, and 1) and quantize the activations to 8-bit precision. After performing the matrix multiplication operation, the data is dequantized, allowing the model to continue with high-precision operations.
```
osanseviero, line 35:
```suggestion
The BitLinear layers quantize the weights using ternary precision (with values of -1, 0, and 1) and quantize the activations to 8-bit precision. After performing the matrix multiplication operation, the data is dequantized, allowing the model to continue with high-precision operations. 
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+ The BitLinear layers quantize the weights using ternary precision (with values of -1, 0, and 1) and quantize the activations to 8-bit precision. After performing the matrix multiplication operation, the data is dequantized, allowing the model to continue with high-precision operations. 
+
+<figure style="text-align: center;">
+  <img src="assets/1.58llm_extreme_quantization/bitlinear.png" alt="Alt Text" />
```
osanseviero: Except for the thumbnail, all images should be in https://huggingface.co/datasets/huggingface/documentation-images/tree/main/blog

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+  <figcaption>The architecture of BitNet with BitLinear layers</figcaption>
+</figure>
+
+In the BitNet layer, during training, we start by making the weights ternary. We first compute the average of the weight matrix absolute value, and use this value as a scale, then divide the weights by the scale, round the values, and keep them between -1 and 1 then rescale them
```
osanseviero, line 42:
```suggestion
In the BitNet layer, during training, we start by making the weights ternary. We first compute the average of the weight matrix absolute value and use this value as a scale, then divide the weights by the scale, round the values, keep them between -1 and 1, and then rescale them 
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+$$
+X_q = \text{clamp}_{[-127,128]}(\text{round}(X*scale))
+$$
+
+$$
+X_{dequantized} = X_q * scale_x
+$$
+
+However during inference, we just quantize the weights, to ternary weights, without rescaling, and we do the same thing with activations on 8 bits, and then do a matmul with an efficient kernel before dividing by both the weight scales, and activations scales. This should to a notable improvement in inference speed especially with appropriate optimized hardware.
```
osanseviero, line 70:
```suggestion
However, during inference, we just quantize the weights to ternary weights without rescaling. We do the same thing with activations on 8 bits, and then we do a matmul with an efficient kernel before dividing by both the weight scales and activations scales. This should result in a notable improvement in inference speed, especially with appropriate optimized hardware.
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+To maintain the variance of the output after quantization, we apply Layer Normalization (LN) before quantizing the activations. This normalization ensures that the variance remains around 1:
+
+$$
+\text{LN}(x) = \frac{x - E(x)}{\sqrt{\text{Var}(x) + \epsilon}}
+$$
+
+where ϵ is a small number to prevent overflow.
+
+After the introduction of the round function which is non differentiable during the training phase, the backward pass computation needs the introduction of a [STE](https://arxiv.org/pdf/1903.05662) as mentioned before using the trick in the code below :
```
osanseviero, line 80:
```suggestion
After the introduction of the round function, which is nondifferentiable during the training phase, the backward pass computation needs the introduction of a [STE](https://arxiv.org/pdf/1903.05662) as mentioned before using the trick in the code below : 
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+$$
+\text{LN}(x) = \frac{x - E(x)}{\sqrt{\text{Var}(x) + \epsilon}}
+$$
+
+where ϵ is a small number to prevent overflow.
+
+After the introduction of the round function which is non differentiable during the training phase, the backward pass computation needs the introduction of a [STE](https://arxiv.org/pdf/1903.05662) as mentioned before using the trick in the code below : 
+
+```python
```
osanseviero, line 82: Add imports of things being used too please

```
import torch
import torch.nn as nn
import torch.nn.functional as F
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+    Only for training, and kernel optimization is needed for efficiency.
+    """
+    def forward(self, x):
+        w = self.weight
+        x_norm = LN(x)
+        # A trick for implementing Straight−Through−Estimator (STE) using detach()
+        x_quant = x_norm + (activation_quant(x_norm) − x_norm).detach()
+        w_quant = w + (weight_quant(w) − w).detach()
+        y = F.linear(x_quant, w_quant)
+        return y
```
osanseviero, line 103:
```suggestion
def activation_quant(x):
    scale = 127.0 / x.abs().max(dim=-1, keepdim=True).values.clamp_(min=1e-5)
    y = (x * scale).round().clamp_(-128, 127) / scale
    return y
 
def weight_quant(w):
    scale = 1.0 / w.abs().mean().clamp_(min=1e-5)
    u = (w * scale).round().clamp_(-1, 1) / scale
    return u

class BitLinear(nn.Linear):
    """
    Only for training; kernel optimization is needed for efficiency.
    """
    def forward(self, x):
        w = self.weight
        x_norm = LN(x)
        
        # A trick for implementing Straight−Through−Estimator (STE) using detach()
        x_quant = x_norm + (activation_quant(x_norm) - x_norm).detach()
        w_quant = w + (weight_quant(w) - w).detach()
        
        # Perform quantized linear transformation
        y = F.linear(x_quant, w_quant)
        return y
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+    scale = 1.0 / w.abs().mean().clamp_(min=1e-5)
+    u = (w * scale).round().clamp_(-1, 1) / scale
+    return u
+class BitLinear(nn.Linear):
+    """
+    Only for training, and kernel optimization is needed for efficiency.
+    """
+    def forward(self, x):
+        w = self.weight
+        x_norm = LN(x)
```
osanseviero, line 98: where does `LN` come from?

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+        # A trick for implementing Straight−Through−Estimator (STE) using detach()
+        x_quant = x_norm + (activation_quant(x_norm) − x_norm).detach()
+        w_quant = w + (weight_quant(w) − w).detach()
+        y = F.linear(x_quant, w_quant)
+        return y
+```
+
+## Results in the paper
```
osanseviero, line 106:
```suggestion
## Results in the Paper
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+### Computational efficiency :
```
osanseviero, line 108: (same for other headings)
```suggestion
### Computational Efficiency :
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+For BitNet, the energy consumption during matrix multiplication is primarily influenced by the addition operations, given that the weights are ternary. The multiplication operations are limited to scaling the output with the weights scale, and activations scale, which provides enormous energy and computation savings.
```
osanseviero, line 110:
```suggestion
For BitNet, the energy consumption during matrix multiplication is primarily influenced by the addition operations, given that the weights are ternary. The multiplication operations are limited to scaling the output with the weights scale and activations scale, which provides enormous energy and computation savings.
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+### Downstream tasks
+
+BitNet is effective in delivering strong performance compared to baseline methods, especially at lower bit levels. According to the paper, BitNet achieves scores that are on par with 8-bit models, but with significantly lower inference costs. In the case of 4-bit models, methods that only quantize weights outperform those that quantize both weights and activations, as activations are harder to quantify. However, BitNet, which uses 1.58-bit weights, surpasses both weight-only and weight-and-activation quantization methods.
```
osanseviero, line 114:
```suggestion
BitNet is effective in delivering strong performance compared to baseline methods, especially at lower bit levels. According to the paper, BitNet achieves scores that are on par with 8-bit models but with significantly lower inference costs. In the case of 4-bit models, methods that only quantize weights outperform those that quantize both weights and activations, as activations are harder to quantify. However, BitNet, which uses 1.58-bit weights, surpasses both weight-only and weight-and-activation quantization methods.
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+## Pretraining results in 1.58b
+
+We tried to reproduce the results of the BitNet paper, we started with a small dataset, tinystories, and a Llama3 8B model architecture. It seemed that adding a normalization function like in the paper, improves the performance, for example after 2000 steps of training, we had a perplexity on the validation part of the dataset equal to 6.3 without a normalization, and 5.9 with normalization, but in both cases, the training was stable.
```
osanseviero, line 118:
```suggestion
We tried to reproduce the results of the BitNet paper; we started with a small dataset, tinystories, and a Llama3 8B model architecture. It seemed that adding a normalization function like in the paper improves the performance. For example, after 2000 steps of training, we had a perplexity on the validation part of the dataset equal to 6.3 without normalization, and 5.9 with normalization, but in both cases, the training was stable.
```

osanseviero, line 118: Please add link to tinystories if possible

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/pretraining.png" alt="Alt Text" style="width: 80%;"/>
+  <figcaption>Pretraining plots without (blue) & with (green) layer normalisation </figcaption>
+</figure>
+
+While this approach looks very interesting for pretraining, only few institutions can afford pretraining models at the necessary scale. However, there are already a wide range of strong pretrained models and it would be extremely useful if they could be converted to 1.58bit after pretraining. Thus we set out on an investigation to see if we can make 1.58 fine-tuning work.
```
osanseviero, line 125:
```suggestion
While this approach looks very interesting for pretraining, only a few institutions can afford pretraining models at the necessary scale. However, there is already a wide range of strong pretrained models, and it would be extremely useful if they could be converted to 1.58bit after pretraining. Thus we set out on an investigation to see if we can make 1.58 fine-tuning work. 
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/pretraining.png" alt="Alt Text" style="width: 80%;"/>
+  <figcaption>Pretraining plots without (blue) & with (green) layer normalisation </figcaption>
+</figure>
+
+While this approach looks very interesting for pretraining, only few institutions can afford pretraining models at the necessary scale. However, there are already a wide range of strong pretrained models and it would be extremely useful if they could be converted to 1.58bit after pretraining. Thus we set out on an investigation to see if we can make 1.58 fine-tuning work. 
+
+## Finetuning in 1.58bit
+
+When we began finetuning from Llama3 weights instead pretraining starting from random weights, the model performed slightly better, but not as good as we expected :
```
osanseviero, line 129:
```suggestion
When we began finetuning from Llama3 weights instead of pretraining starting from random weights, the model performed slightly better but not as well as we expected: 
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+## Finetuning in 1.58bit
+
+When we began finetuning from Llama3 weights instead pretraining starting from random weights, the model performed slightly better, but not as good as we expected : 
+
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/finetuning_basic.png" alt="Alt Text" style="width: 80%;"/>
+  <figcaption>Finetuning plot compared to pretraining plot</figcaption>
+</figure>
+
+To understand why, we tried to inspect both weight distributions of the randomly initialized model, and pre-trained model to identify potential issues.
```
osanseviero, line 136:
```suggestion
To understand why, we tried to inspect both the weight distributions of the randomly initialized model and the pre-trained model to identify potential issues.
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+    <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/poids_aléatoires.png" alt="Image 1" style="width: 400px;" />
+    <figcaption>Random weights distribution (2 merged stds)</figcaption>
+  </figure>
+  <figure style="text-align: center;">
+    <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/poids_llama3.png" alt="Image 2" style="width: 400px;" />
+    <figcaption>Pre-trained Llama3 weights distribution</figcaption>
+  </figure>
+</div>
+
+And the scale values for the two distribution are respectively :
```
osanseviero, line 149:
```suggestion
And the scale values for the two distributions are, respectively : 
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+
+The initial random weight distribution is a mix of two normal distributions:
+
+- One with a standard deviation (std) of 0.025
+- Another with a std of 0.025 / sqrt(2 * num_hidden_layers) = 0.00325
+
+This results from using different stds for column linear and row linear weights in nanotron. In the quantized version, all matrices have only 2 weight scales (50.25 and 402), which are the inverse of the mean absolute value of the weights for each matrix : `scale = 1.0 / w.abs().mean().clamp_(min=1e-5)`
+
+- For scale = 50.25, w.abs().mean() = 0.0199, leading to std = 0.025 which matches our first standard deviation. The formula used to derive the std is based on the expectation of the half-normal distribution of |w|: E(∣w∣)=std(w)⋅sqrt(2/π)
```
osanseviero, line 170:
```suggestion
- For scale = 50.25, `w.abs().mean() = 0.0199`, leading to std = 0.025 which matches our first standard deviation. The formula used to derive the std is based on the expectation of the half-normal distribution of |w|: E(∣w∣)=std(w)⋅sqrt(2/π)
```

osanseviero, line 170: Consider adding some latex here

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+- For scale = 402, w.abs().mean() = 0.0025, leading to std = 0.00325
+
+On the other hand, the pretrained weight's distribution looks like a normal distribution with an std ~ 0.013
+
+Clearly the pretrained model starts with more information (scales) while the randomly initialized model starts with practically no information and adds to it over time. Our conclusion was that starting with random weights gives the model minimal initial information, enabling a gradual learning process, while during finetuning, the introduction of BitLinear layers overwhelms the model into losing all its prior information.
```
osanseviero, line 175:
```suggestion
Clearly, the pretrained model starts with more information (scales), while the randomly initialized model starts with practically no information and adds to it over time. Our conclusion was that starting with random weights gives the model minimal initial information, enabling a gradual learning process, while during finetuning, the introduction of BitLinear layers overwhelms the model into losing all its prior information.
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+To improve the results of the finetuning, we tried different technics , like instead of using per tensor quantization, we tried per row, and per column quantization, to keep more information from the weights of Llama3. We also tried to change the way of how the scale is computed, by instead of just taking the mean absolute value of the weights, as a scale, we take the mean absolute value of the outliers as a scale (an outlier value is a value that exceeds k*mean_absolute_value where k is a constant we tried to vary in our experiments) but we didn’t notice any big improvement.
```
osanseviero, line 177:
```suggestion
To improve the results of the finetuning, we tried different techniques, like instead of using per tensor quantization, we tried per row and per column quantization to keep more information from the weights of Llama3. We also tried to change the way how the scale is computed, by instead of just taking the mean absolute value of the weights as a scale, we take the mean absolute value of the outliers as a scale (an outlier value is a value that exceeds k*mean_absolute_value where k is a constant we tried to vary in our experiments), but we didn’t notice any big improvement. 
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+def scale_outliers(tensor, k=1):
+    mean_absolute_value = torch.mean(torch.abs(tensor))
+    outliers = tensor[torch.abs(tensor) > mean_absolute_value]
+    mean_outlier_value = torch.mean(torch.abs(outliers))
+    return mean_outlier_value
+def weight_quant_scaling(w):
+    scale = 1.0 / scale_outliers(w).clamp_(min=1e-5)
+    u = (w * scale).round().clamp_(-1, 1) / scale
+    return u
+
```
osanseviero, line 189: There was an unused `k`. Is this what it should be doing?

```suggestion
def scale_outliers(tensor, threshold_factor=1):
    mean_absolute_value = torch.mean(torch.abs(tensor))
    threshold = threshold_factor * mean_absolute_value
    outliers = tensor[torch.abs(tensor) > threshold]
    mean_outlier_value = torch.mean(torch.abs(outliers))
    return mean_outlier_value

def weight_quant_scaling(w):
    scale = 1.0 / scale_outliers(w).clamp_(min=1e-5)
    quantized_weights = (w * scale).round().clamp_(-1, 1) / scale
    return quantized_weights
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+```
+
+We observed that both the random weights and Llama3 weights resulted in losses starting at approximately the same value of 13. This suggests that the Llama3 model loses all of its prior information when quantization is introduced. To further investigate how much information the model loses during this process, we experimented with per-group quantization.
+
+As a sanity check, we first set the group size to 1, which essentially means no quantization. In this scenario, the loss started at 1.45, as expected. However, when we increased the group size to 2, the loss jumped to around 11. This indicates that even with a minimal group size of 2, the model still loses nearly all of its information.
+
+To address this issue, we considered the possibility of introducing quantization gradually, rather than applying it abruptly to the weights and activations for each tensor. To achieve this, we implemented a lambda value to control the process :
```
osanseviero, line 196:
```suggestion
To address this issue, we considered the possibility of introducing quantization gradually rather than applying it abruptly to the weights and activations for each tensor. To achieve this, we implemented a lambda value to control the process : 
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+
+To address this issue, we considered the possibility of introducing quantization gradually, rather than applying it abruptly to the weights and activations for each tensor. To achieve this, we implemented a lambda value to control the process : 
+
+```python
+lambda_ = ?
+x_quant = x + lambda_ * (activation_quant(x) - x).detach()
+w_quant = w + lambda_ * (weight_quant(w) - w).detach()
+```
+
+When `lambda` is set to 0, there is essentially no quantization occurring, while at `lambda`=1, full quantization is applied.
```
osanseviero, line 204:
```suggestion
When `lambda` is set to 0, there is essentially no quantization occurring, while at `lambda=1`, full quantization is applied.
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+```
+
+With this configuration, after 2000 steps we have : 
+
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/lambda_training_step.png" alt="Alt Text" style="width: 80%;"/>
+  <figcaption>Finetuning plot with lambda = min(2*training_step/total_training_steps, 1)</figcaption>
+</figure>
+
+Our fine-tuning method shows better convergence overall. You can observe a slight increase in the loss curve around 1,000 steps, which corresponds to when we begin approaching `lambda`=1, or full quantization. However, immediately after this point, the loss starts to converge again, leading to an improved perplexity of approximately 4.
```
osanseviero, line 232:
```suggestion
Our fine-tuning method shows better convergence overall. You can observe a slight increase in the loss curve around 1,000 steps, which corresponds to when we begin approaching `lambda=1`, or full quantization. However, immediately after this point, the loss starts to converge again, leading to an improved perplexity of approximately 4.
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/exp-randoms.png" alt="Alt Text" style="width: 80%;"/>
+  <figcaption>Different Pre-training plots with different learning rates</figcaption>
+</figure>
+
+None of the models trained from random weights performed better than our fine-tuned model. The best perplexity we achieved with those models was 26, which falls short compared to the results from our fine-tuning approach.
+
+### Scaling to 100B !
+
+We scaled our experiments to 100 billion tokens to see if we could match the performance Llama3 8B. We conducted longer training runs, starting from our best-performing checkpoint from the shorter runs, and continued fine-tuning for 45,000 steps. We experimented with different learning rates, and while the model performed closely to the Llama3 model in some metrics, on average, it still lagged behind.
```
osanseviero, line 312:
```suggestion
We scaled our experiments to 100 billion tokens to see if we could match the performance of Llama3 8B. We conducted longer training runs, starting from our best-performing checkpoint from the shorter runs, and continued fine-tuning for 45,000 steps. We experimented with different learning rates, and while the model performed closely to the Llama3 model in some metrics, on average, it still lagged behind.
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/metric_avg.png" alt="Alt Text" style="width: 60%;"/>
+  <figcaption>Average evaluation during the training for different lrs</figcaption>
+</figure>
+
+### Experiments on Smaller Models
+
+In our initial experiments with smaller models like Smol LLMs, we observed that the warmup quantization technique didn’t yield as much improvement as it did with larger models. This suggests that the effectiveness of warmup quantization could be more closely related to model size and complexity.
+
+For example, here are the loss curves for the Smol LLM 135M model, comparing warmup quantization with full quantization from the start. Interestingly, the curves closely align, and the resulting perplexities aren't significantly different.
```
osanseviero, line 331: I suggest adding more links to models and datasets

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+  <figcaption>Metrics comparison with Llama models for the model trained on 100B tokens</figcaption>
+</figure>
+
+For the 70B model
+
+## Kernels used & Benchmarks
+
+To benefit from the BitNet low-precision weights, we pack them into an `int8` tensor. During inference, these weights must be unpacked before performing matrix multiplication. We implemented custom kernels in Cuda and Triton to handle the on-the-fly unpacking during the matrix multiplication process. For the matrix multiplication itself, we employed the cached tiled matrix multiplication technique. To fully grasp this approach, let's first review some Cuda programming fundamentals.
+
+### **Basic GPU Concepts: Threads, Blocks, and Shared Memory**
```
osanseviero, line 368: Very cool section :fire:

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+- **Redundant Data Access**: In matrix multiplication, many elements of the input matrices are used multiple times. If each thread fetches the required data from global memory independently, the same data might be loaded into the GPU multiple times, leading to inefficiency. For example, if each thread is used to compute a single element in the output matrix, the thread responsible for calculating the element at position (i, j) will need to load the i-th row of matrix A and the j-th column of matrix B from global memory. However, other threads, such as the one computing the element at position (i+1, j), cannot reuse this data and will have to reload the same j-th column from global memory again.
+
+### **The Idea of Tiling**
+
+Tiling is a technique used to address these challenges and it was mainly used in FlashAttention to improve the kernels efficiency. The basic idea is to divide the matrices into smaller sub-matrices, called tiles, which can fit into the shared memory of the GPU. Instead of computing the entire output matrix in one go, the computation is broken down into smaller pieces that are processed tile by tile.
```
osanseviero, line 384:
```suggestion
Tiling is a technique used to address these challenges, and it was mainly used in FlashAttention to improve the kernel's efficiency. The basic idea is to divide the matrices into smaller sub-matrices, called tiles, which can fit into the shared memory of the GPU. Instead of computing the entire output matrix in one go, the computation is broken down into smaller pieces that are processed tile by tile.
```

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,566 @@
...
+input_ids = tokenizer.encode(input_text, return_tensors="pt").cuda()
+output = model_.generate(input_ids, max_length=100, do_sample=False)
+generated_text = tokenizer.decode(output[0], skip_special_tokens=True)
+print(generated_text)
+```
+
+With this code, everything is managed seamlessly behind the scenes, so there's no need to worry about additional complexities. 
+
+## Conclusion
+In conclusion, as (LLMs) continue to expand, reducing their computational demands through quantization is essential. This blog has explored the approach of 1.58-bit quantization, which uses a ternary weights. While pretraining models in 1.58 bits is resource-intensive, we’ve demonstrated that, with some tricks, it’s possible to fine-tune existing models to this precision level, achieving efficient performance without sacrificing accuracy. By optimizing inference speed and utilizing specialized kernels, BitNet b1.58 opens new possibilities for making LLMs more practical and scalable.
```
osanseviero, line 555:
```suggestion
In conclusion, as (LLMs) continue to expand, reducing their computational demands through quantization is essential. This blog has explored the approach of 1.58-bit quantization, which uses ternary weights. While pretraining models in 1.58 bits are resource-intensive, we’ve demonstrated that, with some tricks, it’s possible to fine-tune existing models to this precision level, achieving efficient performance without sacrificing accuracy. By optimizing inference speed and utilizing specialized kernels, BitNet b1.58 opens new possibilities for making LLMs more practical and scalable.
```

osanseviero (commented):
Very nice read! I left very minor/rewording comments

1.58llm_extreme_quantization.md
MekkCyber: It's defined in the section : What is BitNet?

1.58llm_extreme_quantization.md
MekkCyber: Done

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,576 @@
...
+- [Finetuning in 1.58bit](#finetuning-in-158bit)
+- [Kernels used & Benchmarks](#kernels-used--benchmarks)
+- [Additional Resources](#additional-resources)
+- [Integration in Transformers & how to use](#integration-in-transformers--how-to-use)
+- [Conclusion](#conclusion)
+- [Additional Resources](#additional-resources)
+
+## What is BitNet?
+
+BitNet replaces traditional Linear layers in Multi-Head Attention and Feed-Forward Networks with specialized layers called BitLinear with ternary precision. The main obstacle to training in ternary precision is that the weight values are discretized (via the `round()` function) and thus non-differentiable. BitLinear solves this with a nice trick: [STE (Straight Through Estimator)](https://arxiv.org/abs/1903.05662). The STE allows gradients to flow through the non-differentiable rounding operation by approximating its gradient as 1 (treating `round()` as equivalent to the identity function). Another way to view it is that, instead of stopping the gradient at the rounding step, the STE lets the gradient pass through as if the rounding never occurred, enabling weight updates using standard gradient-based optimization techniques.
```
SunMarc, line 33: I think it's better to talk about how STE overcomes the gradient issue after explaining how this layer do.

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,576 @@
...
+
+The BitLinear layers quantize the weights using ternary precision (with values of -1, 0, and 1) and quantize the activations to 8-bit precision. After performing the matrix multiplication operation during inference, the data is dequantized, allowing the model to continue with high-precision operations.
```
SunMarc, line 35: Can you give more information about the qunatization (per-token, per-channel, sym/asym) ? It will help users to google search that if they don't totally understand !

1.58llm_extreme_quantization.md
```diff
@@ -0,0 +1,576 @@
...
+
+The BitLinear layers quantize the weights using ternary precision (with values of -1, 0, and 1) and quantize the activations to 8-bit precision. After performing the matrix multiplication operation during inference, the data is dequantized, allowing the model to continue with high-precision operations. 
+
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/bitlinear.png" alt="Alt Text" />
+  <figcaption>The architecture of BitNet with BitLinear layers</figcaption>
+</figure>
+
+In the BitNet layer, during training, we start by making the weights ternary. We first compute the average of the weight matrix absolute value and use this value as a scale, then divide the weights by the scale, round the values, keep them between -1 and 1, and then rescale them
```
SunMarc, line 42: I think it can be nice to add an real example of how you quantize let's say a 3x3 matrix.
```suggestion
In the BitNet layer, during training, we start by quantizing the weights. We first compute the average of the weight matrix absolute value and use this value as a scale, then divide the weights by the scale, round the values, keep them between -1 and 1, and then rescale them 
```

SunMarc, line 42: To dissociate the training and inference part clearly, I think it can be nice to create subsection as the behavior changes a bit. The Bitlinear code is not the same depending on if we do training or inference.

SunMarc (commented):
Did a small review for now ! I'll do the rest tmr ! Thanks for the blogpost @MekkCyber

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,802 @@
...
+
+## Results In the Paper
+
+### Computational Efficiency :
+
+For BitNet, the energy consumption during matrix multiplication is primarily influenced by the addition operations, given that the weights are ternary. The multiplication operations are limited to scaling the output with the weights scale and activations scale, which provides enormous energy and computation savings.
+
+### Downstream Tasks
+
+BitNet is effective in delivering strong performance compared to baseline methods, especially at lower bit levels. According to the paper, BitNet achieves scores that are on par with 8-bit models but with significantly lower inference costs. In the case of 4-bit models, methods that only quantize weights outperform those that quantize both weights and activations, as activations are harder to quantify. However, BitNet, which uses 1.58-bit weights, surpasses both weight-only and weight-and-activation quantization methods.
```
SunMarc, line 349: We are not really discussion the results in the paper. I would just talk about the computational efficiency part in the section where you talk/code the kernel. 
Same for the downstream tasks, we can use that as an introduction when you share the results you got.

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,802 @@
...
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/exp-randoms.png" alt="Alt Text" style="width: 80%;"/>
+  <figcaption>Different Pre-training plots with different learning rates</figcaption>
+</figure>
+
+None of the models trained from random weights performed better than our fine-tuned model. The best perplexity we achieved with those models was 26, which falls short compared to the results from our fine-tuning approach.
+
+### Scaling to 100B !
+
+We scaled our experiments to 100 billion tokens to see if we could match the performance of Llama3 8B. We conducted longer training runs, starting from our best-performing checkpoint from the shorter runs with the linear scheduler, and continued fine-tuning for 45,000 steps. We experimented with different learning rates, and while the model performed closely to the Llama3 model in some metrics, on average, it still lagged behind.
```
SunMarc, line 548: Which model are we fine-tuning (exact name and size) ?

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,802 @@
...
+For example, here are the loss curves for the [Smol LLM 135M](https://huggingface.co/HuggingFaceTB/SmolLM-135M) model, comparing warmup quantization with full quantization from the start. Interestingly, the curves closely align, and the resulting perplexities aren't significantly different.
+
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/smol_llm_exp.png" alt="Alt Text" style="width: 60%;"/>
+  <figcaption>Smoll LLm finetuning experiment with & without warmup quantization</figcaption>
+</figure>
+
+### Experiments On a 70B Model
+
+Noticing that the metrics in our 100 billion token run started to stagnate, we realized that trying to match the performance of a Llama3 8B model using ternary weights might be unrealistic, especially given the reduced compute available for fine-tuning. Instead, we decided to apply our method to fine-tune the [Llama3.1 70B](https://huggingface.co/meta-llama/Meta-Llama-3.1-70B) model and compare its performance on benchmarks against the Llama3 8B model.
```
SunMarc, line 576: I would add the graph from the original paper to better explain why we decided to compare to 7B models: https://arxiv.org/pdf/2402.17764 . There are also other nice graphs that can be reused (energy consumption graph) for this blogpost.

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,802 @@
...
+
+For the 100B tokens experiments, the best performing checkpoint we had is the following : 
+
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/metrics_100B_table.png" alt="Alt Text" style="width: 60%;"/>
+  <figcaption>Metrics comparison with Llama models for the model trained on 100B tokens</figcaption>
+</figure>
+
+For the 70B model
+
```
SunMarc, line 599: ?

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,802 @@
...
+</figure>
+
+After fine-tuning on just 10 billion tokens using ternary weights, the model demonstrates impressive performance, especially when compared to other models that underwent much more extensive training. For instance, it outperforms the Bitnet 7B model, which was trained on a significantly larger dataset of 100 billion tokens. Additionally, it performs better than the FBI LLM (Fully Binarized LLM), a model that was distilled on an even more massive 1.26 trillion tokens. This highlights the model's efficiency and effectiveness despite the relatively smaller scale of its fine-tuning process.
+
+For the 100B tokens experiments, the best performing checkpoint we had is the following : 
+
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/metrics_100B_table.png" alt="Alt Text" style="width: 60%;"/>
+  <figcaption>Metrics comparison with Llama models for the model trained on 100B tokens</figcaption>
+</figure>
```
SunMarc, line 596: Instead of our model, we can maybe give a better description ? Also the (100B) or (10B) can be a bit confusing. Maybe add a token columns.

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,802 @@
...
+
+input_text = ""
+input_ids = tokenizer.encode(input_text, return_tensors="pt").cuda()
+output = model_.generate(input_ids, max_length=100, do_sample=False)
+generated_text = tokenizer.decode(output[0], skip_special_tokens=True)
+print(generated_text)
+```
+
+With this code, everything is managed seamlessly behind the scenes, so there's no need to worry about additional complexities. 
+
```
SunMarc, line 789: Maybe let's put this section a bit earlier (after what's bitnet) ?

SunMarc (commented):
Rest of the review, thanks for your work !

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+<div style="display: flex; justify-content: center;">
+  <figure style="margin-right: 20px; text-align: center;">
+    <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/poids_aléatoires.png" alt="Image 1" style="width: 400px;" />
+    <figcaption>Random weights distribution (2 merged stds)</figcaption>
+  </figure>
+  <figure style="text-align: center;">
+    <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/poids_llama3.png" alt="Image 2" style="width: 400px;" />
+    <figcaption>Pre-trained Llama3 weights distribution</figcaption>
+  </figure>
+</div>
```
pcuenca, line 400: Small nit: also use the caption in the `alt` text.

pcuenca: The axes and the titles are hard to read. In addition, would it help to use the same scales, or show them together?

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+<div style="display: flex; justify-content: center;">
+  <figure style="margin-right: 20px; text-align: center;">
+    <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/scales_random.png" alt="Image 1" style="width: 400px;" />
+    <figcaption>Random weights scales distribution</figcaption>
+  </figure>
+  <figure style="text-align: center;">
+    <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/scales_llama3.png" alt="Image 2" style="width: 400px;" />
+    <figcaption>Pre-trained Llama3 weights distribution</figcaption>
+  </figure>
+</div>
```
pcuenca, line 413: Same: would it help to use the same scales?

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+
+The initial random weight distribution is a mix of two normal distributions:
+
+- One with a standard deviation (std) of $0.025$
+- Another with a std of $ \frac{0.025}{\sqrt{2 \cdot \text{num\_hidden\_layers}}} = 0.00325 $
```
pcuenca, line 418: Is this a standard result from weights initialization, or nanotron specific?

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+
+This results from using different stds for column linear and row linear weights in `nanotron`. In the quantized version, all matrices have only 2 weight scales (50.25 and 402), which are the inverse of the mean absolute value of the weights for each matrix: `scale = 1.0 / w.abs().mean().clamp_(min=1e-5)`
+
+- For $\text{scale} = 50.25$, $w.abs().mean() = 0.0199$, leading to $\text{std} = 0.025$ which matches our first standard deviation. The formula used to derive the std is based on the expectation of the half-normal distribution of $ |w| $:
```
pcuenca, line 422: Latex fixes required here as well.

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,802 @@
...
+
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/pretraining.png" alt="Alt Text" style="width: 80%;"/>
+  <figcaption>Pretraining plots without (blue) & with (green) layer normalisation </figcaption>
+</figure>
+
+While this approach looks very interesting for pretraining, only a few institutions can afford pretraining models at the necessary scale. However, there is already a wide range of strong pretrained models, and it would be extremely useful if they could be converted to 1.58bit after pretraining. Thus we set out on an investigation to see if we can make 1.58 fine-tuning work. 
+
+## Finetuning in 1.58bit
```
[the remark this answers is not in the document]
pcuenca: Also, we should be consistent in the use of `finetuning` vs `fine-tuning` (I think the latter is used most times in the post, but not always).

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+
+On the other hand, the pretrained weight's distribution looks like a normal distribution with an $\text{std} = 0.013$
+
+Clearly, the pretrained model starts with more information (scales), while the randomly initialized model starts with practically no information and adds to it over time. Our conclusion was that starting with random weights gives the model minimal initial information, enabling a gradual learning process, while during finetuning, the introduction of BitLinear layers overwhelms the model into losing all its prior information.
+
+To improve the results of the finetuning, we tried different techniques, like instead of using per tensor quantization, we tried per row and per column quantization to keep more information from the weights of Llama3. We also tried to change the way how the scale is computed, by instead of just taking the mean absolute value of the weights as a scale, we take the mean absolute value of the outliers as a scale (an outlier value is a value that exceeds k*mean_absolute_value where k is a constant we tried to vary in our experiments), but we didn’t notice any big improvement.
```
pcuenca, line 430:
```suggestion
To improve the fine-tuning results, we tried different techniques. For example, instead of using per-tensor quantization, we tried per-row and per-column quantization to keep more information from the Llama 3 weights. We also tried to change the way the scale is computed: instead of just taking the mean absolute value of the weights as a scale, we take the mean absolute value of the outliers as a scale (an outlier value is a value that exceeds k*mean_absolute_value, where k is a constant we tried to vary in our experiments), but we didn’t notice big improvements.
```

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+    mean_outlier_value = torch.mean(torch.abs(outliers))
+    return mean_outlier_value
+
+def weight_quant_scaling(w):
+    scale = 1.0 / scale_outliers(w).clamp_(min=1e-5)
+    quantized_weights = (w * scale).round().clamp_(-1, 1) / scale
+    return quantized_weights
+```
+
+We observed that both the random weights and Llama3 weights resulted in losses starting at approximately the same value of 13. This suggests that the Llama3 model loses all of its prior information when quantization is introduced. To further investigate how much information the model loses during this process, we experimented with per-group quantization.
```
pcuenca, line 446:
```suggestion
We observed that both the random weights and the Llama 3 weights resulted in losses starting at approximately the same value of 13. This suggests that the Llama 3 model loses all of its prior information when quantization is introduced. To further investigate how much information the model loses during this process, we experimented with per-group quantization.
```

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+
+As a sanity check, we first set the group size to 1, which essentially means no quantization. In this scenario, the loss started at 1.45, as expected. However, when we increased the group size to 2, the loss jumped to around 11. This indicates that even with a minimal group size of 2, the model still loses nearly all of its information.
```
pcuenca, line 448:
```suggestion
As a sanity check, we first set the group size to 1, which essentially means no quantization. In this scenario, the loss started at 1.45, same as we see during normal fine-tuning. However, when we increased the group size to 2, the loss jumped to around 11. This indicates that even with a minimal group size of 2, the model still loses nearly all of its information.
```

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+
+```python
+lambda_ = ?
+x_quant = x + lambda_ * (activation_quant(x) - x).detach()
+w_quant = w + lambda_ * (weight_quant(w) - w).detach()
+```
+
+When `lambda` is set to 0, there is essentially no quantization occurring, while at `lambda=1`, full quantization is applied.
+
+We initially tested some discrete `lambda` values, such as 0.25, 0.5, 0.75, and 1. However, this approach did not lead to any significant improvement in results mainly because `lambda=0.25` is already high enough for the loss to start very high.
```
pcuenca, line 460:
```suggestion
We initially tested some discrete `lambda` values, such as 0.25, 0.5, 0.75, and 1. However, this approach did not lead to any significant improvement in results, mainly because `lambda=0.25` is already high enough for the loss to start very high.
```

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+  <figcaption>Finetuning plot with lambda = 0.25->0.5->0.75->1</figcaption>
+</figure>
+
+As a result, we decided to experiment with a `lambda` value that adjusts dynamically based on the training step.
+
+```python
+lambda_ = training_step / total_training_steps
+```
+
+Using this dynamic `lambda` value led to better loss convergence, but the perplexity (ppl) results during inference, when `lambda` was set to 1, were still far from satisfactory. We realized this was likely because the model hadn't been trained long enough with `lambda`=1. To address this, we adjusted our `lambda` value to improve the training process.
```
pcuenca, line 473:
```suggestion
Using this dynamic `lambda` value led to better loss convergence, but the perplexity (ppl) results during inference, when `lambda` was set to 1, were still far from satisfactory. We realized this was likely because the model hadn't been trained long enough with `lambda=1`. To address this, we adjusted our `lambda` value to improve the training process.
```

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+```python
+lambda_ = min(2 * training_step / total_training_steps, 1)
+```
+
+With this configuration, after 2000 steps we have : 
+
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/lambda_training_step.png" alt="Alt Text" style="width: 80%;"/>
+  <figcaption>Finetuning plot with lambda = min(2*training_step/total_training_steps, 1)</figcaption>
+</figure>
```
pcuenca, line 484: Legend suggestions:
- Pre-training from random weights
- Fine-tuning from Llama 3 (naive)
- Fine-tuning from Llama 3 (progressive lambda)

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+
+Our fine-tuning method shows better convergence overall. You can observe a slight increase in the loss curve around 1,000 steps, which corresponds to when we begin approaching `lambda=1`, or full quantization. However, immediately after this point, the loss starts to converge again, leading to an improved perplexity of approximately 4.
+
+Despite this progress, when we tested the quantized model, which was fine-tuned on the tinystories small dataset, using the WikiText dataset, the model showed a very high perplexity. This suggests that fine-tuning the model in low-bit mode on a specific dataset causes it to lose much of its general knowledge. This issue might arise because the minimal representations we aim for with ternary weights can vary significantly from one dataset to another. To address this problem, we scaled our training process to include the larger [FineWeb-edu](https://huggingface.co/datasets/HuggingFaceFW/fineweb) dataset. We maintained a `lambda` value of:
```
pcuenca, line 488:
```suggestion
Despite this progress, when we tested the quantized model on the WikiText dataset (instead of the tinystories one we used for fine-tuning), it showed a very high perplexity. This suggests that fine-tuning the model in low-bit mode on a specific dataset causes it to lose much of its general knowledge. This issue might arise because the minimal representations we aim for with ternary weights can vary significantly from one dataset to another. To address this problem, we scaled our training process to include the larger [FineWeb-edu](https://huggingface.co/datasets/HuggingFaceFW/fineweb) dataset. We maintained a `lambda` value of:
```

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+</figure>
+
+We ran 5 experiments this time with k in [15, 20, 25, 40, 100] : 
+
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/sig_scheduler_exps.png" alt="Alt Text" style="width: 80%;"/>
+  <figcaption>Finetuning plots with sigmoid scheduler</figcaption>
+</figure>
+
+The sharp increase in lambda caused instability around the 500th step and didn’t fix the first divergence issue. However, for \( k = 100 \), we observed some improvement in downstream tasks (see the results table), although perplexity remained around ~13.5. Despite this, it didn’t show a clear performance boost over a linear scheduler.
```
pcuenca, line 552:
```suggestion
The sharp increase in lambda caused instability around the 500th step and didn’t fix the first divergence issue. However, for \\( k = 100 \\), we observed some improvement in downstream tasks (see the results table), although perplexity remained around ~13.5. Despite this, it didn’t show a clear performance boost over a linear scheduler.
```

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/exp-randoms.png" alt="Alt Text" style="width: 80%;"/>
+  <figcaption>Different Pre-training plots with different learning rates</figcaption>
+</figure>
+
+None of the models trained from random weights performed better than our fine-tuned model. The best perplexity we achieved with those models was 26, which falls short compared to the results from our fine-tuning approach.
+
+### Scaling to 100B Tokens !
+
+We scaled our experiments to 100 billion tokens to see if we could match the performance of Llama3 8B. We conducted longer training runs, starting from our best-performing checkpoint from the shorter runs with the linear scheduler, and continued fine-tuning for 45,000 steps. We experimented with different learning rates, and while the model performed closely to the Llama3 model in some metrics, on average, it still lagged behind.
```
pcuenca, line 566:
```suggestion
We scaled our experiments to 100 billion tokens to see if we could match the performance of Llama 3 8B. We conducted longer training runs, starting from our best-performing checkpoint from the shorter runs with the linear scheduler, and continued fine-tuning for 45,000 steps. We experimented with different learning rates, and while the model performed closely to the Llama 3 model in some metrics, on average, it still lagged behind.
```

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+and the average score looks like : 
+
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/metric_avg.png" alt="Alt Text" style="width: 60%;"/>
+  <figcaption>Average evaluation during the training for different lrs</figcaption>
+</figure>
+
+### Experiments on Smaller Models
+
+In our initial experiments with smaller models like Smol LLMs, we observed that the warmup quantization technique didn’t yield as much improvement as it did with larger models. This suggests that the effectiveness of warmup quantization could be more closely related to model size and complexity.
```
pcuenca, line 583:
```suggestion
In our initial experiments with smaller models like SmolLM, we observed that the warmup quantization technique didn’t yield as much improvement as it did with larger models. This suggests that the effectiveness of warmup quantization could be more closely related to model size and complexity.
```

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+
+For example, here are the loss curves for the [Smol LLM 135M](https://huggingface.co/HuggingFaceTB/SmolLM-135M) model, comparing warmup quantization with full quantization from the start. Interestingly, the curves closely align, and the resulting perplexities aren't significantly different.
```
pcuenca, line 585:
```suggestion
For example, here are the loss curves for the [SmolLM 135M](https://huggingface.co/HuggingFaceTB/SmolLM-135M) model, comparing warmup quantization with full quantization from the start. Interestingly, the curves closely align, and the resulting perplexities aren't significantly different.
```

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/metrics_100B_table.png" alt="Alt Text" style="width: 60%;"/>
+  <figcaption>Metrics comparison with Llama models for the model trained on 100B tokens</figcaption>
+</figure>
+
+To replicate these results, you can check out this [PR](https://github.com/huggingface/nanotron/pull/174) to convert models to nanotron format, unpack the weights (check the function [unpack_weights](https://gist.github.com/MekkCyber/78c1532e8767e8da0588b778faf61866)), and use lighteval
+
+Note that even though the models are fine-tuned from an Instruct-tuned model, they still need to be fine-tuned using an Instruct dataset as well. These can be considered base models.
+
+## Kernels & Benchmarks
```
pcuenca, line 616:
```suggestion
## Custom Kernels & Benchmarks
```

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+
+To benefit from the BitNet low-precision weights, we pack them into an `int8` tensor (this makes the number of parameters go from 8B to 2.8B!). During inference, these weights must be unpacked before performing matrix multiplication. We implemented custom kernels in Cuda and Triton to handle the on-the-fly unpacking during the matrix multiplication process. For the matrix multiplication itself, we employed the cached tiled matrix multiplication technique. To fully grasp this approach, let's first review some Cuda programming fundamentals.
+
+### **Basic GPU Concepts: Threads, Blocks, and Shared Memory**
```
pcuenca, line 620:
```suggestion
### Basic GPU Concepts: Threads, Blocks, and Shared Memory
```

No need to bold the section heading imo.

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+
+Before diving into cached tiled matrix multiplication, it's important to understand some basic GPU concepts:
+
+- **Threads and Blocks**: GPUs execute thousands of threads simultaneously. These threads are grouped into blocks, and each block runs independently. The grid is made up of these blocks, and it represents the entire problem space. For example, in matrix multiplication, each thread might be responsible for computing a single element of the output matrix.
+- **Shared Memory**: Each block has access to a limited amount of shared memory, which is much faster than global memory (the main memory on the GPU). However, shared memory is limited in size and shared among all threads within a block. Using shared memory effectively is key to improving performance in GPU programs.
+
+### **Challenges in Matrix Multiplication**
```
pcuenca, line 627:
```suggestion
### Challenges in Matrix Multiplication
```

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+
+- **Memory Bandwidth**: Accessing global memory is relatively slow compared to the speed at which the GPU cores can perform computations. If each thread reads matrix elements directly from global memory, the memory access times can become a bottleneck.
+- **Redundant Data Access**: In matrix multiplication, many elements of the input matrices are used multiple times. If each thread fetches the required data from global memory independently, the same data might be loaded into the GPU multiple times, leading to inefficiency. For example, if each thread is used to compute a single element in the output matrix, the thread responsible for calculating the element at position (i, j) will need to load the i-th row of matrix A and the j-th column of matrix B from global memory. However, other threads, such as the one computing the element at position (i+1, j), cannot reuse this data and will have to reload the same j-th column from global memory again.
+
+### **The Idea of Tiling**
```
pcuenca, line 634:
```suggestion
### The Idea of Tiling
```

pcuenca (approved):
Very interesting! 🔥 I made a few suggestions as well as fixes for the latex blocks.

In terms of content, I wonder if we could maybe defer some of the quantization details (equations and examples) to a later point, so the reader is progressively presented with more details as they go deeper. So I'd maybe introduce what BitNet is (conceptually, without the equations), what we did (a short intro about what comes next: pre-training, fine-tuning, models we released), how to use those models in transformers, then go deep into each section.

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,799 @@
...
+  $$
+
+</details>
+
+<details>
+  <summary>Example 2: Activation Matrix Quantization</summary>
+
+  Let the activation matrix \( X \) be:
+
+  $$
```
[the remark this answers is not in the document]
MekkCyber: With the new syntax, equations aren't rendered correclty on vscode, is this normal ?

1_58_llm_extreme_quantization.md
MekkCyber: It's nanotron specific, but based on a paper

1_58_llm_extreme_quantization.md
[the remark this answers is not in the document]
pcuenca: I think so, the rendering engine for the blog uses a special format for latex.

1_58_llm_extreme_quantization.md
```diff
@@ -0,0 +1,807 @@
...
+<figure style="text-align: center;">
+  <img src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/1.58llm_extreme_quantization/with_bitblas.png" alt="Bitblas benchmark" style="width: 100%;"/>
+  <figcaption>Bitblas benchmark</figcaption>
+</figure>
+
+However, during model loading, BitBlas needs to compile kernels tailored to the shape of the weight matrix and store them in a local database, which can increase the initial loading time. 
+
+## Conclusion
+In conclusion, as (LLMs) continue to expand, reducing their computational demands through quantization is essential. This blog has explored the approach of 1.58-bit quantization, which uses ternary weights. While pre-training models in 1.58 bits is resource-intensive, we’ve demonstrated that, with some tricks, it’s possible to fine-tune existing models to this precision level, achieving efficient performance without sacrificing accuracy. By optimizing inference speed through specialized kernels, BitNet opens new possibilities for making LLMs more practical and scalable.
+
```
SunMarc, line 795:
```suggestion
In conclusion, as LLMs continue to expand, reducing their computational demands through quantization is essential. This blog has explored the approach of 1.58-bit quantization, which uses ternary weights. While pre-training models in 1.58 bits is resource-intensive, we’ve demonstrated that, with some tricks, it’s possible to fine-tune existing models to this precision level, achieving efficient performance without sacrificing accuracy. By optimizing inference speed through specialized kernels, BitNet opens new possibilities for making LLMs more practical and scalable.

```

SunMarc (approved):
Nice blog ! Thanks for iterating !

PR #2407, opened 2024-10-10
fixes: the video link of gradio5

chenglu:
Hi @abidlabs 

I've update the video link for the 2nd video demo of Gradio 5, since the existing link seems broken on the blog page.

cc/ @xianbaoqian @adeenayakup 

Thanks!
Luke


abidlabs (approved):
Thanks @chenglu, this makes sense to me, although both  videos are weirdly not showing up for me in my browser when I test the direct urls.

xianbaoqian:
Thx Luke! Checked hf.co/blog and the video works for me 👍 

@abidlabs I had a similar issue before and solved it by clearing caches / changing the browser :)

PR #2444, opened 2024-10-25
add intel-gcp-c4

yao-matrix:
@kding1 @IlyasMoutawwakil  @echarlaix , pls help review and comment anything, thx.


assets/intel-gcp-c4/text_generation_perf.png
IlyasMoutawwakil, line 1: love this chart, it shows how throughput increases without compromising as much latency.
maybe use the term and metric "decoding throughput" instead of "per-token throughput".

yao-matrix: > love this chart, it shows how throughput increases without compromising as much latency. maybe use the term and metric "decode throughput" instead of "per-token throughput".

seems in optimum-benchmark, `decode throughput` and `per-token throughput` are 2 different metrics, I understood in our e-mail that you may prefer `per-token throughput`, let me know if you prefer `decode throughput`, I'll change to it.

IlyasMoutawwakil: `decode-throughput` and `per-token-throughput` are the same metric but they might be slightly different numerically because they are measured differently (holistic vs fine-grained), for example 36 vs 37 in https://huggingface.co/datasets/optimum-benchmark/cpu/blob/main/cpu_inference_transformers_text-generation_openai-community/gpt2/benchmark_report.json
It's just a matter of wording, decoding throughput makes more sense for me than per-token throughput.

IlyasMoutawwakil:
@yao-matrix can provide some version/commit pins for reproduction ; the ipex backend was just refactored in optimum-intel and i'm not sure how that impacts the results.

yao-matrix:
> @yao-matrix can you provide some version/commit pins for reproduction ; the ipex backend was just refactored in optimum-intel and i'm not sure how that impacts the results.

done, thx. I put version and commit pinning in commands of set up env section, thx.

intel-gcp-c4.md
```diff
@@ -0,0 +1,147 @@
...
+
+Now, you have one C4 instance.
+
+## Set up environment
+Follow below steps to set up the environment easily. For reproducibility, we list the version and commit we are using in the commands.
+
+1. SSH connect to instance
+2. `$ pip install "optimum-intel[ipex]"@git+https://github.com/huggingface/optimum-intel.git@6a3b1ba5924b0b017b0b0f5de5b10adb77095b`
+3. `$ pip install torch==2.3.1 torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu`
+4. `$ python -m pip install intel-extension-for-pytorch==2.3.10`
```
IlyasMoutawwakil, line 63: shouldn't this come after running the container ?

intel-gcp-c4.md
```diff
@@ -55,19 +55,27 @@ Follow the below steps to create a 96-vcpu instance which corresponds to one Int
...
+5. `$ git clone https://github.com/huggingface/optimum-benchmark.git`
+6. `$ cd ./optimum-benchmark`
+7. `$ git checkout d58bb2582b872c25ab476fece19d4fa78e190673`
+8. `$ cd ./docker/cpu`
+9. `$ sudo docker build . -t <your_docker_image_tag>`
+10. `$ sudo docker run -it --rm --privileged -v /home/<your_home_folder>:/workspace <your_docker_image_tag> /bin/bash`
+
+We are in container now, do following steps:
+
+1. `$ pip install huggingface-hub`
```
IlyasMoutawwakil, line 73: a login is required after this step (with a token that has gated llama repos permissions)

yao-matrix: done

intel-gcp-c4.md
yao-matrix: yes, updated

IlyasMoutawwakil (approved):
Great work !

PR #2484, opened 2024-11-23
[annotated-diffusion] fix typos

kashif:
Fix issue in normalization layer and some typoes.


annotated-diffusion.md
```diff
@@ -240,7 +242,7 @@ class WeightStandardizedConv2d(nn.Conv2d):
         weight = self.weight
         mean = reduce(weight, "o ... -> o 1 1 1", "mean")
         var = reduce(weight, "o ... -> o 1 1 1", partial(torch.var, unbiased=False))
-        normalized_weight = (weight - mean) / (var + eps).rsqrt()
+        normalized_weight = (weight - mean) * (var + eps).rsqrt()
```
pcuenca, line 245: I must surely be wrong, but not following this right now. For example, in [this implementation](https://github.com/huggingface/transformers/blob/3a8eb74668e9c2cc563b2f5c62fac174797063e0/src/transformers/models/bit/modeling_bit.py#L140) it's implemented with batch_norm, which is defined like https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm2d.html#torch.nn.BatchNorm2d. Also, eq. 10 of https://arxiv.org/pdf/1903.10520 also seems to use the division.

pcuenca (approved):
Thanks for the fixes! I have a question about the weight standardization, but it's probably a confusion on my side.

annotated-diffusion.md
kashif: no its confusing! it because of `rsqrt` = `1/sqrt` and so it is dividing now while previously it was multiplying

NielsRogge (approved):
Thanks for fixing!

PR #2501, opened 2024-11-28
[FIX] Resolves compilation error in demo.c

SureD:
Add missing header for ggml_graph_compute_with_ctx function

the error :
`demo/demo.c:59:5: error: call to undeclared function 'ggml_graph_compute_with_ctx'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]`


pcuenca:
Thanks @ngxson!

PR #2525, opened 2024-12-10
[paligemma2] Fix output decoding, install from PyPi

pcuenca:


MoritzLaurer (approved):
lgtm. Could maybe add a brief half-sentence somewhere mentioning that there must not be a space after the <image> token, otherwise the model breaks down.

PR #2534, opened 2024-12-16
[Update][Dask] Use `df.to_parquet` instead of custom function

lhoestq:
...and suggest `super_squash_history` to not end up with too many commits.

Doing this change because concurrent commits work well now, no need for the custom function with preupload !

Cc @scharlottej13 for viz :)

I'll also update the HF docs on `dask`

### DIFF
--- a/dask-scaling.md
+++ b/dask-scaling.md
@@ -129,18 +129,24 @@ df = df[df["edu-classifier-score"] >= min_edu_score]
 
 Note that we’ve picked a `batch_size` that works well for this example, but you’ll likely want to customize this depending on the hardware, data, and model you’re using in your own workflows (see the [HF docs on pipeline batching](https://huggingface.co/docs/transformers/en/main_classes/pipelines#pipeline-batching)).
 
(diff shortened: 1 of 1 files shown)


dask-scaling.md
```diff
@@ -129,18 +129,24 @@ df = df[df["edu-classifier-score"] >= min_edu_score]
...
+Since this creates one commit per file, it is recommended to squash the history after the upload:
+
+```python
+from huggingface_hub import HfApi
+
+HfApi().super_squash_history(repo_id=repo_id, repo_type="dataset")
 ```
 
-In the future, we’ll create a direct integration with Dask and `huggingface_hub`, but for now, you can copy + paste [this custom function](https://gist.github.com/lhoestq/8f73187a4e4b97b9bb40b561e35f6ccb) for your own use. In this example, we’ve saved it to a file called `dask_hf.py`, which is the file referenced in the `import` clause above.
+Alternatively can use [this custom function](https://gist.github.com/lhoestq/8f73187a4e4b97b9bb40b561e35f6ccb) which uploads multiple files per commit.
```
scharlottej13, line 149:
```suggestion
Alternatively you can use [this custom function](https://gist.github.com/lhoestq/8f73187a4e4b97b9bb40b561e35f6ccb) which uploads multiple files per commit.
```

davanstrien (approved):
LGTM

PR #2538, opened 2024-12-17
Fix yaml

pcuenca:
Also, @ariG23498 is not in the list of authors but appears as the author in `_blog.yml`, it's ok that it's a guest author.


pcuenca:
#2537

PR #2539, opened 2024-12-17
Update falcon3.md

younesbelkada:
# What does this PR do ?

Fixes an author name for Falcon3 blogpost

### DIFF
--- a/falcon3.md
+++ b/falcon3.md
@@ -47,7 +47,7 @@ authors:
 - karnakar
   guest: true
   org: tiiuae
-- kpiskorski
+- kasper-piskorski
   guest: true
   org: tiiuae
 - melaseddik


falcon3.md
```diff
@@ -47,7 +47,7 @@ authors:
 - user: karnakar
   guest: true
   org: tiiuae
-- user: kpiskorski
+- kasper-piskorski
```
ariG23498, line 50:
```suggestion
- user: kasper-piskorski
```

ariG23498: @Vaibhavs10 I think we are missing the `user` parameter here.

ariG23498 (changes requested):
Could we please update using the `user:` parameter?

PR #2558, opened 2024-12-20
Update Falcon blogpost

younesbelkada:
cc @Vaibhavs10 and @ariG23498 !


Vaibhavs10 (approved):
There goes the HF record of most number of authors haha.

PR #2573, opened 2025-01-02
[smolagents] Typos and minor nits.

pcuenca:
A couple of additional comments:

- I suppose `HfApiModel` supports arbitrary models and endpoints and will use the user's rate limits if they are authenticated. Would it be useful to write a small comment about that?
- Why does `additional_authorized_imports` only list `datetime` (and not `googlemaps`, which is also used in the implementation?)
- The exported Space does not work because the environment variable with the google clouds api key is not defined. Perhaps we could add a note about it in the Space (and/or the blog), or recommend users to duplicate and provide their own.

### DIFF
--- a/smolagents.md
+++ b/smolagents.md
@@ -134,14 +134,14 @@ We built [`smolagents`](https://github.com/huggingface/smolagents) with these ob
 
 To build an agent, you need at least two elements:
 
-- `tools`: a list of  that the agent has access to
-- `model`: a LLM that will be the engine of your agent.
+- `tools`: a list of tools the agent has access to
+- `model`: an LLM that will be the engine of your agent.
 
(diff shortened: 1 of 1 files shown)


aymeric-roucher:
Good points @pcuenca , thank you for reviewing this.
Indeed adding your first point about HfApiModel would be nice!
Re: imports, since the import is used only in the tool, it will be invisible to the model: `additional_authorized_imports` is meant to authorize using some imports **directly in the code blob that the agent writes**: maybe I should clarify that in the doc.
And re:space, I was thinking of modifying my space to fix this issue by adding my API key as a secret, and didn't have time: doing it right now.

aymeric-roucher (approved):

aymeric-roucher:
Merging to add you proposed changes directly while the blog post is hot but we can open other PRs!

PR #2574, opened 2025-01-02
Update smolagents.md [Fix code typo]

KshitijKarthick:
### Why?
In code block when departure_time was set as None, rather than set departure_time, a new unused variable was set called monday.

### Solution
Set the default value as departure_time rather than monday

```python
if departure_time is None:
   from datetime import datetime

   # Unused variable, ends up calling gmaps with None, when departure_time is None
   # Set the default datetime to departure_time instead
   # Replace below line with departure_time = datetime(2025, 1, 6, 11, 0)
   monday = datetime(2025, 1, 6, 11, 0)

directions_result = gmaps.directions(
    start_location,
    destination_location,
    mode="transit",
    departure_time=departure_time
)
```


KshitijKarthick:
@pcuenca Would love to get a review, found a small nitpick while reading the blog

pcuenca (approved):
Looks good to me! cc @aymeric-roucher for confirmation.

aymeric-roucher (approved):

aymeric-roucher:
Yes, this seems fine! Thank you for spotting this @KshitijKarthick!

KshitijKarthick:
@aymeric-roucher  @pcuenca  Thank you for your work and support :)

PR #2581, opened 2025-01-09
Smolagents blog: Improve get-travel-duration-tool

aymeric-roucher:


aymeric-roucher:
@pcuenca this makes the tool shown in blog post more reliable!

aymeric-roucher:
@pcuenca this is ready for review now! Also I could ask someone else if you're taken.

pcuenca (approved):
Nice! Are the results in the example call (just after the tool code is introduced) relying on driving directions?

aymeric-roucher:
yes! I updated these for bike direction by the way to make them more time-independant.

aymeric-roucher:
Thank you for the review @pcuenca !

PR #2583, opened 2025-01-09
Update big-bench-audio-release.md

CharlesCNorton:
fix: correct typo "Artifical" → "Artificial" in speech arena reference

- Replaces the incorrect "Artifical Analysis Speech Arena" with "Artificial Analysis Speech Arena"


pcuenca (approved):
Thanks again!

PR #2586, opened 2025-01-11
Update vdr-2b-multilingual.md

CharlesCNorton:
fix: correct "scrapped" → "scraped" and "it’s kind" → "its kind"

- Replace "scrapped" with "scraped" when referring to data gathering
- Replace "it’s kind" with "its kind" for correct possessive usage


pcuenca (approved):
Thank you!

PR #2636, opened 2025-01-31
fix(minor): broken link

LPX55:
Pretty sure the link was meant to go to the model page, as there is no blog post titled remotely similar to the current link.

Congrats on the integrations!


pcuenca (approved):
lol, yes

PR #2660, opened 2025-02-10
Fixed link to video processing pipeline code

bharathsivaram10:
Fixed link to video processing pipeline code

### DIFF
--- a/smolvlm.md
+++ b/smolvlm.md
@@ -146,7 +146,7 @@ SmolVLM's tiny memory footprint also implies that it requires far fewer computat
 
 Given SmolVLM's long context and the possibility of tweaking the internal frame resizing of the model, we explored its suitability as an accessible option for basic video analysis tasks, particularly when computational resources are limited.
 
-In our evaluation of SmolVLM's video understanding capabilities, we implemented a straightforward [video processing pipeline code](https://github.com/huggingface/smollm/blob/main/tools/smolvlm_local_inference/SmolVLM_video_inference.py), extracting up to 50 evenly sampled frames from each video while avoiding internal frame resizing.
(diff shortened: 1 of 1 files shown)


smolvlm.md
```diff
@@ -146,7 +146,7 @@ SmolVLM's tiny memory footprint also implies that it requires far fewer computat
 
 Given SmolVLM's long context and the possibility of tweaking the internal frame resizing of the model, we explored its suitability as an accessible option for basic video analysis tasks, particularly when computational resources are limited.
 
-In our evaluation of SmolVLM's video understanding capabilities, we implemented a straightforward [video processing pipeline code](https://github.com/huggingface/smollm/blob/main/inference/smolvlm/SmolVLM_video_inference.py), extracting up to 50 evenly sampled frames from each video while avoiding internal frame resizing.
+In our evaluation of SmolVLM's video understanding capabilities, we implemented a straightforward [video processing pipeline code](https://github.com/huggingface/smollm/blob/main/tools/smolvlm_local_inference/SmolVLM_video_inference.py), extracting up to 50 evenly sampled frames from each video while avoiding internal frame resizing.
```
pcuenca, line 149: Thank you! Can we use a permalink instead?

```suggestion
In our evaluation of SmolVLM's video understanding capabilities, we implemented a straightforward [video processing pipeline code](https://github.com/huggingface/smollm/blob/7dfcd81c046a946031291c11451e9398609a0aeb/tools/smolvlm_local_inference/SmolVLM_video_inference.py), extracting up to 50 evenly sampled frames from each video while avoiding internal frame resizing.
```

bharathsivaram10: Yes that would work too although, is there a reason we want to reference the non-main branch if files exists on main? This is my first PR so just trying to understand :)

pcuenca: The main advantage is that the link will still work if the files are moved, and it represents the state of affairs when the post was written. But we can use `main` if you think it's better to always refer to the latest.

bharathsivaram10: I think your reasoning makes sense! Will go ahead and do peramlink.

pcuenca (approved):
Thank you!

PR #2669, opened 2025-02-12
fix: Correct typos vid_ds_scripts.md

CharlesCNorton:
- Fixed the phrase 'In our we filtering we required all scores to be below the threshold…' by removing the extra 'we'. This correction clarifies the sentence and ensures proper grammar.

- Fixed the phrase 'for each filters as well as the captions from Florence-2' by changing 'filters' to 'filter'. 'Each' should be followed by a singular noun, which improves the accuracy of the text.


hlky (approved):

sayakpaul:
Thank you!

PR #2703, opened 2025-02-25
update installation instruction

ariG23498:
CC: @sayakpaul


sayakpaul (approved):
Thanks! @hlky we can confirm that with `diffusers==0.32.3` it won't work because https://github.com/huggingface/diffusers/pull/10441 isn't released yet.

PR #2714, opened 2025-02-28
Update Remote VAE blog

hlky:
Depends on https://github.com/huggingface/diffusers/pull/10898

Depends on endpoint containers being updated https://endpoints.huggingface.co/diffusers/endpoints/dedicated

### DIFF
--- a/remote_vae.md
+++ b/remote_vae.md
@@ -16,6 +16,8 @@ Therefore, we want to pilot an idea with the community — delegating the decodi
 
 No data is stored or tracked, and code is open source. We made some changes to [huggingface-inference-toolkit](https://github.com/hlky/huggingface-inference-toolkit/tree/fix-text-support-binary) and use [custom handlers](https://huggingface.co/hlky/sd-vae-ft-mse/blob/main/handler.py).
 
+This experimental feature is developed by [Diffusers 🧨](https://huggingface.co/docs/diffusers/hybrid_inference/overview)
+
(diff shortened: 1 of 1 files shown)


remote_vae.md
```diff
@@ -246,70 +122,6 @@ with open("video.mp4", "wb") as f:
   </video>
 </figure>
 
-### Options
```
hlky, line 249: Confusing for the average reader imo. Developers can check the docstrings.

sayakpaul: I think it's better to keep it. We can add a note "Users are not required to go through this" or something like that.

remote_vae.md
```diff
@@ -337,6 +149,7 @@ latent = pipe(
 image = remote_decode(
     endpoint="https://q1bj3bpq6kzilnsu.us-east-1.aws.endpoints.huggingface.cloud/",
     tensor=latent,
+    scaling_factor=0.18215,
```
sayakpaul, line 152: We should discuss how users can get this value I think. 

Something like:

```py
from huggingface_hub import hf_hub_download

file_path = hf_hub_download(repo_id, filename="vae/config.json")

...
```

sayakpaul (approved):
Left some comments.

remote_vae.md
hlky: It was also outdated, will add it back based on current docstrings.

PR #2774, opened 2025-03-27
Update train-reranker.md

CharlesCNorton:
- Correct plural form: "CrossEncoder are" → "CrossEncoders are"

- Fix passive phrasing: "batches are samples" → "batches are sampled"


tomaarsen (approved):
Thank you for this! I completely missed these issues.

PR #2816, opened 2025-04-16
refactor: resize protect ai 6 months headers

McPatate:
Resizing the headers as they are a bit big in the current blog post

cc @willarmiros @seanpmorgan


willarmiros (approved):
Thanks for catching this!

PR #2821, opened 2025-04-18
notify that cuda is not the only hardware choice

faaany:
This notebook assumes users to have NV GPU by default. But this is not a must. This notebook also works on other hardware accelerators, e.g. Intel GPU, TPU, because accelerator would detect the underly hardware accelerators. This PR add a note to make it clear. 

cc @pcuenca


pcuenca (approved):
Thank you.

PR #2830, opened 2025-04-26
fix: correct indefinite article before initialisms in tinyagents.md

CharlesCNorton:
Correct all occurrences of the mis-phrased “a MCP …” and “a LLM …”:

* Title → “an MCP-powered agent”
* Intro (Inference Client section) → “an MCP client”
* TIP block → “an MCP client”
* Quote following TIP → “an MCP client”
* Implementation section (“Now that we have…”) → “an MCP client”
* Agent component list → “an LLM inference client”


abidlabs (approved):
Thanks!

julien-c (commented):
is this actually correct? "An MCP server" sound weird to me but idk 🤷‍♂️

PR #2837, opened 2025-04-30
Update llama-guard-4.md

CharlesCNorton:
- Change “TLDR;” to “TL;DR:” in the frontmatter

- Add missing period after “filtered out by production services”

- Update “24 GBs of VRAM” to “24 GB of VRAM”

- Correct hazard count from 13 to 14 types to match the listed categories


abidlabs (approved):
Nice thanks @CharlesCNorton!

PR #2853, opened 2025-05-14
Update vlms-2025.md

CharlesCNorton:
- Change "they're earliest attempts" to "their earliest attempts" to use the correct possessive pronoun.

- Change "Here's our picks" to "Here are our picks" to ensure subject-verb agreement.


sergiopaniego (approved):
lgtm! thanks!

PR #2858, opened 2025-05-15
Update gradio-mcp.md to include example with `stdio` protocol

abidlabs:


yvrjsharma:
lgtm too! Thanks for making the updates.

abidlabs:
Thanks guys for reviewing

PR #2884, opened 2025-06-03
[Add] KV Cache blog post for nanoVLM

ariG23498:
Adding the blog post on KV Cache for nanoVLM


_blog.yml
```diff
@@ -6076,14 +6076,26 @@
     - llm
 
 - local: structured-codeagent
-  title: "CodeAgents + Structure: A Better Way to Execute Actions"
+  title: "CodeAgents + Structure: A Better Way to Execute Actions"
   author: akseljoonas
   thumbnail: /blog/assets/structured-codeagent/thumbnail-codeagent.png
-  date: May 28, 2024
+  date: May 28, 2025
```
pcuenca, line 6082: lol, nice

kv-cache.md
```diff
@@ -0,0 +1,269 @@
+---
+title: "KV Cache from scratch in nanoVLM
+thumbnail: /blog/assets/kv_cache_quantization/thumbnail.png
```
pcuenca:
```suggestion
thumbnail: /blog/assets/kv-cache/thumbnail.png
```

kv-cache.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Introduction
+
+Autoregressive language models generate text by sampling *one token at a time*. During inference, the model processes a given input sequence, predicts the next token, appends it to the input, and repeats this process until some stopping criterion:
+
+![diagram for autoregression](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/kv-cache/autoregression.png)
+
+This step-by-step generation is inherently sequential:
+
+- To generate token \\( t_{i+1} \\), the model must consider the entire sequence from \\( t_0 \\) to \\( t_i \\). From the first instance in the above example \\( t_{i+1} \\) would be `the` , while all the previous tokens \\( t_0 \\) to \\( t_i \\) would be `[What, is, in,]`.
```
lusxvr, line 28: The math is not show correctly for me when I look at the rich-diff
image

lusxvr: Same for almost all the other formulas

ariG23498: If you go to [new blog](hf.co/new-blog) and paste the math there, it should render correctly.
The idea here is that math rendering for HF blogs is a little different to the latex rendering.

kv-cache.md
```diff
@@ -0,0 +1,269 @@
...
+K_ext = extended_input @ W_k
+V_ext = extended_input @ W_v
+
+# (output_ext would be computed using Q_ext, K_ext, V_ext + masking)
+```
+
+To confirm the redundancy:
+
+```python
+torch.testing.assert_close(K, K_ext[:input_seq_length])
```
lusxvr: maybe put a ` -> returns true` here

kv-cache.md
```diff
@@ -0,0 +1,269 @@
...
+torch.testing.assert_close(V, V_ext[:input_seq_length])
+```
+
+These checks show that for all but the newest token, \\( K \\) and \\( V \\) are identical to previously computed values.
+
+```
+Original (5×5):         Extended (6×6):
```
lusxvr, line 124: Really love this graphic!

kv-cache.md
```diff
@@ -0,0 +1,269 @@
...
+    decode_output, kv_cache_list = self.forward(
+        next_token,
+        kv_cache=kv_cache_list,
+        start_pos=current_position  # updated with each step
+    )
+
+    prompt_output = decode_output
+```
+
+> By separating these phases, we avoid redundant computation and dramatically speed up inference, especially for long prompts.
```
lusxvr, line 250: Maybe we could put an additional table here that shows the speedup wit kv cache for a generation with 1, 10, 100, 1000, 10000 tokens. This would hammer home the point that it really helps the longer the sequence gets but also doesn't degrade performance on short prompts

kv-cache.md
```diff
@@ -0,0 +1,269 @@
...
+- user: lusxvr
+- user: andito
+- user: pcuenq
+---
+
+# KV Cache from scratch in nanoVLM
+
+## TL;DR
+
+We have implemented KV Caching from scratch in our [nanoVLM](https://github.com/huggingface/nanoVLM) repository (a small code base to train your own Vision Language Model with pure PyTorch). This gave us a of **38%** speedup in generation. In this blog post we cover KV Caching and all our experiences while implementing it. The lessons learnt are general and can be applied to all autoregressive language model generations.
```
pcuenca:
```suggestion
We have implemented KV Caching from scratch in our [nanoVLM](https://github.com/huggingface/nanoVLM) repository (a small code base to train your own Vision Language Model with pure PyTorch). This gave us a **38%** of speedup in generation. In this blog post we cover KV Caching and all our experiences while implementing it. The lessons learnt are general and can be applied to all autoregressive language model generations. Implementing from scratch on a small codebase is a great learning experience, come along for the ride!
```

kv-cache.md
```diff
@@ -0,0 +1,269 @@
...
+
+![bar plot showcasing improvement in generation speed](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/kv-cache/speed_improved.png)
+
+## Introduction
+
+Autoregressive language models generate text by sampling *one token at a time*. During inference, the model processes a given input sequence, predicts the next token, appends it to the input, and repeats this process until some stopping criterion:
```
pcuenca:
```suggestion
Autoregressive language models generate text by sampling *one token at a time*. During inference, the model processes a given input sequence, predicts the next token, appends it to the sequence, and repeats this process until some stopping criterion:
```

maybe?

kv-cache.md
pcuenca: Yes, this is the correct inline LaTeX syntax for the blog engine.

kv-cache.md
pcuenca: It's an assertion, so if we want to be super explicit we could say that the assertion passes. But I think it's not necessary, an assertion is usually a strong indication for the reader that this is what's expected.

kv-cache.md
```diff
@@ -0,0 +1,269 @@
...
+
+- After processing the initial prompt, we **cache** the computed keys ( \\( K \\) ) and values ( \\( V \\) ) for each layer.
+- During generation, we **only compute** \\( K \\) **and** \\( V \\) **for the new token**, and **append** them to the cache.
+- We compute \\( Q \\) for the current token and use it with the **cached \\( K \\) and \\( V \\)** to get the output.
+
+This changes generation from full-sequence re-computation to a lightweight, incremental update.
+
+> ✅ In practice, this cache is a per-layer dictionary with keys "key" and "value", each of shape (`batch_size`, `num_heads`, `seq_len_cached`, `head_dim`).
+
+This is the foundation of how modern LLMs can generate long outputs efficiently, and KV Caching is what makes it possible.
```
pcuenca:
```suggestion
This is the foundation of how modern LLMs can generate long outputs efficiently.
```

(nit: a bit redundant?)

kv-cache.md
```diff
@@ -0,0 +1,269 @@
...
+
+## KV Caching in nanoVLM: From Theory to Practice
+
+Now that we understand the theory behind KV Caching, let’s see how it’s implemented in practice inside our [nanoVLM](https://github.com/huggingface/nanoVLM) repository.
```
pcuenca:
```suggestion
Now that we understand the theory behind KV Caching, let’s see how it’s implemented in practice inside our [nanoVLM](https://github.com/huggingface/nanoVLM) repository. This is an ideal testbed, as it's a super concise and self-contained codebase.
```

kv-cache.md
```diff
@@ -0,0 +1,269 @@
...
+
+KV caching is enabled across three key components in our model:
+
+1. The **Attention block** that uses and updates the KV cache
+2. The **Language model** that tracks cache per layer
+3. The **Generation loop** that separates **prefill** and **decode** phases
```
pcuenca:
```suggestion
3. The **Generation loop** that separates **prefill** (the initial pass with the input prompt) and sequential **decode** phases
```

kv-cache.md
ariG23498: speed_improved
something like this?

kv-cache.md
```diff
@@ -0,0 +1,269 @@
...
+The biggest architectural change is in the `generate()` method of the `VisionLanguageModel`.
+
+We **split generation into two stages**:
+
+- **PREFILL PHASE:** Encode the full prompt and build the initial cache.
+- **DECODE PHASE:** Generate tokens one at a time using cached keys/values.
+
+```
+PREFILL PHASE (cache construction)
+[Prompt: "What is"] → [Transformer] → [Cache: K, V for all layers]
```
pcuenca, line 221: Perhaps a longer prompt would help drive the point home that this is just for the initial user input?

pcuenca (approved):
🔥

kv-cache.md
ariG23498: Could you help me with an example of that you think would be a nice addition?

kv-cache.md
ariG23498: speed_improved-more
This is with 10K as well.

lusxvr: Looks super nice! I was thinking to have this for the different number of tokens together with the time it would have taken without kv cache (similar to the picture in the beginning), to hammer home the point that KV cache speeds up generation significantly when the sequence gets longer.
image
But my bad for answering to late, I still think the blogpost is gonna be great!

PR #2886, opened 2025-06-03
Fix authors smolvla blog

andimarafioti:
Quick fix for authors


smolvla.md
```diff
@@ -6,10 +6,10 @@ authors:
 - user: andito
 - user: merve
 - user: ariG23498
-- user: francesco
+- user: fracapuano
```
danaaubakirova, line 9: fcapuano

PR #2925, opened 2025-07-02
fix: correct typos

CharlesCNorton:
- Remove redundant "more" from "The more broader term"

- Fix spelling of "architecture" (was "arhitecture") in multiple code examples

- Add missing article "a" in "is a large set corpus"

- Fix spelling of "Overview" (was "Overiew") in documentation reference

### DIFF
--- a/train-sparse-encoder.md
+++ b/train-sparse-encoder.md
@@ -57,7 +57,7 @@ We list the most prominent ones along with their benchmark results in [Pretraine
 
 ## What are Sparse Embedding models?
 
-The more broader term "embedding models" refer to models that convert some input, usually text, into a vector representation (embedding) that captures the semantic meaning of the input. Unlike with the raw inputs, you can perform mathematical operations on these embeddings, resulting in similarity scores that can be used for various tasks, such as search, clustering, or classification.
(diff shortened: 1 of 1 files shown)


train-sparse-encoder.md
```diff
@@ -210,7 +210,7 @@ from sentence_transformers import SparseEncoder
 
 model = SparseEncoder("google-bert/bert-base-uncased")
 # SparseEncoder(
-#   (0): MLMTransformer({'max_seq_length': 512, 'do_lower_case': False, 'arhitecture': 'BertForMaskedLM'})
+#   (0): MLMTransformer({'max_seq_length': 512, 'do_lower_case': False, 'architecture': 'BertForMaskedLM'})
```
pcuenca, line 213: ah, this might be a typo in `SparseEncoder` itself, cc @tomaarsen @arthurbr11

pcuenca (approved):
Thank you! 🙌

train-sparse-encoder.md
arthurbr11: Just checked and there is no sign of `arhitecture` in the repo, but thanks for the ping ! 

The error probably occurred when we modified it ourselves as the format like this was introduce after the first draft of the blog.

PR #2928, opened 2025-07-08
[Add] Efficient MultiModal Data Pipeline

ariG23498:


mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+You've got everything ready - data, model, a beefy GPU setup. You hit "run" and... wait. And wait some more. Your GPUs are barely breaking a sweat while your wallet's getting lighter by the hour.
+
+Sound familiar? We've been there. After some detective work on our [nanoVLM](https://github.com/huggingface/nanovlm) project, we [discovered](https://x.com/andimarafioti/status/1937518080474906641) the real culprit wasn't our model or hardware, it was our data pipeline being incredibly wasteful.
+
+Here's what we found:
+
+1. **Idle GPUs**: Our model was literally waiting around for data to show up
+2. **Padding hell**: Every batch was stuffed with useless padding tokens that contributed nothing to training
+
+In this post we build an efficient pipeline in **five stages**. In each stage we add or remove from the previous step and comment on what went right and did not.
```
pcuenca:
```suggestion
In this post we build an efficient pipeline in **five stages**. In each stage we add or remove from the previous step and comment on what went right and what did not.

The lessons we learned were extracted from [nanoVLM](https://huggingface.co/blog/nanovlm), but they are applicable to any other project.
```

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+- [Stage 0: Pre Requisites](#stage-0-pre-requisites)
+- [Stage 1: Visualising the Dataset](#stage-1-visualising-the-dataset)
+- [Stage 2: Naive Padding](#stage-2-naive-padding)
+- [Stage 3: Constrained Padding](#stage-3-constrained-padding)
+- [Stage 4: Packing Smarter with Knapsacks](#stage-4-packing-smarter-with-knapsacks)
+- [Stage 5: Knapsack for Multimodal Data](#stage-5-knapsacks-for-multimodal-data)
+- [Conclusion](#conclusion)
+
+
+## [Stage 0] Pre Requisites
```
pcuenca:
```suggestion
## [Stage 0] Preparation
```

maybe?

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+We realise that it might be difficult for the reader to map each section with the code in the nanoVLM repository. To address this, we have created a separate repository that aims only at the data pipeline.
```
pcuenca:
```suggestion
To make it easier to follow the data preparation tasks, we created a separate repo laser-focused on the data pipeline only. We hope this will be much easier to understand that reading the code once integrated with the nanoVLM repository. In addition, this could be useful to bootstrap other data pipelines!
```

(a bit silly suggestion, just trying to put the burden on us to make it easier for the reader)

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+Repository: [https://github.com/ariG23498/mmdp](https://github.com/ariG23498/mmdp)
+
+To follow along, all you need to do is clone the repository:
```
pcuenca:
```suggestion
To follow along, all you need to do is clone the repository. It contains the final data preparation tasks, but it's designed to showcase each step of the way.
```

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+```bash
+$ git clone [https://github.com/ariG23498/mmdp.git](https://github.com/ariG23498/mmdp.git)
+```
+
+## [Stage 1] Visualising the Dataset
+
+Before optimizing anything, we needed to understand what we were working with. Our multimodal dataset has images, text prompts, and responses.
```
pcuenca:
```suggestion
Before optimizing anything, we need to understand what we are working with. Our multimodal dataset has images, text prompts, and responses.
```

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+```bash
+$ uv run 01_check_dataset.py
+```
+
+![Dataset Sample](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/mmdp/01.png)
+
+Nothing fancy here, just getting familiar with our data structure.
```
pcuenca:
```suggestion
Getting familiar with your training data is crucial for success. The previous script shows a random sample each time you run it; you may want to copy the snippet to a notebook and run it multiple times to get a feeling about the data.
```

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+## [Stage 2] Naive Padding
+
+Our first attempt was the obvious one:
```
pcuenca:
```suggestion
Our first training attempt used the obvious (and very frequent) approach:
```

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+```bash
+$ uv run 02_naive_pad_dataloader.py
+```
+
+The results were painful. Look at this visualization:
+
+![Naive Padding Waste](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/mmdp/02.png)
+
+See all that gray? That's padding. That's the GPU processing absolutely nothing while you pay for compute time. We were wasting roughly 60% of our batch on empty tokens.
```
pcuenca, line 75: Question for my understanding: is the fact that we use images impacting negatively on the padding ratio, due to the very large amount of tokens usually required? Or are the image tokens ~constant across samples and it's just the text what's causing the difference?

(If it's the images, I'd stress it so readers understand VLMs suffer more from this problem than other models).

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+```python
+def _get_data_range(self):
+    worker_info = get_worker_info()
+    # Logic to split data range across workers
+```
+
+### Producer-Consumer Magic
+
+Packing sequences can be slow, especially if we’re sorting or shuffling. To keep things moving, we use a **producer-consumer** pattern:
```
pcuenca:
```suggestion
Packing sequences can be slow, especially if we’re sorting or shuffling. To keep things moving, we use a **producer-consumer** pattern using Python queues:
```

(maybe link to some `Queue` docs, not sure)

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+This walks through the data sequentially, adding items to a pack until it’s full, then starting a new one. It’s fast but not perfect. Here’s what the batches look like:
+
+```bash
+=== Strategy: GREEDY ===
+[tensor([1]), tensor([2]), tensor([3]), tensor([4]), tensor([5]), tensor([6]), tensor([7]), tensor([8]), tensor([9]), tensor([10]), tensor([11]), tensor([12]), tensor([13])]
+[tensor([14]), tensor([15]), tensor([16]), tensor([17]), tensor([18]), tensor([19])]
+[tensor([20]), tensor([21]), tensor([22]), tensor([23])]
+[tensor([24])]
```
pcuenca, line 166: - But these are not random
- Also, I'd maybe show the max batch size, just as we did with the blue/gray plots. I suppose it's 100, so the last batch is super empty (although not very representative, as there's only 1 item remaining)

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+    for item in buffer:
+        for pack in knapsacks:
+            if sum(pack) + item <= self.max_length:
+                pack.append(item)
+                break
+        else:
+            knapsacks.append([item])
+```
+
+This sorts sequences by length (biggest first) and tries to fit each one into the first pack that has room. If none fit, it starts a new pack. The result?
```
pcuenca:
```suggestion
This sorts sequences by length (longest first) and tries to fit each one into the first pack that has room. If none fits, it starts a new pack. The result?
```

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+| Concept | Stage 4 (Toy Data) | Stage 5 (Multimodal Data) | Function(s) |
+|---------|--------------------|---------------------------|-------------|
+| **Item** | Integer (sequence length) | Full sample (image, prompt, response) | `VQADataset.__getitem__` |
+| **Weight** | The integer itself | Number of tokens (`len(input_ids)`) | — |
+| **Knapsack** | Batch of integers ≤ `max_length` | Batch of samples ≤ `seq_length` and image limit | `_balanced_greedy_knapsack` |
+| **Packing Strategy** | Greedy or Binpack | Greedy packing with token and image constraints | `_balanced_greedy_knapsack` |
+| **Producer-Consumer** | Producer fills queue | Same, but with real samples | `_producer`, `__iter__` |
+| **Sample Filtering** | Skip integers > `max_length` | Skip samples with too many tokens or images | `_producer` |
+| **Sharding** | Split integer range | Shard dataset indices | `make_base_iterator()` |
+| **Batching** | Group integers | Concatenate and align tokens/images | `_pack_one_group` |
```
pcuenca, line 220: "collating", maybe, if the whole process is about creating batches?

(not sure if this would cause confusion with existing collating functions in the code)

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+The `ConstantLengthDataset` does it all:
+- Reads samples (images and text).
+- Filters out samples that are too long or have too many images.
+- Packs samples into batches using a greedy knapsack strategy, balancing token count and image count.
+- Pads the final batches to a fixed length, but with *way less* padding than before.
+
+Here’s the result:
+
+![Knapsack Padding](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/mmdp/05.png)
```
pcuenca, line 231: This is still the fixed max length strategy.

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+Look at that! The gray (padding) is minimal, and the batches are dense with useful data. It’s like packing a suitcase so well you can still zip it up without sitting on it.
```
pcuenca, line 233: lol

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
```
pcuenca, line 234: At this point, as a reader I'd love to see a brief discussion about randomness, especially in the presence of multi-epoch training runs. Just mentioning that we can partition data into big random chunks at the beginning of each epoch and then apply these techniques would be enough in my opinion.

pcuenca (approved):
Amazing, great work! 🔥

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+title: "Efficient MultiModal Data Pipeline in nanoVLM" 
+thumbnail: /blog/assets/mmdp/thumbnail.png
+authors:
+- user: ariG23498
+- user: lusxvr
+- user: andito
+- user: sergiopaniego
+---
+
+# Efficient MultiModal Data Pipeline in nanoVLM
```
pcuenca: Crazy idea: thoughts on removing nanoVLM from the title to make it more general?

mmdp.md
sergiopaniego: If this is applied, it needs to also be applied in the table of contents.

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+## [Stage 0] Pre Requisites
+
+We realise that it might be difficult for the reader to map each section with the code in the nanoVLM repository. To address this, we have created a separate repository that aims only at the data pipeline.
+
+Repository: [https://github.com/ariG23498/mmdp](https://github.com/ariG23498/mmdp)
+
+To follow along, all you need to do is clone the repository:
+
+```bash
+$ git clone [https://github.com/ariG23498/mmdp.git](https://github.com/ariG23498/mmdp.git)
```
sergiopaniego:
```suggestion
$ git clone https://github.com/ariG23498/mmdp.git
```

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+Our first attempt was the obvious one:
+- Tokenize everything
+- Find the longest sequence in each batch  
+- Pad everything else to match
+
+```bash
+$ uv run 02_naive_pad_dataloader.py
+```
+
+The results were painful. Look at this visualization:
```
sergiopaniego, line 71: There is a mix of past and present tenses throughout the text. The blog would benefit from a thorough review to ensure consistency!

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+See all that gray? That's padding. That's the GPU processing absolutely nothing while you pay for compute time. We were wasting roughly 60% of our batch on empty tokens.
+
+## [Stage 3] Constrained Padding
+
+Our next move was simple. Set a global maximum length and stick to it. If a sample was too long, we'd just drop it.
+
+![Constrained Padding](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/mmdp/03.png)
+
+This helped, but we were still padding everything to the same fixed length regardless of actual content. Better than before, but still wasteful.
```
sergiopaniego: Question for a reader perspective :) 
Why are we using 512?

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+## [Stage 4]: Packing Smarter with Knapsacks
+
+Now we’re ready to rethink batching entirely. Padding is the enemy, and we need a strategy to minimize it while maximizing how much data we can fit into each batch. Enter the **knapsack problem**, a classic from computer science that’s perfect for this.
```
sergiopaniego: Example of the problem with present and past tenses 😄

sergiopaniego: Maybe link to a broader explanation of the problem? Like: https://en.wikipedia.org/wiki/Knapsack_problem

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+These batches are *much* tighter, with less wasted space. It’s like playing Tetris with your data, fitting pieces together snugly.
+
+## [Stage 5] Knapsacks for Multimodal Data
+
+Now for the real deal, applying knapsack packing to our *multimodal* dataset.
+
+We’re back to images, prompts, and responses, and we need to pack them efficiently while respecting both token limits *and* image budgets (since GPUs can only handle so many images per batch).
+
+Our new [`ConstantLengthDataset`](https://github.com/ariG23498/mmdp/blob/98ffaa03a5df82b7ebeef2d75270059d64d85663/src/mmdp/advanced_torch_datasets.py#L13) class handles the heavy lifting. Here’s how it works, compared to Stage 4:
```
sergiopaniego:
```suggestion
Our new [`ConstantLengthDataset`](https://github.com/ariG23498/mmdp/blob/main/src/mmdp/advanced_torch_datasets.py#L13) class handles the heavy lifting. Here’s how it works, compared to Stage 4:
```

sergiopaniego (approved):
Nice! 🤏

mmdp.md
sergiopaniego: +1
To present it in a more general way instead of narrowing the audience.

andimarafioti: Completely fine by me, we can say that this is implemented in nanoVLM if people want to see it in action

mmdp.md
andimarafioti: It's not the images. It's the variety in sample length among VLM SFT datasets, more are quite short but some are longer. But without padding, at this stage, each of those examples has one image.

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+![Naive Padding Waste](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/mmdp/02.png)
+
+See all that gray? That's padding. That's the GPU processing absolutely nothing while you pay for compute time. We were wasting roughly 60% of our batch on empty tokens.
+
+## [Stage 3] Constrained Padding
+
+Our next move was simple. Set a global maximum length and stick to it. If a sample was too long, we'd just drop it.
+
+![Constrained Padding](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/mmdp/03.png)
```
andimarafioti, line 81: here the plot has only 7 examples, if it isn't too much work, I think keeping it to 8 would help since people might think we are changing the batch size if we drop a sample

andimarafioti: which we actually kind of did in nanovlm for a while xD

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Switching to an Iterable Dataset
+
+Most PyTorch datasets are *map-style* (you access them with `dataset[i]`). But for dynamic batching, we need something more flexible. So, we built an *iterable-style* dataset by subclassing `torch.utils.data.IterableDataset`. This lets us generate batches on the fly and handle tricks like sharding data across multiple workers:
+
+```python
+def _get_data_range(self):
+    worker_info = get_worker_info()
+    # Logic to split data range across workers
+```
```
andimarafioti: This is hardly the logic to split the data range across workers, or is the comment something like a code TLDR? (Here would come...). Not clear to me

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Producer-Consumer Magic
+
+Packing sequences can be slow, especially if we’re sorting or shuffling. To keep things moving, we use a **producer-consumer** pattern:
+
+```python
+def _producer(self, data_iter, queue, stop_signal):
+    # Compute batches and fill queue
```
andimarafioti: again, this comments in code make me think that you're explaining what comes, and then the code below does something else completely. Why make the queue.put(stop_signal) explicit and everything else implicit? At this point, maybe pseudo-code is better

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+This sorts sequences by length (biggest first) and tries to fit each one into the first pack that has room. If none fit, it starts a new pack. The result?
+
+```bash
+=== Strategy: BINPACK ===
+[tensor([24]), tensor([23]), tensor([22]), tensor([21]), tensor([10])]
+[tensor([20]), tensor([19]), tensor([18]), tensor([17]), tensor([16]), tensor([9]), tensor([1])]
+[tensor([15]), tensor([14]), tensor([13]), tensor([12]), tensor([11]), tensor([8]), tensor([7]), tensor([6]), tensor([5]), tensor([4]), tensor([3]), tensor([2])]
+```
+
+These batches are *much* tighter, with less wasted space. It’s like playing Tetris with your data, fitting pieces together snugly.
```
andimarafioti, line 201: I would put a visualization here instead of the bash stuff, you can follow the barplots from before

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+## [Stage 5] Knapsacks for Multimodal Data
+
+Now for the real deal, applying knapsack packing to our *multimodal* dataset.
+
+We’re back to images, prompts, and responses, and we need to pack them efficiently while respecting both token limits *and* image budgets (since GPUs can only handle so many images per batch).
```
andimarafioti: IT's not that GPUs can only handle so many images, is that we want images per sample to be balanced since to avoid the case where one GPU needs to process way more images than another one

mmdp.md
andimarafioti: lolx2

mmdp.md
andimarafioti: actually we do randomize the knapsacks, otherwise there is some performance loss :) You can say that we use way more samples than necessary for a batch and then randomize, I think we are creating like 1024 samples (fully knapsacked)

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+Here’s the result:
+
+![Knapsack Padding](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/mmdp/05.png)
+
+Look at that! The gray (padding) is minimal, and the batches are dense with useful data. It’s like packing a suitcase so well you can still zip it up without sitting on it.
+
+## Conclusion
+
+What started as a simple "why is training so slow?" investigation led to a complete rethink of how we handle multimodal data.
```
andimarafioti, line 246: Maybe mention somewhere that this balanced knapsack strategy comes from the Eagle 2 paper from NVidia? It sounds a little bit like we are inventing something here

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+**The key lessons:**
+- Don't pad everything to the longest sequence (that's wasteful)
```
andimarafioti: maybe more like "padding everything to the longest sequences is a good first approach, but once you care about speed you need smth better"? But less long xD

andimarafioti (commented):
Great work ! It's looking amazing already, left some comments where I found some room for improvement :)

mmdp.md
ariG23498: The title is present in the thumbnail as well.
Do you all want me to edit the thumbnail as well?

The title would be "Efficient MultiModal Data Pipeline".

mmdp.md
ariG23498: @pcuenca I agree with the section name. Thank @sergiopaniego for the tip.

mmdp.md
ariG23498: I wanted to keep this in order for the reader to understand that we have indeed dropped one of the sample due to its length.

Do you think I should document it better after the image?

mmdp.md
ariG23498: With a smaller number a lot of samples were being discarded, with a larger number the bar plots seemed very sparse.

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+
+Here's what we found:
+
+1. **Idle GPUs**: Our model was literally waiting around for data to show up
+2. **Padding hell**: Every batch was stuffed with useless padding tokens that contributed nothing to training
+
+In this post we build an efficient pipeline in **five stages**. In each stage we add or remove from the previous step and comment on what went right and did not.
+
+Table of Contents:
+- [Stage 0: Pre Requisites](#stage-0-pre-requisites)
```
merveenoyan: putting stage repetitively makes it a bit hard to read imo

mmdp.md
```diff
@@ -0,0 +1,225 @@
...
+Sound familiar? We've been there. After some detective work on our [nanoVLM](https://github.com/huggingface/nanovlm) project, we [discovered](https://x.com/andimarafioti/status/1937518080474906641) the real culprit wasn't our model or hardware, it was our data pipeline being incredibly wasteful.
+
+Here's what we found:
+
+1. **Idle GPUs**: Our model was literally waiting around for data to show up
+2. **Padding hell**: Every batch was stuffed with useless padding tokens that contributed nothing to training
+
+In this post we build an efficient pipeline in **five stages**. In each stage we add or remove from the previous step and comment on what went right and did not.
+
+Table of Contents:
```
merveenoyan:
```suggestion
### Table of Contents
```
I'd format

mmdp.md
```diff
@@ -0,0 +1,247 @@
...
+
+Our new [`ConstantLengthDataset`](https://github.com/ariG23498/mmdp/blob/main/src/mmdp/advanced_torch_datasets.py#L13) class handles the heavy lifting. Here’s how it works, compared to Stage 4:
+
+| Concept | Stage 4 (Toy Data) | Stage 5 (Multimodal Data) | Function(s) |
+|---------|--------------------|---------------------------|-------------|
+| **Item** | Integer (sequence length) | Full sample (image, prompt, response) | `VQADataset.__getitem__` |
+| **Weight** | The integer itself | Number of tokens (`len(input_ids)`) | — |
+| **Knapsack** | Batch of integers ≤ `max_length` | Batch of samples ≤ `seq_length` and image limit | `_balanced_greedy_knapsack` |
+| **Packing Strategy** | Greedy or Binpack | Greedy packing with token and image constraints | `_balanced_greedy_knapsack` |
+| **Producer-Consumer** | Producer fills queue | Same, but with real samples | `_producer`, `__iter__` |
```
merveenoyan: I don't get this part :( Same, but with real samples

merveenoyan (approved):
mostly nits, I learnt a lot :)

mmdp.md
ariG23498: 1. These are not meant to be random. The dataset is a sequential range of 1 through 25.
2. Nice point. I have added the graph to be more intuitve.

mmdp.md
ariG23498: Knapsack needs a max length (the size of the knapsack). This is why it seems similar to the fixed length strategy.

The difference is that we have less of padding and more of tokens. The visualization is a little off, as we use the attention masks for it, and we cannot demarcate two or three sequences with different color scheme.

mmdp.md
ariG23498: What I meant to say is that, it is the same with the toy dataset, but with multimodal inputs.

I will rephrase this.

mmdp.md
ariG23498: In the TOC you mean? What would be a better approach? I wanted to highlight the **five stages**, hence the prefix.

I am open to suggestions.

mmdp.md
pcuenca: Yeah, I'd keep the image and explain in the text. Or, alternatively, add the row to the image and explain it's dropped in the image itself, but that's more work.

mmdp.md
pcuenca: Why does it have to be sequential? I thought we were simplifying the problem by representing sequence lengths with single numbers, but if the sequences are sorted then the problem is different.

(But ok to publish like this, I was anticipating questions readers may ask themselves)

mmdp.md
pcuenca: maybe put the fixed + the knapsack side by side? (in that case perhaps we need the dropped batch)

mmdp.md
pcuenca: It's ok for me if the thumbnail title does not fully match the post's.

PR #2943, opened 2025-07-09
fix: correct typos in Reachy Mini blog post

CharlesCNorton:
- Fix decimal notation in weight specification (1,5 kg → 1.5 kg) Use standard English decimal point notation instead of European comma notation for consistency with the rest of the document (e.g., "3.3 lbs")

- Add missing article "the" before "Reachy-Mini community" Proper English grammar requires the definite article when referring to a specific community group

- Remove extra hyphen in "human-robot interactions" Follow standard compound noun formatting where "human-robot" is hyphenated as a compound modifier, but "interactions" remains separate without additional hyphenation


pcuenca (approved):
Thank you @CharlesCNorton, I agree with them all.

cc @ariG23498, @Vaibhavs10 

(I would personally use the International Units first to avoid sounding too us-centric, but I respect the authors style choice).

PR #2995, opened 2025-07-25
Update lora-fast post with more technical details

BenjaminBossan:
Add a new subsection to explain the steps required for hotswapping to work with torch compilation.

Note: My editor removed trailing whitespace, I think that's fine.

### DIFF
--- a/lora-fast.md
+++ b/lora-fast.md
@@ -12,7 +12,7 @@ LoRA adapters provide a great deal of customization for models of all shapes and
 
 In this post, we take the [Flux.1-Dev model](https://huggingface.co/black-forest-labs/FLUX.1-dev) for text-to-image generation because of its widespread popularity and adoption, and how to optimize its inference speed when using LoRAs (\~2.3x). It has over 30k adapters trained with it ([as reported](https://huggingface.co/models?other=base_model:adapter:black-forest-labs/FLUX.1-dev) on the Hugging Face Hub platform). Therefore, its importance to the community is significant.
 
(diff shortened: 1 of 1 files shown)


lora-fast.md
```diff
@@ -174,9 +174,13 @@ In the table below, we show the inference latency numbers with different combina
 
 **Quick notes**:
 
-* Compilation provides a massive 2x speedup over the baseline.  
+* Compilation provides a massive 2x speedup over the baseline.
 * The other options yielded OOM errors even with offloading enabled.
 
+## Technical details of hotswapping
+
+To enable hotswapping without triggering recompilation, two hurdles have to be overcome. First, the LoRA scaling factor has to be converted into torch tensors from floats, which is achieved fairly easily. Second, the shape of the LoRA weights needs to padded to the largest required shape. That way, the data in the weights can be replaced without the need to reassign the whole attribute. This is why the `max_rank` argument discussed above is crucial. To see the nitty-gritty of this implementation, visit the [`hotswap.py` file in PEFT](https://github.com/huggingface/peft/blob/92d65cafa51c829484ad3d95cf71d09de57ff066/src/peft/utils/hotswap.py).
```
sayakpaul: I think we should also talk a bit about `target_modules` i.e., the first LoRA we're using for hotswapping should have the maximal set of target modules? And maybe why that's the case?

And then making it explicitly clear that padding the LoRA params with zeroes isn't a problem in terms of the expected outputs.

BenjaminBossan: Good point, I added a few more sentences on that.

PR #2997, opened 2025-07-26
Fix typos in Parquet CDC blog post

CharlesCNorton:
- Fix heading: "Re-uploading an Exact Copies" → "Re-uploading Exact Copies"
- Fix spelling: "regisions" → "regions"
- Add missing article: "in case of pyarrow" → "in the case of pyarrow"
- Add missing verb: "Datasets often split" → "Datasets are often split"


pcuenca (approved):
Thank you!

PR #3000, opened 2025-07-29
fix: assisted-generation code example #2 in zh

boilcy:
- Fix code syntax errors in the Chinese (zh) version  
- Align implementation with the English (en) version


boilcy:
@pcuenca

PR #3001, opened 2025-07-29
fix: Correct indentation in usage example

CharlesCNorton:
The 'trackio.finish()' call in the simulate_multiple_runs example was incorrectly placed outside the main loop. This change moves it inside the loop to correctly finalize each run after its epochs are completed, reflecting the proper usage pattern.

### DIFF
--- a/trackio.md
+++ b/trackio.md
@@ -84,7 +84,7 @@ def simulate_multiple_runs():
                 "val_accuracy": val_acc
             })
             time.sleep(0.2)
-    trackio.finish()
+        trackio.finish()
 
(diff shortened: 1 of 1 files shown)


pcuenca:
Thank you @CharlesCNorton! Perhaps there's no practical difference, but it could potentially be clearer the way you suggest. cc @abidlabs @qgallouedec for opinion.

abidlabs (approved):
lgtm thanks @CharlesCNorton! Also used this opportunity to fix a few other things in the blog post

PR #3023, opened 2025-08-12
Add months instead of seasons for regions where seasons are different

clmnt:
from a user comment on discord


pcuenca (approved):
Thank you!

PR #3060, opened 2025-09-09
Explicit `orgs` field in the frontmatter

mishig25:
Explicit `orgs` field in the frontmatter

<details>
<summary>python script that was used</summary>

```py
#!/usr/bin/env python3
"""
Add an `orgs` field to YAML frontmatter of all Markdown files by extracting the
`org` value from each author in the `authors` list.

- Only files that start with YAML frontmatter delimited by `---` are processed
- If no authors with `org` are present, the file is left unchanged
- Uses ruamel.yaml RoundTrip to preserve formatting, comments, and key order

Usage:
  python add_orgs_to_frontmatter.py [root_dir]
If root_dir is omitted, current working directory is used.

Requires:
  pip install ruamel.yaml
"""

import io
import os
import re
import sys
from typing import List, Optional, Tuple

try:
    from ruamel.yaml import YAML
    from ruamel.yaml.comments import CommentedMap
except Exception as exc:  # pragma: no cover
    print(
        "ERROR: Missing dependency 'ruamel.yaml'.\n"
        "Install it with: pip install ruamel.yaml",
        file=sys.stderr,
    )
    sys.exit(1)

FRONTMATTER_REGEX = re.compile(r"^---\s*\r?\n(.*?)\r?\n---\s*\r?\n?", re.DOTALL)


def find_frontmatter_bounds(text: str) -> Optional[Tuple[int, int, str]]:
    """Return (start_index, end_index, yaml_str) for the YAML frontmatter.

    If not present, return None.
    """
    if not text.startswith("---"):
        return None
    match = FRONTMATTER_REGEX.match(text)
    if not match:
        return None
    yaml_str = match.group(1)
    return (match.start(), match.end(), yaml_str)


def extract_orgs_from_authors(authors_value) -> List[str]:
    """Extract unique org names from `authors` while preserving order.

    Accepts typical structures like a list of mappings where each mapping
    may contain an `org` key.
    """
    orgs: List[str] = []
    if not isinstance(authors_value, list):
        return orgs

    for item in authors_value:
        if isinstance(item, dict) and "org" in item:
            org_value = item.get("org")
            if isinstance(org_value, str):
                normalized = org_value.strip()
                if normalized and normalized not in orgs:
                    orgs.append(normalized)
    return orgs


def update_frontmatter_with_orgs(yaml_text: str) -> Optional[str]:
    """Given the YAML frontmatter text, return updated YAML text or None if no change.

    The function preserves formatting via ruamel.yaml round-trip loader/dumper.
    """
    yaml = YAML(typ="rt")
    yaml.preserve_quotes = True
    yaml.width = 4096  # avoid line wrapping

    data = yaml.load(yaml_text)

    if not isinstance(data, CommentedMap):
        return None

    authors_value = data.get("authors")
    orgs = extract_orgs_from_authors(authors_value)

    if not orgs:
        return None

    # Set or replace the orgs field. This will append to the end in round-trip mode.
    data["orgs"] = orgs

    stream = io.StringIO()
    yaml.dump(data, stream)
    updated_yaml_text = stream.getvalue()

    # ruamel.yaml tends to ensure trailing newline; keep as-is.
    return updated_yaml_text


def process_markdown_file(file_path: str) -> bool:
    """Process a single .md file. Returns True if the file was modified."""
    try:
        with open(file_path, "r", encoding="utf-8") as f:
            original_text = f.read()
    except Exception as exc:
        print(f"SKIP (read error): {file_path} -> {exc}")
        return False

    bounds = find_frontmatter_bounds(original_text)
    if not bounds:
        return False

    start_idx, end_idx, yaml_text = bounds

    updated_yaml_text = update_frontmatter_with_orgs(yaml_text)
    if updated_yaml_text is None or updated_yaml_text == yaml_text:
        return False

    # Reconstruct the document with updated frontmatter
    before = original_text[:start_idx]
    after = original_text[end_idx:]

    # Ensure delimiters and newlines mirror the typical pattern
    new_text = f"---\n{updated_yaml_text}---\n{after}"

    if new_text == original_text:
        return False

    try:
        with open(file_path, "w", encoding="utf-8") as f:
            f.write(new_text)
        return True
    except Exception as exc:
        print(f"SKIP (write error): {file_path} -> {exc}")
        return False


def iter_markdown_files(root_dir: str):
    for current_dir, _subdirs, files in os.walk(root_dir):
        for name in files:
            if name.lower().endswith(".md"):
                yield os.path.join(current_dir, name)


def main() -> int:
    root_dir = sys.argv[1] if len(sys.argv) > 1 else os.getcwd()
    root_dir = os.path.abspath(root_dir)

    if not os.path.isdir(root_dir):
        print(f"ERROR: Not a directory: {root_dir}", file=sys.stderr)
        return 2

    modified_count = 0
    total_count = 0

    for file_path in iter_markdown_files(root_dir):
        total_count += 1
        if process_markdown_file(file_path):
            modified_count += 1
            print(f"UPDATED: {file_path}")

    print(f"Done. Checked {total_count} file(s), updated {modified_count}.")
    return 0


if __name__ == "__main__":
    raise SystemExit(main()) 
```

</details>


autoround.md
```diff
@@ -2,30 +2,30 @@
...
+- user: n1ck-guo
+  guest: true
+  org: Intel
+- user: isaacmac
+  guest: true
+  org: Intel
+- user: kding1
+  guest: true
+  org: Intel
+- user: IlyasMoutawwakil
```
Vaibhavs10, line 23: is it correct that we don't tag `huggingface` org (not sure how you want to show it in the front-end)

julien-c (changes requested):
i don't think we need this TBH. We can only use this going forward (and only when we want to just display the org(s), which is not always what we want)

PR #3089, opened 2025-09-19
Add a `xet` tag to find all the articles easily

julien-c:
https://huggingface.co/blog?tag=xet

cc @jsulz @rajatarya


jsulz (approved):
Sweet - thank you!

PR #3101, opened 2025-09-24
add author to scaleway blogpost

SBrandeis:


pcuenca (approved):
(sidenote: they doesn't seem to be members of the org)

PR #3138, opened 2025-10-21
Fix thumbnail path

merveenoyan:


merveenoyan:
IDK why it's not fixed ugh

PR #3173, opened 2025-11-18
Update remote_vae.md to include hlky's contribs

sayakpaul:
Similar to https://huggingface.co/blog/vid_ds_scripts.


pcuenca (approved):

sayakpaul:
They don't which is why the first blog post was edited like that.

pcuenca:
I would encourage them to sign up, if possible.

PR #3176, opened 2025-11-19
[New blog post] continuous_batching

remi-or:
This PR adds the `continuous_batching` article titled "Continuous batching from first principles"
The drafts were reviewed by @McPatate and @ArthurZucker 
I followed the README so the article *should* have the right format.


continuous_batching.md
```diff
@@ -0,0 +1,171 @@
...
+- user: ror
+- user: ArthurZ
+- user: mcpotato
+---
+
+# Continuous batching
+
+![Title card](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/banner.png)
+
+## Introduction
```
burtenshaw:
```suggestion
```

assets/continuous_batching/thumbnail.png
LysandreJik, line 1: cute

continuous_batching.md
```diff
@@ -0,0 +1,171 @@
...
+
+This generation process is computationally expensive: it requires processing billions of parameters for each token generated. To make these models practical for real-world applications, particularly when serving many users simultaneously, researchers and engineers have developed a range of efficient inference techniques.  
+One of the most impactful optimizations is **continuous batching**, which is the focus of this article. To understand how continuous batching works and why it's so effective, we'll need to build up from the fundamentals of how LLMs process tokens.
+
+## Attention
+
+In order to understand continuous batching, it’s useful to have a clear understanding of the attention mechanism. In a Llama-like LLM, the neural network processes text by breaking it down into pieces that we call tokens, which are represented by vectors. For simplicity's sake, let's pretend one token corresponds to one word. For each token, the network computes a prediction of what the next token should be. It does this by applying a sequence of operations to each token.
```
LysandreJik: You say "For simplicity's sake, let's pretend one token corresponds to one word.", but in the first example in the next paragraph, you're already splitting "project" into "pro" and "ject" 😁

continuous_batching.md
```diff
@@ -0,0 +1,171 @@
...
+
+Most operations in the network are **token-wise**: each token is processed independently, and the output for a given token depends only on that token's content, not on any other tokens in the sequence. Operations like this include layer normalization or matrix multiplication. However, to create connections between words in a sentence, we need some operations where tokens can influence each other. 
+
+This is where attention comes in. **Attention layers are the only place where different tokens interact with each other**—something that even seasoned ML practitioners sometimes overlook. Understanding how a network connects tokens together means understanding attention.
```
LysandreJik: Let's try not to use em dashes as they're often a synonym of AI usage

continuous_batching.md
LysandreJik: IMO it's probably better to say something like

"We can think of tokens generally representing a single, simple word. For longer or rare words, several tokens may be used"

or something along those lines

continuous_batching.md
```diff
@@ -0,0 +1,171 @@
...
+
+Let's see how this works in practice. To explain the basics of attention, we'll focus on the case where we have one input prompt. This is the same as saying batch size is 1.
+
+Consider the initial prompt "I am sure this project". It is tokenized as 7 tokens: `[<bos>, I, am, sure, this, pro, ject]`. The `<bos>` token is a special token that is added at the start of the prompt (BoS stands for Beginning of Sequence).  
+As a broad-stroke picture, attention can be represented this way:
+
+![attention.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/attention.png)
```
LysandreJik: I find this picture followed by the text a bit complex to go through as it displays a lot of terms we have not seen before. We've heard of tokens and attention, and now we have attention masks, scores, outputs, some Wq, some Wk, some Wv, etc.

Would it make sense to break this down in simpler blocks, and gradually build up to that image?

LysandreJik: It's likely not a lot of work: your paragraphs are already split in a way where each concept is isolated. I think it's probably a matter of having isolated images to accompany each concept

continuous_batching.md
```diff
@@ -0,0 +1,171 @@
...
+
+Finally, after applying the attention mask, we take a row-wise softmax and multiply the result by the value projection $V$ to get the output of one attention head, of shape $\left[ 1, n , A \right]$.
+
+We are going to use a lot of attention visualization in this post, so to simplify things, we are going to condense the figure above just a bit.
+
+**Why this matters:** In continuous batching, $Q$, $K$, and $V$ can have different numbers of tokens because we're processing different stages together (prefill and decode). Let's say $Q$ has shape $\left[1, n_Q , A \right]$, $K$ has shape $\left[ 1, n_K , A \right]$, and $V$ has shape $\left[ 1, n_V , A \right]$.
```
LysandreJik: This is the first time we hear of the term "prefill" which is then used a lot in the blog. Would it make sense to explain what it is?

continuous_batching.md
```diff
@@ -0,0 +1,171 @@
...
+For instance, Llama-2-7B with $\mathcal{L}=32$ layers, $H=32$ heads, and $A=128$ requires $2 \times 32 \times 128 = 8,192$ values per token per layer. With `float16` precision, this takes $2AH \times 2$ bytes $= 16$ KB in memory.
+
+KV caching is useful when we have one input token, which is a stage we call **decoding**. But it can also be useful in the **prefill** stage, when we process the initial prompt and have many input tokens. Especially when there are large initial prompts that don't fit in GPU memory all at once.
+
+## Chunked prefill
+
+Up till now, we have looked at an example of prefill where we have $n=7$ tokens, but in practice initial prompts can be much longer. For instance, when using Cursor, you can add your repository as context before the prompt: this increases the prompt size by a lot. In such cases, the memory needed to store the activations for $n$ tokens can be larger than the available memory on the GPU. Thus we cannot perform prefill in a single forward pass: we have to split the prefill in chunks. This is called **chunked prefill**, and it's going to be one of the components needed to enable efficient inference.
```
LysandreJik: "you can add your repository as context before the prompt: this increases the prompt size by a lot."

This is a bit weird: you differentiate context and prompt, yet when adding context, the prompt changes

```suggestion
Up till now, we have looked at an example of prefill where we have $n=7$ tokens, but in practice initial prompts can be much longer. For instance, when using Cursor, you can add your repository to the prompt, where it acts as context: this significantly increases the prompt size. In such cases, the memory needed to store the activations for $n$ tokens can be larger than the available memory on the GPU. Thus we cannot perform prefill in a single forward pass: we have to split the prefill in chunks. This is called **chunked prefill**, and it's going to be one of the components needed to enable efficient inference.
```

assets/continuous_batching/thumbnail.png
remi-or: thanks!

continuous_batching.md
remi-or: Done!

continuous_batching.md
remi-or: Your rephrase works well, using it.

continuous_batching.md
remi-or: Added it in the introduction!

continuous_batching.md
remi-or: Did just that, let me know if the breakdown is comprehensive enough!

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+- user: mcpotato
+---
+
+# Continuous batching
+
+![Title card](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/banner.png)
+
+## Introduction
+
+If you've ever used Qwen, Claude, or any other AI chatbot, you've probably noticed something: it takes a while for the first word of the response to appear, and then words appear one-by-one on your screen with (hopefully) a regular and fast-paced frequency. That's because at the heart of it, all LLMs are just fancy next token predictors. First comes the **prefill** phase, where they ingest your initial prompt to predict one new token. Then they ingest that one as well, to produce another new token: this is the **decoding** phase. The process is repeated until they feel like they have generated enough.
```
merveenoyan:
```suggestion
If you've ever used Qwen, Claude, or any other AI chatbot, you've probably noticed something: it takes a while for the first word of the response to appear, and then words appear one-by-one on your screen with (hopefully) a regular and fast-paced frequency. That's because at the heart of it, all LLMs are just fancy next token predictors. During generation, first comes the **prefill** phase, where the model ingests your initial prompt to predict one new token. Then they ingest that token as well, to produce another new token: this is the **decoding** phase. The process is repeated until they feel like they have generated enough.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+This generation process is computationally expensive: it requires processing billions of parameters for each token generated. To make these models practical for real-world applications, particularly when serving many users simultaneously, researchers and engineers have developed a range of efficient inference techniques.
```
merveenoyan:
```suggestion
This generation process is computationally expensive: it requires passing the input through billions of parameters for each token generated. To make these models practical for real-world applications, particularly when serving many users simultaneously, researchers and engineers have developed a range of efficient inference techniques.  
```
otherwise sounds like you process the params themselves

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+Next, tensors $Q$ and $K$ are multiplied together to measure similarity between tokens, producing a tensor of shape $\left[ 1, n , n \right]$. This is why we say that attention has quadratic complexity in sequence length. Computing $QK^T$ requires $\mathcal{O} \left( n^2 d \right)$ operations, so the cost is a square of $n$ the sequence length. It is represented on the right in the figure above.
+
+We then apply a boolean **attention mask** to $QK^T$ to control which tokens can interact, as represented in the figure below. In this figure, the attention mask is a **causal mask**, meaning each token only interacts with tokens that came before it. This follows the intuition that a cause must come before its consequence, hence the name causal mask. The attention mask is crucial because it dictates all token interactions in the network. **Set all attention mask values to False and no token will ever interact with another in the whole network.** We'll examine attention masks more closely in a few paragraphs.
+
+![masking_and_softmax.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/masking_and_softmax.png)
```
merveenoyan, line 45: not a must but I'd love to see the tokens on the left hand side of the matrices

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+Finally, after applying the attention mask, we take a row-wise softmax and multiply the result by the value projection $V$ to get the output of one attention head, of shape $\left[ 1, n , A \right]$. We offer a visual summary of the whole process in the following figure.
+
+![attention.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/attention.png)
```
merveenoyan, line 49: ^ this is what I had in mind, you already to this here, would be nice to also have in above

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+We then apply a boolean **attention mask** to $QK^T$ to control which tokens can interact, as represented in the figure below. In this figure, the attention mask is a **causal mask**, meaning each token only interacts with tokens that came before it. This follows the intuition that a cause must come before its consequence, hence the name causal mask. The attention mask is crucial because it dictates all token interactions in the network. **Set all attention mask values to False and no token will ever interact with another in the whole network.** We'll examine attention masks more closely in a few paragraphs.
+
+![masking_and_softmax.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/masking_and_softmax.png)
+
+Finally, after applying the attention mask, we take a row-wise softmax and multiply the result by the value projection $V$ to get the output of one attention head, of shape $\left[ 1, n , A \right]$. We offer a visual summary of the whole process in the following figure.
```
merveenoyan: nit: would be nice to say token-wise softmax to be coherent

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+The attention scores $QK^T$ then have shape $\left[ 1, n_Q , n_K \right]$, and the attention mask has the same shape since it's applied point-wise to the scores.
+
+After applying the attention mask and row-wise softmax, we multiply by $V$. Since we're multiplying a matrix of shape $\left[ 1, n_Q , n_K \right]$ by one of shape $\left[ 1, n_V , A \right]$, the inner dimensions must match: $n_K = n_V$. This means $V$ and $K$ always have the same length, so we can simplify our visualizations by only showing $K$.  
+Don't worry if this seems abstract: the figures will make it concrete.
+
+
+Furthermore, since we know that the attention mask is applied to $QK^T$, we know they have the same shape. Instead of representing the attention scores, we will represent the attention mask in its place.
+Finally, since $Q$, $K$ and $V$ are direct projections of $x$, no need to represent $x$. This gives the simplified figure where we only represent $Q$, $K$ and the attention mask:
+
+![simple_attention.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/simple_attention.png)
```
merveenoyan, line 65: super cool

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+![naive_generate.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/naive_generate.png)
+
+To compute the attention scores of the new token, we still need the key and value projections of the previous tokens. So we need to repeat the matrix multiplication of the old tokens (in grey in the figure above) with $W_k$ and $W_v$ to retrieve a result that was already computed once before. In other terms, we are wasting compute. Let's see how we can avoid that.
+
+## KV-cache
+
+Right off the bat, we notice that the last token does not impact the attention calculation of the other tokens:
+
+![cant_see_me.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/cant_see_me.png)
```
merveenoyan, line 87: lol

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+In the figure above, only the tokens in white are computed: instead of computing the keys and values for 8 tokens, we compute them for 1. You can see that through KV caching, a lot of compute is saved.
+
+Let's be a bit more specific around the cache size, because it's a good opportunity to tour the shapes present in our model. For a model with $\mathcal L$ attention layers and $H$ attention heads with head dimension $A$, the total cache size needed to store one token will be $2 *\mathcal L * AH$ with a factor of $2$ to account for both $K$ and $V$.  
+For instance, Llama-2-7B with $\mathcal{L}=32$ layers, $H=32$ heads, and $A=128$ requires $2 \times 32 \times 128 = 8,192$ values per token per layer. With `float16` precision, this takes $2AH \times 2$ bytes $= 16$ KB in memory.
+
+KV caching is useful when we have one input token, which is a stage we call **decoding**. But it can also be useful in the **prefill** stage, when we process the initial prompt and have many input tokens. Especially when there are large initial prompts that don't fit in GPU memory all at once.
```
merveenoyan: to clarify a bit, because it sounds like you have an input of one token which defeats the purpose of the KV cache 😂
```suggestion
KV caching is useful when we want to generate the next token, which is a stage we call **decoding**. But it can also be useful in the **prefill** stage, when we process the initial prompt and have many input tokens. Especially when there are large initial prompts that don't fit in GPU memory all at once.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+Let's pretend that the available memory is very constrained, and that we can only pass $m=4$ tokens per forward pass. If we have an initial prompt with $n = 7$ tokens, we need to split it in $\lceil n /m \rceil = 2$ chunks (rounding up 7/4 = 1.75 to 2). We illustrate the example below using the same $n$ and $m$ notations:
+
+![chunked prefill.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/chunked_prefill.png)
+
+We can do that thanks to the KV cache. We store the KV states during the first prefill split, and during the second prefill split, we concatenate them left of the new KV states. We also adapt the attention mask accordingly. Visually, it looks like we split the non-chunked prefill in the middle.
```
merveenoyan:
```suggestion
We can do that thanks to the KV cache. We store the KV states during the first prefill split, and during the second prefill split, we prepend the stored KV states to the new KV states. We also adapt the attention mask accordingly. Visually, it looks like we split the non-chunked prefill in the middle.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+The key insight: cached KV states let us process the prompt incrementally without losing information.
+
+Although we showed here an example where we split the prefill into 2 chunks, chunked prefill can be used to split the prefill in any way we want, adapting flexibly to memory constraints.  
+You should by now be equipped with the tools to understand Continuous Batching.
+
+## Continuous batching
+
+So far, we have only treated the case of batch size one, i.e. we only generate tokens for one prompt at a time. But in the context of evaluation or model serving, we want to generate tokens for a large number of prompts. To increase the **throughput**, which is the number of generated tokens divided by the time it took to generate them, the best course of action is to generate tokens for a batch of several prompts.
```
merveenoyan: nit
```suggestion
So far, we have only treated the case of batch size one, i.e. we only generate tokens for one prompt at a time. But in the context of evaluation or model serving, we want to generate tokens for a large number of prompts. To increase the **throughput**, which is the number of generated tokens in a given time period (e.g. tokens per second), the best course of action is to generate tokens for a batch of several prompts.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+![batched_generation.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/batched_generation.png)
+
+where `<eos>` means "End Of Sequence", this is a special token to indicate the model has reached the end of generation.
+
+The drawback of batched generation is that if one prompt finishes generation before the other by generating an `<eos>` token, any further generated token is useless. And this goes on until the longest request of the batch finishes. Of course, we can remove the prompts that have reached an `<eos>` token from the batch and save some compute and memory, but saving resources is not the goal here: throughput is. 
+
+Instead of just removing the finished prompt from the batch, we can replace it with a prompt that's waiting for generation. We will call this **dynamic scheduling**, but it is also called dynamic batching. Dynamic scheduling is great to maintain throughput while ensuring any token generated by a forward pass is relevant. But because of the way we batched prompts together, it has a major drawback: we need a lot of padding when swapping prompts.
```
merveenoyan: TIL, very cool

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+Furthermore, practical optimizations like CUDA graphs or `torch.compile` require static tensor shapes. This forces us to pad all prompts to a fixed maximum length, dramatically increasing the padding waste. 
+
+At this point, our main problem is padding, which is a consequence of the axis we added to batch sentences together. Thus, the ideal would be to get rid of this axis entirely, a radical rethinking of batching. If we do so, the only way to batch prompts together is to concatenate them:
+
+![concatenate.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/concatenate.png)
+
+ 
+But we don't want tokens from prompt 0 to interact with the tokens of prompt 1! Luckily for us, we have a way to control how tokens interact with one another: the attention mask. How we do this is displayed below:
+
+![ragged_batching.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/ragged_batching.png)
```
merveenoyan, line 153: very nice!

continuous_batching.md
remi-or: Indeed, thanks!

continuous_batching.md
remi-or: Will be a good visual aid, adding it!

continuous_batching.md
remi-or: ty!

continuous_batching.md
remi-or: i must credit @ArthurZucker for this one :bow:

continuous_batching.md
remi-or: Not sure about this one, it feels like it's just a number of token. How do you feel about "the number of tokens generated per second" ?

continuous_batching.md
remi-or: Done, but also specified the two are the same just to keep the semantic and mathematical meanings aligned.

continuous_batching.md
merveenoyan: yep sounds good!

_blog.yml
```diff
@@ -4905,3 +4905,12 @@
     - mlx
     - community
     - open-source
+
+- local: continuous_batching
+  date: Nov 21, 2025
```
pcuenca, line 4917: Reminder to update before publishing.

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+- user: ArthurZ
+- user: mcpotato
+---
+
+# Continuous batching
+
+![Title card](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/banner.png)
+
+## Introduction
+
```
pcuenca:
```suggestion
```

I think it's not necessary to add another section header so close to the title and banner :)

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+If you've ever used Qwen, Claude, or any other AI chatbot, you've probably noticed something: it takes a while for the first word of the response to appear, and then words appear one-by-one on your screen with (hopefully) a regular and fast-paced frequency. That's because at the heart of it, all LLMs are just fancy next token predictors. During generation, first comes the **prefill** phase, where the model ingests your initial prompt to predict one new token. Then they ingest that token as well, to produce another new token: this is the **decoding** phase. The process is repeated until they feel like they have generated enough.
```
pcuenca: Do we need to introduce prefill here? I understand we want to explain it at some point, but it feels a bit tangential / distracting when all we are saying in the intro is that LLM generation is iterative and there are multiple implementation optimizations. I'd maybe just explain the sequential nature of next-token generation, and then proceed with the second paragraph.

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+This generation process is computationally expensive: it requires passing the input through billions of parameters for each token generated. To make these models practical for real-world applications, particularly when serving many users simultaneously, researchers and engineers have developed a range of efficient inference techniques.  
+One of the most impactful optimizations is **continuous batching**, which is the focus of this article. To understand how continuous batching works and why it's so effective, we'll need to build up from the fundamentals of how LLMs process tokens.
```
pcuenca: I think it might be helpful to start providing some conceptual intuition to readers so they can better frame what comes next. How about we say something like: 

```suggestion
One of the most impactful optimizations is **continuous batching**, which attempts to maximize performance by processing multiple conversations in parallel and swapping them out when they are done.

To understand how continuous batching works and why it's so effective in high-load serving scenarios, we'll build up from the fundamentals of how LLMs process tokens.
```

`performance` is not defined at all and `conversations` (instead of `sequences` or `requests`) is casual, but it carries the meaning of multiple users better. We could potentially expand this to explain that finished requests are replaced and this works great for mixed-length requests, but I think it's fine to keep it simple and just provide a glimpse of what we are talking about.

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Attention
+
+In order to understand continuous batching, it’s useful to have a clear understanding of the attention mechanism. In a Llama-like LLM, the neural network processes text by breaking it down into pieces that we call tokens, which are represented by vectors. We can think of tokens generally representing a single, simple word. For longer or rare words, several tokens may be used. For each token, the network computes a prediction of what the next token should be. It does this by applying a sequence of operations to each token.
```
pcuenca:
```suggestion
The attention mechanism is the central piece of how LLMs work. A language model processes text by breaking it down into pieces that we call tokens. We can conceptually think of "tokens" as "words", but sometimes a word might be composed of several tokens. For each token sequence, the network computes a prediction of what the next token should be.
```

(`llama-like` might be confusing for people not familiar with model ancestry 😂)
Also trying to defer the vector embeddings concept to some later point, otherwise I feel it's a piece of info readers will have to hold on to until it's used.

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+Most operations in the network are **token-wise**: each token is processed independently, and the output for a given token depends only on that token's content, not on any other tokens in the sequence. Operations like this include layer normalization or matrix multiplication. However, to create connections between words in a sentence, we need some operations where tokens can influence each other.
```
pcuenca:
```suggestion
Many operations in the network are **token-wise**: each token is processed independently, and the output for a given token depends only on that token's content, not on any other tokens in the sequence. Operations like this include layer normalization or matrix multiplication. However, to create connections between words in a sentence, we need operations where tokens can influence each other. 
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+Most operations in the network are **token-wise**: each token is processed independently, and the output for a given token depends only on that token's content, not on any other tokens in the sequence. Operations like this include layer normalization or matrix multiplication. However, to create connections between words in a sentence, we need some operations where tokens can influence each other. 
+
+This is where attention comes in. **Attention layers are the only place where different tokens interact with each other**, something that even seasoned ML practitioners sometimes overlook. Understanding how a network connects tokens together means understanding attention.
```
pcuenca:
```suggestion
This is where attention comes in. **Attention layers are the only place where different tokens interact with each other**. Understanding how a network connects tokens together means understanding attention.
```

maybe? Feel free to keep if you want to stress

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+Let's see how this works in practice. To explain the basics of attention, we'll focus on the case where we have one input prompt, which is the same as saying batch size is 1.
+
```
pcuenca:
```suggestion
```

Let's see if we can get away with not mentioning batch size for now.

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+Consider the initial prompt "I am sure this project". It is tokenized as 7 tokens: `[<bos>, I, am, sure, this, pro, ject]`. The `<bos>` token is a special token that is added at the start of the prompt (BoS stands for Beginning of Sequence).
```
pcuenca:
```suggestion
Consider the initial prompt `I am sure this project`, tokenized as 7 tokens: `[<bos>, I, am, sure, this, pro, ject]`. The `<bos>`, or "Beginning of Sequence", is a special token we add at the start of the prompt to tell the language model that a new conversation starts here.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+This is where attention comes in. **Attention layers are the only place where different tokens interact with each other**, something that even seasoned ML practitioners sometimes overlook. Understanding how a network connects tokens together means understanding attention.
+
+Let's see how this works in practice. To explain the basics of attention, we'll focus on the case where we have one input prompt, which is the same as saying batch size is 1.
+
+Consider the initial prompt "I am sure this project". It is tokenized as 7 tokens: `[<bos>, I, am, sure, this, pro, ject]`. The `<bos>` token is a special token that is added at the start of the prompt (BoS stands for Beginning of Sequence).  
+
+Those incoming tokens form a tensor we call $x$, with shape $\left[1, n , d \right]$. This is because batch size is $1$, $n$ is the number of tokens and $d$ is the hidden dimension.
```
pcuenca:
```suggestion
Each token is represented inside the network with a vector of length `d` (the _hidden dimension_). Therefore, the seven incoming tokens form a tensor \\(x\\) with shape \\(\left[1, 7, d \right]\\). `1` is the number of sequences, or batch size, which is just one in our case. `7` is the sequence length, and `d` is the hidden dimension, or the size of each token representation. Going forward, we'll use \\(n\\) instead of `7` as the sequence length.
```

Note that inline latex does not work with `$`, we need to use `\\(` and `\\)`.

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+Let's see how this works in practice. To explain the basics of attention, we'll focus on the case where we have one input prompt, which is the same as saying batch size is 1.
+
+Consider the initial prompt "I am sure this project". It is tokenized as 7 tokens: `[<bos>, I, am, sure, this, pro, ject]`. The `<bos>` token is a special token that is added at the start of the prompt (BoS stands for Beginning of Sequence).  
+
+Those incoming tokens form a tensor we call $x$, with shape $\left[1, n , d \right]$. This is because batch size is $1$, $n$ is the number of tokens and $d$ is the hidden dimension. 
+
+Those tokens $x$ are then projected by three matrices: the query projection $W_q$,  the key projection $W_k$ and the value projection $W_v$. This produces three tensors $Q$, $K$ and $V$, all of shape $\left[1, n , A \right]$ where $A$ is the dimension of an attention head. We call them respectively the **query, key and value states.** This is represented on the left in the figure below.
```
pcuenca:
```suggestion
Tensor \\(x\\), which represents our input sequence, is then projected by three matrices: the query projection \\(W_q\\),  the key projection \\(W_k\\), and the value projection \\(W_v\\). This produces three tensors \\(Q\\), \\(K\\) and \\(V\\), all of shape \\(\left[1, n, A \right]\\) where \\(A\\) is the dimension of the so-called attention head. We call them the **query, key, and value states,**, respectively. This is represented on the left in the figure below.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+![proj_and_mul.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/proj_and_mul.png)
+
+Next, tensors $Q$ and $K$ are multiplied together to measure similarity between tokens, producing a tensor of shape $\left[ 1, n , n \right]$. This is why we say that attention has quadratic complexity in sequence length. Computing $QK^T$ requires $\mathcal{O} \left( n^2 d \right)$ operations, so the cost is a square of $n$ the sequence length. It is represented on the right in the figure above.
```
pcuenca: (I'm not fixing inline latex going forward, can you please do it @remi-or? 🙏)

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+To compute the attention scores of the new token, we still need the key and value projections of the previous tokens. So we need to repeat the matrix multiplication of the old tokens (in grey in the figure above) with $W_k$ and $W_v$ to retrieve a result that was already computed once before. In other terms, we are wasting compute. Let's see how we can avoid that.
+
+## KV-cache
+
+Right off the bat, we notice that the last token does not impact the attention calculation of the other tokens:
+
+![cant_see_me.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/cant_see_me.png)
+
+This follows the idea of the causal mask: since "*will*" comes after all previous tokens, it does not change their attention calculation.
+For text, causal attention is by far the most common. We will focus on that case from now on. But non-causal attention also exists, especially when dealing with images.
```
pcuenca:
```suggestion
For text generation, causal attention is by far the most common, so we will focus on that case from now on. Keep in mind that non-causal attention schemes can also be used, especially when dealing with images.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+Considering we only need the next-token prediction for the "*will*" token, we can simplify the attention mechanism by only computing the output for this token.
+
+Moreover, we already computed the $K$ and $V$ states for the tokens "*\<bos\>*", … ,  "*ject*" during the previous forward pass: if they have been stored, we do not need to recompute them again. This is the **KV cache**: the list of key and value states created during generation. It essentially allows one to reduce the compute cost of generating token $n+1$ from $\mathcal{O} \left( n^2 \right)$ to $\mathcal{O} \left( n \right)$ by avoiding recomputation of key and value projections, while paying a memory cost of  $\mathcal{O} \left( n \right)$.
+
+![kv_cache.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/kv_cache.png)
```
pcuenca, line 95: Somewhere in this section we could potentially link to https://huggingface.co/blog/not-lain/kv-caching or https://huggingface.co/blog/kv-cache for people that want to dive deeper or see other visualizations.

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+- user: ror
+- user: ArthurZ
+- user: mcpotato
+---
+
+# Continuous batching
+
+![Title card](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/banner.png)
+
+## Introduction
```
merveenoyan:
```suggestion
```
no need for this imo

continuous_batching.md
merveenoyan, line 15: maybe drop a small TL;DR here so people know what you will be talking about before reading a lengthy post 🙌🏻

continuous_batching.md
merveenoyan: these (and rest of the LaTeX too!) aren't rendered properly, just checked on hf.co/new-blog

syntax guide says: 

`You can mix latex with markdown. We pick $$YOUR LATEX$$ (for display) and \\(YOUR LATEX\\) (for inline) as delimiters. (example: \\( a=b+c \\) is my equation)`

Screenshot 2025-11-25 at 11 48 06

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+![masking_and_softmax.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/masking_and_softmax.png)
+
+Finally, after applying the attention mask, we take a token-wise softmax (which is the same as saying a row-wise softmax) and multiply the result by the value projection $V$ to get the output of one attention head, of shape $\left[ 1, n , A \right]$. We offer a visual summary of the whole process in the following figure.
+
+![attention.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/attention.png)
+
```
pcuenca, line 50:
```suggestion

The process we just described, where we take an input sequence, pass it through multiple attention layers and compute a score for the next token, is called "prefill". This is because, as we'll see in a moment, much of the computation we performed can be cached and reused – hence, we are _prefilling_ the cache. Thanks to the use of the cache, sequence generation can proceed from this moment on by handling one token at a time and expanding the cache, making successive tokens much faster to generate than the initial full-sequence computation.
```

Trying to provide some breadcrumbs and introducing prefill here, I hope it makes sense 🤞

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+We are going to use a lot of attention visualization in this post, so to simplify things, we are going to condense the figure above just a bit.
+
+**Why this matters:** In continuous batching, $Q$, $K$, and $V$ can have different numbers of tokens because we're processing different stages together (prefill and decode). Let's say $Q$ has shape $\left[1, n_Q , A \right]$, $K$ has shape $\left[ 1, n_K , A \right]$, and $V$ has shape $\left[ 1, n_V , A \right]$.
```
pcuenca:
```suggestion
**Why this matters:** In continuous batching, $Q$, $K$, and $V$ can have different numbers of tokens because we will be processing different stages together (prefill and decode), as we'll see. So to make things general, let's say $Q$ has shape $\left[1, n_Q , A \right]$, $K$ has shape $\left[ 1, n_K , A \right]$, and $V$ has shape $\left[ 1, n_V , A \right]$.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+![kv_cache.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/kv_cache.png)
+
+In the figure above, only the tokens in white are computed: instead of computing the keys and values for 8 tokens, we compute them for 1. You can see that through KV caching, a lot of compute is saved.
+
+Let's be a bit more specific around the cache size, because it's a good opportunity to tour the shapes present in our model. For a model with $\mathcal L$ attention layers and $H$ attention heads with head dimension $A$, the total cache size needed to store one token will be $2 *\mathcal L * AH$ with a factor of $2$ to account for both $K$ and $V$.
```
pcuenca:
```suggestion
Let's be a bit more specific about the cache size, because it's a good opportunity to tour the shapes present in our model. For a model with $\mathcal L$ attention layers and $H$ attention heads with head dimension $A$, the total cache size needed to store one token will be $2 *\mathcal L * AH$ with a factor of $2$ to account for both $K$ and $V$.  
```

continuous_batching.md
remi-or: Removed it, thanks

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+Let's pretend that the available memory is very constrained, and that we can only pass $m=4$ tokens per forward pass. If we have an initial prompt with $n = 7$ tokens, we need to split it in $\lceil n /m \rceil = 2$ chunks (rounding up 7/4 = 1.75 to 2). We illustrate the example below using the same $n$ and $m$ notations:
+
+![chunked prefill.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/chunked_prefill.png)
+
+We can do that thanks to the KV cache. We store the KV states during the first prefill split, and during the second prefill split, we prepend the stored KV states to the new KV states. We also adapt the attention mask accordingly. Visually, it looks like we split the non-chunked prefill in the middle.
+
+The key insight: cached KV states let us process the prompt incrementally without losing information.
+
+Although we showed here an example where we split the prefill into 2 chunks, chunked prefill can be used to split the prefill in any way we want, adapting flexibly to memory constraints.  
+You should by now be equipped with the tools to understand Continuous Batching.
```
pcuenca:
```suggestion

We are now finally equipped with all the tools we need to understand Continuous Batching.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Continuous batching
+
+So far, we have only treated the case of batch size one, i.e. we only generate tokens for one prompt at a time. But in the context of evaluation or model serving, we want to generate tokens for a large number of prompts. To increase the **throughput**, which is the number of tokens generated per second, the best course of action is to generate tokens for a batch of several prompts.
```
pcuenca:
```suggestion
In our previous examples we have only considered the case of batch size one, i.e. we only generate tokens for one prompt at a time. In the context of evaluation or model serving, we want to generate tokens for a large number of prompts. To increase the **throughput**, which is the number of tokens generated per second, the best course of action is to generate tokens in parallel for a batch of several prompts.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+To batch prompts together, the naive way is to add an axis to both input tensors, which will be the batch axis. This way we can pass two prompts and two attention masks, one for each. However, this comes with a constraint on the shape of the inputs: we need all prompts to have the same length (since tensors must be rectangular). To achieve this, we usually add padding on the left so the new token prediction always comes from the rightmost token. We also modify the attention mask of each prompt accordingly. This is shown below:
```
pcuenca:
```suggestion
To batch prompts together, the naive way is to add an axis to both input tensors: token sequence and attention mask. However, this comes with a constraint on the shape of the inputs: we need all prompts to have the same length, because tensors must be rectangular. To achieve this, we usually add padding on the left so the new token prediction always comes from the rightmost token. We also modify the attention mask of each prompt accordingly, as shown below:
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+![padding.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/padding.png)
+
+where the padding tokens `<pad>` are coloured in orange. Then we can perform the forward pass as we used to, with the added dimension of the batch size. This is called **batched generation**: efficient for same-length prompts, but wasteful when lengths vary. It is illustrated below:
+
+![batched_generation.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/batched_generation.png)
```
pcuenca, line 132: We can potentially point out that the top section of the figure is prefill.

Also, should the batch sizes below be `[2, 1, d]` rather than `[2, 2, d]` (I'm probably confused)

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+where `<eos>` means "End Of Sequence", this is a special token to indicate the model has reached the end of generation.
```
pcuenca:
```suggestion
where `<eos>` means "End Of Sequence", this is a special token to indicate the model has reached the end of generation for the corresponding sequence.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+The drawback of batched generation is that if one prompt finishes generation before the other by generating an `<eos>` token, any further generated token is useless. And this goes on until the longest request of the batch finishes. Of course, we can remove the prompts that have reached an `<eos>` token from the batch and save some compute and memory, but saving resources is not the goal here: throughput is.
```
pcuenca:
```suggestion
The drawback of batched generation is that if one prompt finishes generation before the other one by generating an `<eos>` token, all further generated token are useless. And this goes on until the longest request of the batch finishes. Of course, we can remove the prompts that have reached an `<eos>` token from the batch and save some compute and memory, but saving resources is not the goal here: throughput is. 
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+![batched_generation.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/batched_generation.png)
+
+where `<eos>` means "End Of Sequence", this is a special token to indicate the model has reached the end of generation.
+
+The drawback of batched generation is that if one prompt finishes generation before the other by generating an `<eos>` token, any further generated token is useless. And this goes on until the longest request of the batch finishes. Of course, we can remove the prompts that have reached an `<eos>` token from the batch and save some compute and memory, but saving resources is not the goal here: throughput is. 
+
+Instead of just removing the finished prompt from the batch, we can replace it with a prompt that's waiting for generation. We will call this **dynamic scheduling**, but it is also called dynamic batching. Dynamic scheduling is great to maintain throughput while ensuring any token generated by a forward pass is relevant. But because of the way we batched prompts together, it has a major drawback: we need a lot of padding when swapping prompts.
```
pcuenca:
```suggestion
Instead of just removing the finished prompt from the batch, we can replace it with a prompt that's waiting for generation. We will call this **dynamic scheduling**, or dynamic batching. Dynamic scheduling is great to maintain throughput while ensuring any token generated by a forward pass is relevant. But because of the way we batched prompts together, it has a major drawback: we need a lot of padding when swapping prompts.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+![dynamic_batching.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/dynamic_batching.png)
```
pcuenca, line 140: Maybe point out that the newly-inserted sequence needs to go through a prefill phase while the others are decoding one token at a time, so we need to add as many tokens (per sequence) as the new sequence length to make shapes match.

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+The problem becomes even worse when batch size increases and initial prompts are long. The padding cost grows quadratically with both batch size and prompt length. If we have a batch of $B$ prompts that are in decoding phase and one finishes, dynamically introducing a prompt of $n$ initial tokens in the batch requires $(n-1)(B-1)$ padding tokens. For instance, with $B=8$ and $n=100$, we'd need $99 \times 7 = 693$ padding tokens!
+
+Furthermore, practical optimizations like CUDA graphs or `torch.compile` require static tensor shapes. This forces us to pad all prompts to a fixed maximum length, dramatically increasing the padding waste.
```
pcuenca, line 144: ah, good point

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+ 
+But we don't want tokens from prompt 0 to interact with the tokens of prompt 1! Luckily for us, we have a way to control how tokens interact with one another: the attention mask. How we do this is displayed below:
+
+![ragged_batching.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/ragged_batching.png)
+
+Although we use different tints of green to illustrate the different parts of the attention mask, this is still a boolean mask with only greens for `True` and white for `False`. 
+This way of batching prompts together is called **ragged batching** (because sequence lengths are 'ragged' or uneven), and it offers the benefit of added throughput without introducing the need for padding tokens.
+
+In the figure above, we batch two full prompts together, although there is no limit to how many prompts can be batched together using ragged batching. The only limit is $m$, the number of tokens we can fit in a batch, with $m$ depending on the available memory on the GPU.
```
pcuenca:
```suggestion
In the figure above, we concatenate two prompts together with ragged batching, but we can batch as many as memory allows. The only limit is $m$, the number of tokens we can fit in a batch, with $m$ depending on the available memory on the GPU. 
```

continuous_batching.md
remi-or: Added this:
*TL;DR: in this blog post, starting from attention mechanisms and KV caching, we derive continuous batching by optimizing for throughput.*

continuous_batching.md
remi-or: A big blunder on my part! Forgot to change the delimiters. Thanks a lot for catching this!

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+Although we use different tints of green to illustrate the different parts of the attention mask, this is still a boolean mask with only greens for `True` and white for `False`. 
+This way of batching prompts together is called **ragged batching** (because sequence lengths are 'ragged' or uneven), and it offers the benefit of added throughput without introducing the need for padding tokens.
+
+In the figure above, we batch two full prompts together, although there is no limit to how many prompts can be batched together using ragged batching. The only limit is $m$, the number of tokens we can fit in a batch, with $m$ depending on the available memory on the GPU. 
+
+For instance, here is one batching strategy used in practice to maximize throughput:
+- We try to always reach $m$ tokens per batch
+- We first add all the prompts in decoding phase to the batch, each accounting for 1 token
+- We fill the remaining space with prefill phase prompts, relying on the flexibility of chunked prefill to exactly fill the batch as desired
```
pcuenca: Perhaps this detail is not necessary, as we are now combining decoding + prefill generations in a way that doesn't feel obvious from what's been discussed so far (we just concatenated two input prompts).

continuous_batching.md
remi-or: and thanks for the tip of using hf.co/new-blog

continuous_batching.md
pcuenca: Or maybe we can reword to work our way towards CB:

```suggestion
Ragged batching is one of the key components of continuous batching. Note that we don't need to always use full prompts to compose our batch. Instead, we can apply a dynamic strategy where we:
- Try to always reach our memory budget of $m$ tokens per batch
- Add all the prompts in decoding phase to the batch, each accounting for 1 token
- Fill the remaining space with prefill phase prompts, relying on the flexibility of chunked prefill to split inputs as needed
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+This way of batching prompts together is called **ragged batching** (because sequence lengths are 'ragged' or uneven), and it offers the benefit of added throughput without introducing the need for padding tokens.
+
+In the figure above, we batch two full prompts together, although there is no limit to how many prompts can be batched together using ragged batching. The only limit is \\( m \\), the number of tokens we can fit in a batch, with \\( m \\) depending on the available memory on the GPU. 
+
+For instance, here is one batching strategy used in practice to maximize throughput:
+- We try to always reach \\( m \\) tokens per batch
+- We first add all the prompts in decoding phase to the batch, each accounting for 1 token
+- We fill the remaining space with prefill phase prompts, relying on the flexibility of chunked prefill to exactly fill the batch as desired
+
+Furthermore, we can also use dynamic scheduling to remove finished prompts from the batch whenever they finish. This combination of ragged batching and dynamic scheduling is called **continuous batching**, and it's the technique that powers modern LLM serving systems.
```
pcuenca:
```suggestion
Dynamic scheduling is the final piece that contributes to the _continuous batching_ technique: we remove finished prompts from the batch as soon as they are done, and replace them with new chunked prompts that correspond to incoming requests.

This combination of ragged batching and dynamic scheduling is called **continuous batching**, and it's the technique that powers modern LLM serving systems.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+ 
+But we don't want tokens from prompt 0 to interact with the tokens of prompt 1! Luckily for us, we have a way to control how tokens interact with one another: the attention mask. How we do this is displayed below:
+
+![ragged_batching.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/ragged_batching.png)
+
+Although we use different tints of green to illustrate the different parts of the attention mask, this is still a boolean mask with only greens for `True` and white for `False`. 
+This way of batching prompts together is called **ragged batching** (because sequence lengths are 'ragged' or uneven), and it offers the benefit of added throughput without introducing the need for padding tokens.
+
+In the figure above, we batch two full prompts together, although there is no limit to how many prompts can be batched together using ragged batching. The only limit is \\( m \\), the number of tokens we can fit in a batch, with \\( m \\) depending on the available memory on the GPU.
```
pcuenca:
```suggestion
In the figure above, we use ragged batching to combine two full prompts together, but we can batch as many as memory allows. The only limit is \\( m \\), the number of tokens we can fit in a batch, with \\( m \\) depending on the available memory on the GPU. 
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+Although we use different tints of green to illustrate the different parts of the attention mask, this is still a boolean mask with only greens for `True` and white for `False`. 
+This way of batching prompts together is called **ragged batching** (because sequence lengths are 'ragged' or uneven), and it offers the benefit of added throughput without introducing the need for padding tokens.
+
+In the figure above, we batch two full prompts together, although there is no limit to how many prompts can be batched together using ragged batching. The only limit is \\( m \\), the number of tokens we can fit in a batch, with \\( m \\) depending on the available memory on the GPU. 
+
+For instance, here is one batching strategy used in practice to maximize throughput:
+- We try to always reach \\( m \\) tokens per batch
+- We first add all the prompts in decoding phase to the batch, each accounting for 1 token
+- We fill the remaining space with prefill phase prompts, relying on the flexibility of chunked prefill to exactly fill the batch as desired
```
pcuenca: We can maybe use this as an opportunity to work our way towards CB:

```suggestion
Ragged batching is one of the key components of continuous batching. To maximize throughput, we can combine prefill and decoding sequences following an algorithm like this:
- We try to always reach our memory budget of \\( m \\) tokens per batch
- We first add all the prompts in decoding phase to the batch, each accounting for 1 token
- We fill the remaining space with prefill phase prompts, relying on the flexibility of chunked prefill to split inputs as needed
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+## Conclusion
+
+Continuous batching combines three key techniques to maximize throughput in LLM serving:
+1. **KV caching** to avoid recomputing past token representations
+2. **Chunked prefill** to handle variable-length prompts within memory constraints  
+3. **Ragged batching with dynamic scheduling** to eliminate padding waste and keep the GPU fully utilized
+
+By removing the batch dimension and using attention masks to control token interactions, continuous batching allows mixing prefill and decode phases in the same batch, dramatically improving efficiency for serving multiple requests. This is why services like ChatGPT can handle thousands of concurrent users efficiently. 
+
+In the next article in this series, we'll explore efficient KV cache management. If you'd like to see a deep dive on other continuous batching topics, please let us know in the comments!
```
pcuenca: Perhaps we could mention "paged attention" as one technique that's usually associated to CB, but not really a part of it, maybe as a potential future deep dive? (or linking to some other resource)

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+This is where attention comes in. **Attention layers are the only place where different tokens interact with each other**, something that even seasoned ML practitioners sometimes overlook. Understanding how a network connects tokens together means understanding attention.
+
+Let's see how this works in practice. To explain the basics of attention, we'll focus on the case where we have one input prompt, which is the same as saying batch size is 1.
+
+Consider the initial prompt "I am sure this project". It is tokenized as 7 tokens: `[<bos>, I, am, sure, this, pro, ject]`. The `<bos>` token is a special token that is added at the start of the prompt (BoS stands for Beginning of Sequence).  
+
+Those incoming tokens form a tensor we call \\( x \\), with shape \\( \left[1, n , d \right] \\). This is because batch size is \\( 1 \\), \\( n \\) is the number of tokens and \\( d \\) is the hidden dimension.
```
pcuenca:
```suggestion
Each token is represented inside the network with a vector of length `d` (the _hidden dimension_). Therefore, the seven incoming tokens form a tensor \\(x\\) with shape \\(\left[1, 7, d \right]\\). `1` is the number of sequences, or batch size, which is just one in our case. `7` is the sequence length, and `d` is the hidden dimension, or the size of each token representation. Going forward, we'll use \\(n\\) instead of `7` as the sequence length.
 ```

File: continuous_batching.md [hunk 53]
```diff
@@ -0,0 +1,173 @@
...
+
+Let's see how this works in practice. To explain the basics of attention, we'll focus on the case where we have one input prompt, which is the same as saying batch size is 1.
+
+Consider the initial prompt "I am sure this project". It is tokenized as 7 tokens: `[<bos>, I, am, sure, this, pro, ject]`. The `<bos>` token is a special token that is added at the start of the prompt (BoS stands for Beginning of Sequence).  
+
+Those incoming tokens form a tensor we call \\( x \\), with shape \\( \left[1, n , d \right] \\). This is because batch size is \\( 1 \\), \\( n \\) is the number of tokens and \\( d \\) is the hidden dimension. 
+
+Those tokens \\( x \\) are then projected by three matrices: the query projection \\( W_q \\),  the key projection \\( W_k \\) and the value projection \\( W_v \\). This produces three tensors \\( Q \\), \\( K \\) and \\( V \\), all of shape \\( \left[1, n , A \right] \\) where \\( A \\) is the dimension of an attention head. We call them respectively the **query, key and value states.** This is represented on the left in the figure below.
```
pcuenca:
```suggestion
Input tensor \\( x \\) is then projected by three matrices: the query projection \\( W_q \\),  the key projection \\( W_k \\) and the value projection \\( W_v \\). This produces three tensors \\( Q \\), \\( K \\) and \\( V \\), all of shape \\( \left[1, n , A \right] \\), where \\( A \\) is the dimension of the attention head. We call them the **query, key and value states,** respectively. This is represented on the left in the figure below.
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+
+![kv_cache.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/kv_cache.png)
+
+In the figure above, only the tokens in white are computed: instead of computing the keys and values for 8 tokens, we compute them for 1. You can see that through KV caching, a lot of compute is saved.
+
+Let's be a bit more specific around the cache size, because it's a good opportunity to tour the shapes present in our model. For a model with \\( \mathcal L \\) attention layers and \\( H \\) attention heads with head dimension \\( A \\), the total cache size needed to store one token will be \\( 2 *\mathcal L * AH \\) with a factor of \\( 2 \\) to account for both \\( K \\) and \\( V \\).
```
pcuenca:
```suggestion
Let's be a bit more specific about the cache size, because it's a good opportunity to examine the shapes present in our model. For a model with \\( \mathcal L \\) attention layers and \\( H \\) attention heads with head dimension \\( A \\), the total cache size needed to store one token will be \\( 2 *\mathcal L * AH \\) with a factor of \\( 2 \\) to account for both \\( K \\) and \\( V \\).  
```

continuous_batching.md
```diff
@@ -0,0 +1,173 @@
...
+![masking_and_softmax.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/masking_and_softmax.png)
+
+Finally, after applying the attention mask, we take a token-wise softmax (which is the same as saying a row-wise softmax) and multiply the result by the value projection \\( V \\) to get the output of one attention head, of shape \\( \left[ 1, n , A \right] \\). We offer a visual summary of the whole process in the following figure.
+
+![attention.png](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_batching/attention.png)
+
+
+We are going to use a lot of attention visualization in this post, so to simplify things, we are going to condense the figure above just a bit.
+
+**Why this matters:** In continuous batching, \\( Q \\), \\( K \\), and \\( V \\) can have different numbers of tokens because we're processing different stages together (prefill and decode). Let's say \\( Q \\) has shape \\( \left[1, n_Q , A \right] \\), \\( K \\) has shape \\( \left[ 1, n_K , A \right] \\), and \\( V \\) has shape \\( \left[ 1, n_V , A \right] \\).
```
pcuenca:
```suggestion
**Why this matters:** In continuous batching, \\( Q \\), \\( K \\), and \\( V \\) can have different numbers of tokens because, as we'll see, we'll be processing different stages (prefill and decode) at the same time. To make it more general, let's say \\( Q \\) has shape \\( \left[1, n_Q , A \right] \\), \\( K \\) has shape \\( \left[ 1, n_K , A \right] \\), and \\( V \\) has shape \\( \left[ 1, n_V , A \right] \\).
```

continuous_batching.md
remi-or: Agreed, it's a good glimpse into the future.

continuous_batching.md
remi-or: Indeed, it might be a little too much

continuous_batching.md
remi-or: Agreed we may not need to mention batch size, but I kind of want to specify that there is only one input prompt. Will try to do a more succinct transition!

continuous_batching.md
remi-or: Better way to introduce `n`, thanks for the suggestion!

continuous_batching.md
remi-or: Done!

continuous_batching.md
remi-or: Not confused, you are right! Thanks for the catch.

continuous_batching.md
remi-or: Good point, done

continuous_batching.md
remi-or: added `through __paged attention__` after `we'll explore efficient KV cache management`

continuous_batching.md
remi-or: I moved the breadcrumbs a bit but kept them, thanks!

continuous_batching.md
remi-or: I agree that it was a bit shoehorned in. I stuck to the auto-regressive explanation and direct transition to 2nd paragraph

PR #3207, opened 2025-12-03
blog post on hf skills + fine-tuning

burtenshaw:
This is a blog on training an LLM with claude (or any cli agent) with skills and hf mcp. It uses [hf skills](https://github.com/huggingface/skills) and hf mcp.

❓I wasn't 100% sure I should use this blog, or an org article. Happy to change if there are better options.

### DIFF
--- a/hf-skills-training.md
+++ b/hf-skills-training.md
@@ -141,6 +141,12 @@ The model will be pushed to Hub automatically. Should I submit?
 
 This is your chance to adjust anything. Change the output repo name, pick different hardware, or ask Claude to modify training parameters. Once you approve, the agent submits the job.
 
+For example, you can ask the agent to try a test run:
+
+```
+Do a quick test run on 100 examples.
+```
+
 ### Track Progress
 
 After submission, you get job details:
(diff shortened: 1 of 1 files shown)


hf-skills-training.md
```diff
@@ -0,0 +1,361 @@
...
+For **medium models (3-7B)**, you need `a10g-large` or `a100-large` with LoRA. Full fine-tuning doesn't fit, but LoRA makes these very trainable. Budget $15-40 for production.
+
+For **large models (7B+)**, this HF skills job is not suitable. 
+
+### Demo vs Production
+
+When testing a workflow, start small:
+
+```
+Do a quick demo run on Qwen-0.5B with 100 examples.
```
merveenoyan: which skill does this call? demo run is a bit vague (also does it automatically pick dataset when told like this?)

hf-skills-training.md
```diff
@@ -0,0 +1,361 @@
...
+```
+
+Th coding agent configures minimal training—enough to verify your pipeline works without real cost.
+
+For production, be explicit:
+
+```
+Train for production on the full dataset.
```
merveenoyan: does it directly do SFT when we say "train"?

hf-skills-training.md
burtenshaw: It's relying on the context of previous messages, once the agent has returned with its proposal. But that's not clear. I'll change these two to be self-contained.

burtenshaw: I think [this](https://github.com/huggingface/blog/pull/3207/commits/1706360567a8c8e1341af2f5933b454b161b5544) makes it clearer what's happening.

PR #3233, opened 2025-12-17
[Add] Tokenization Blog Post

ariG23498:
This PR adds the new tokenization blog post.

CC: https://github.com/itazap @ArthurZucker @sergiopaniego

### DIFF
--- a/tokenizers.md
+++ b/tokenizers.md
@@ -235,7 +235,7 @@ The `transformers` library organizes tokenizers into a class hierarchy. At the t
 The base class handles functionality that doesn't depend on the tokenization backend:
 
 * **Special token properties.** Properties like `bos_token`, `eos_token`, `pad_token`, and `unk_token` are defined here. These properties provide access to the special tokens that models use to mark sequence boundaries and handle unknown inputs.  
-* **Encoding interface.** The `__call__` method, `encode`, `encode_plus`, and `batch_encode_plus` methods are defined here. These methods accept text input and return token IDs along with attention masks and other metadata.  
+* **Encoding interface.** The `__call__` method, `encode`, and `encode_plus` methods are defined here. These methods accept text input and return token IDs along with attention masks and other metadata.  
 * **Decoding interface.** The `decode` and `batch_decode` methods convert token IDs back to text.  
 * **Serialization.** The `save_pretrained` and `from_pretrained` methods handle downloading the correct files, reading information, saving tokenizers to disk etc.  
 * **Chat template support.** The `apply_chat_template` method lives here, formatting conversations according to Jinja templates stored in the tokenizer configuration.


tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+
+## Table of Contents
+
+- [What is Tokenization?](#what-is-tokenization)
+- [The Tokenization Pipeline](#the-tokenization-pipeline)
+- [Tokenization Algorithms](#tokenization-algorithms)
+- [Accessing `tokenizers` through `transformers`](#accessing-tokenizers-through-transformers)
+- [The Tokenizer Class Hierarchy in `transformers`](#the-tokenizer-class-hierarchy-in-transformers)
+- [`AutoTokenizer` Automatically Selects the Correct Tokenizer Class](#autotokenizer-automatically-selects-the-correct-tokenizer-class)
+- [v5 Seperates Tokenizer Architecture from Trained Vocab](#v5-separates-tokenizer-architecture-from-trained-vocab)
```
sergiopaniego:
```suggestion
- [v5 Separates Tokenizer Architecture from Trained Vocab](#v5-separates-tokenizer-architecture-from-trained-vocab)
```

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+- [Summary](#summary)
+
+> [!TIP]
+> For experts: If you're already familiar with the concepts and want to understand the changes in v5, go to [v5 Seperates Tokenizer Architecture from Trained Vocab](#v5-separates-tokenizer-architecture-from-trained-vocab)
```
sergiopaniego:
```suggestion
> For experts: If you're already familiar with the concepts and want to understand the changes in v5, go to [v5 Separates Tokenizer Architecture from Trained Vocab](#v5-separates-tokenizer-architecture-from-trained-vocab)
```

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+
+print(f"{tokenizer._tokenizer.decoder=}")
+# Sequence(decoders=[Replace(...), ByteFallback(), Fuse()])
+```
+
+## Tokenization algorithms
+
+The following algorithms dominate modern language model tokenizers:
+
+1. **Byte Pair Encoding (BPE)** iteratively merges the most frequent character pairs. This algorithm is deterministic and widely used. (Read mre about [BPE](https://huggingface.co/learn/llm-course/en/chapter6/5))
```
sergiopaniego:
```suggestion
1. **Byte Pair Encoding (BPE)** iteratively merges the most frequent character pairs. This algorithm is deterministic and widely used. (Read more about [BPE](https://huggingface.co/learn/llm-course/en/chapter6/5))
```

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+> [!NOTE]
+> More than adding, `transformers` defines the API for tokenization that is now most commonly used is the entire ML! (`encode`, `decode`, `convert_tokens_to_ids`, etc.)
+
+## The tokenizer class hierarchy in transformers
+
+The `transformers` library organizes tokenizers into a class hierarchy. At the top sits a base class that defines the common interface. Below it, backend classes handle the actual tokenization using different engines. At the bottom, model-specific classes configure the backends for particular models.
+
+| ![class hierarchy](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/tokenizers/hierarchy.png) |
+| :--: |
+| The class hierarchy for tokenziers inside transformers |
```
sergiopaniego:
```suggestion
| The class hierarchy for tokenizers inside transformers |
```

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+tokenizer = AutoTokenizer.from_pretrained("HuggingFaceTB/SmolLM3-3B")
+
+text = "Hello world"
+tokens = tokenizer(text)
+
+print(tokens["input_ids"])
+# [9906, 1917]
+
+print(tokenizer.convert_ids_to_tokens(tokens["input_ids"]))
+# ['Hello', 'Ġworld']
```
merveenoyan, line 62: the Ğ is due to a bug we recently discovered btw but not super sure 😅 it looks confusing for the users

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+```
+
+A **token** is the smallest string unit the model sees. It can be a character, word, or subword chunk like "play" or "##ing". The **vocabulary** maps each unique token to the token ID.
```
merveenoyan: can you explain or remove the hash here it could be confusing for a beginner imo

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+tokenizer = AutoTokenizer.from_pretrained("HuggingFaceTB/SmolLM3-3B")
+print(tokenizer.vocab)
+
+# {'ÎĹÎľ': 106502, 'ĠPeel': 89694, '.languages': 91078, ...}
+```
+
+Two numbers matter:
+
+* **Vocabulary size of the tokenizer**: how many unique tokens exist in the vocabulary of the tokenizer  
+* **Context length of the model**: how many tokens the model was trained to attend to at once, and can process in a single forward pass.
```
merveenoyan: this part is a bit hard for readability: how many tokens the model was trained to attend to at once
I'd honestly do: how many tokens a model can attend to in a single forward pass.

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+
+A good tokenizer *compresses* text into the smallest amount of tokens. Fewer tokens means more usable context without increasing model size.Training a tokenizer boils down to finding the best compression rules for your datasets. For example if you work on Chinese corpus you can sometimes find [very nice surprises 😉](https://x.com/suchenzang/status/1697862650053660721).
```
merveenoyan: idk if you'll talk about BPE later but I'd also mention something around representation of text in frequency aspect with least tokens possible!

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+# ['Hello', 'Ġworld']
+```
+
+The output is raw tokenization. You get token IDs and the string pieces they correspond to. Nothing more.
+
+Now consider what's missing. The `SmolLM3-3B` is a *conversational model*. When you interact with it, you typically structure your input as a conversation with roles like "user" and "assistant." The model expects special formatting tokens to indicate these roles. The raw `tokenizers` library has no concept of any of this.
+
+### How do you bridge the gap between raw tokenization and model requirements?
+
+The `transformers` library bridges this gap. The library is primarily known as a model repository (containing model definitions), but it also provides a tokenizer abstraction layer that wraps the raw `tokenizers` backend and adds model-aware functionality.
```
merveenoyan:
```suggestion
The `transformers` library bridges this gap. The library is primarily known as a model definition library, but it also provides a tokenizer abstraction layer that wraps the raw `tokenizers` backend and adds model-aware functionality.
```

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+messages = [{"role": "user", "content": prompt}]
+text = tokenizer.apply_chat_template(
+    messages,
+    tokenize=False,
+    add_generation_prompt=True,
+)
+
+print(text)
+
+# <|im_start|>system
```
merveenoyan, line 201: maybe explain special tokens a bit before this. also interesting SmolLM3 tokenizer has image start token, didn't know! (maybe it inherited some multimodal llama)

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+# <|im_start|>user
+# Give me a brief explanation of gravity in simple terms.<|im_end|>
+# <|im_start|>assistant
+
+model_inputs = tokenizer([text], return_tensors="pt")
+```
+
+The `transformers` tokenizer adds everything the raw library lacks:
+
+* **Chat template application.** The `apply_chat_template` method formats conversations according to the model's expected format, inserting the correct special tokens and delimiters.
```
merveenoyan, line 214: I'd explain a bit above special tokens and need for templates if we go for beginners

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+model_inputs = tokenizer([text], return_tensors="pt")
+```
+
+The `transformers` tokenizer adds everything the raw library lacks:
+
+* **Chat template application.** The `apply_chat_template` method formats conversations according to the model's expected format, inserting the correct special tokens and delimiters.  
+* **Automatic special token insertion.** Beginning-of-sequence and end-of-sequence tokens are added where the model expects them.  
+* **Truncation to context length.** You can specify `truncation=True` and the tokenizer will respect the model's maximum sequence length.  
+* **Batch encoding with padding.** Multiple inputs can be padded to the same length with the correct padding token and direction.  
+* **Return format options.** You can request PyTorch tensors (`return_tensors="pt"`), TensorFlow tensors, or NumPy arrays.
```
merveenoyan: do we still support TF?

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+
+```py
+from transformers import AutoTokenizer
+
+tokenizer = AutoTokenizer.from_pretrained("gpt2")
+```
+
+Behind the scenes, `AutoTokenizer` performs these steps:
+
+1. **Download the tokenizer configuration.** The `from_pretrained` method fetches `tokenizer_config.json` from the Hub (or from a local directory).
```
merveenoyan, line 325: if we go very deep I'd also explain tokenizer_config.json and the params there. honestly I had to debug tokenizers many times and these files often have culprits 😄

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+| :---: | :---: | :---: |
+| Entry Point | `AutoTokenizer` | Automatically selects and instantiates the correct tokenizer class |
+| Model-Specific | `LlamaTokenizer`, `GPT2Tokenizer`, etc. | Configures the backend with model-specific architecture of normalizer, pre tokenizer, etc, special tokens, and settings |
+| Backend | `TokenizersBackend`, `PythonBackend`, `SentencePieceBackend` | Implements the actual tokenization using a specific engine |
+| Base | `PreTrainedTokenizerBase` | Defines the common interface and shared functionality |
+| Engine | `tokenizers` (Rust), SentencePiece, Pure Python | Performs raw tokenization |
+
+## v5 Separates Tokenizer Architecture from Trained Vocab
+
+The most significant change in Transformers v5 is a philosophical shift in how tokenizers are defined. **Tokenizers now work like PyTorch's `nn.Module`**: you define the architecture first, then fill it with learned parameters.
```
merveenoyan, line 356: here it sounds a bit like you have learnt params of a tokenizer from a different framework and you load it into HF tokenizers with an architecture. but when I read a bit down below I noticed it's opposite, you train your own tokenizer. if latter is not the case I'd clarify a bit

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+
+### The problem with v4: tokenizers were opaque and tightly coupled
+
+In v4, tokenizers were black boxes tied to pretrained checkpoint files. If you loaded `LlamaTokenizerFast`, you couldn't easily answer basic questions about it:
```
merveenoyan, line 360: very nice change!

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+from transformers import LlamaTokenizer
+
+# Instantiate the architecture
+tokenizer = LlamaTokenizer()
+
+# Train on your own data to fill in vocab and merges
+tokenizer.train(files=["my_corpus.txt"])
+```
+
+The tokenizer class now explicitly declares its structure. Looking at `LlamaTokenizer` in v5, you can immediately see:
```
merveenoyan, line 416: maybe show a snippet that shows below properties so people know how to check them!

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+
+| ![v5 llama](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/tokenizers/v5-llama.png) |
+| :--: |
+| `LlamaTokenizer` as seen in v5 `transformers` |
+
+This transparency was impossible in v4, where the same information was buried in serialized files.
+
+### One file, one backend, one recommended path
+
+v5 consolidates the two-file system *into a single file per model*. `LlamaTokenizer` now inherits from `TokenizersBackend`, which wraps the fast Rust tokenizer by default. The slow Python backend (`PythonBackend`) and SentencePiece backend (`SentencePieceBackend`) still exist for models that need them, but **Rust-backed tokenization is the preferred default**.
```
merveenoyan: do new models only have fast? I'd add that

tokenizers.md
ariG23498: Was this a bug? 🤔 

I always thought it was due to the mapping to bytes. I can hardcode "G" here, but that might be concerning for users reading the post and going through the code themselves.

tokenizers.md
ariG23498: I agree, but we wanted to keep this as is, [internal comment here](https://docs.google.com/document/d/1XSDW1gRFRw3owN8jRAmNGUk-edsriEhU4_EybJ07zH4/edit?disco=AAABxk9q92k).

_blog.yml
```diff
@@ -4992,4 +4992,12 @@
     - Claude
     - Codex
     - Gemini
-    - agents
\ No newline at end of file
+    - agents
+
+- local: tokenizers
+  date: Dec 17, 2025
```
pcuenca:
```suggestion
  date: Dec 18, 2025
```

tokenizers.md
ariG23498: Does `<|im_start|>` mean image start? I always thought this was the imaginary monologue start. [From a FastAI post](https://www.fast.ai/posts/2025-10-16-karpathy-tokenizers).
image

tokenizers.md
ariG23498: Yes! This caught me offguard too.

tokenizers.md
ariG23498: We do showcase the file as a screengrab that does exactly this.

tokenizers.md
ariG23498: The idea with v5 is, we will never have fast from now on. The `TokenizersBackend` is fast and we want to keep it that way.

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+- [Summary](#summary)
+
+> [!TIP]
+> For experts: If you're already familiar with the concepts and want to understand the changes in v5, go to [v5 Separates Tokenizer Architecture from Trained Vocab](#v5-separates-tokenizer-architecture-from-trained-vocab)
+
+Before diving into the changes, let's quickly cover what tokenization does and how the pieces fit together.
+
+> [!NOTE]
+> Tokenization is a broad concept used across natural language processing and text processing generally. This post focuses specifically on tokenization for Large Language Models (LLMs) using the [`transformers`](https://github.com/huggingface/transformers) and [`tokenizers`](https://github.com/huggingface/tokenizers) libraries.
+
```
pcuenca: Perhaps too many boxes in the intro, we could potentially move this to a regular paragraph in the next section.

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+	frameborder="0"
+	width="850"
+	height="450"
+></iframe>
+
+Language models don't read raw text. They consume sequences of integers usually called **token IDs or input IDs**. Tokenization is the process of converting raw text into these token IDs.
+
+```py
+from transformers import AutoTokenizer
+# Let's use a tokenizer that was already trained:
```
pcuenca:
```suggestion
```

"training" a tokenizer may be a confusing concept for people that are reading this section to understand what tokenizers are. I think it's enough to focus on the text -> IDs conversion that we just introduced.

tokenizers.md
pcuenca: I don't think it's a bug, it's a representation of the `_` prefix space, no? It may look puzzling indeed, not sure if we discuss more details later or if we should use `decode` instead to retrieve text intended for human consumption.

pcuenca: Perhaps we could say (now or later) that `'Ġworld'` is a single token that represents the character sequence " world" (with the space).

tokenizers.md
pcuenca: Why does context length matter to understand tokenizers? Perhaps the discussion should be about trade-offs between more compression vs better representation. For example, in Chinese a token per character might be useful because characters carry a lot of information. However, in English this would result in very few tokens that carry very little semantic information, so it'd be hard for the model to learn good representations. If we consider a token per word, then we'd have lots of unknown tokens (typos, slang), and difficulty to represent derivatives that share the same stem etc.

tl;dr: it's complicated and that's why modern tokenizers try to strike a balance; language matters; training is a thing.

tokenizers.md
pcuenca:
```suggestion
A good tokenizer *compresses* text into the smallest amount of tokens. Fewer tokens means more usable context without increasing model size. Training a tokenizer boils down to finding the best compression rules for your datasets. For example, if you work on Chinese corpus you can sometimes find [very nice surprises 😉](https://x.com/suchenzang/status/1697862650053660721).
```

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+A good tokenizer *compresses* text into the smallest amount of tokens. Fewer tokens means more usable context without increasing model size.Training a tokenizer boils down to finding the best compression rules for your datasets. For example if you work on Chinese corpus you can sometimes find [very nice surprises 😉](https://x.com/suchenzang/status/1697862650053660721).
+
+## The tokenization pipeline
+
+Tokenization happens in stages. Each stage transforms text before passing it to the next:
+
+| Stage | Purpose | Example |
+| :---: | :---: | :---: |
+| **Normalizer** | Standardizes text (lowercasing, unicode normalization, whitespace cleanup) | `"HELLO World"` → `"hello world"` |
+| **Pre-tokenizer** | Splits text into preliminary chunks | `"hello world"` → `["hello", "world"]` |
```
pcuenca:
```suggestion
| **Pre-tokenizer** | Splits text into preliminary chunks | `"hello world"` → `["hello", " world"]` |
```

Careful, we lost a char here

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+| **Model** | Applies the tokenization algorithm (BPE, Unigram, etc.) | `["hello", "world"]` → `[9906, 1917]` |
```
pcuenca:
```suggestion
| **Model** | Applies the tokenization algorithm (BPE, Unigram, etc.) | `["hello", " world"]` → `[9906, 1917]` |
```

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+| **Post-processor** | Adds special tokens (BOS, EOS, padding) | `[9906, 1917]` → `[1, 9906, 1917, 2]` |
+| **Decoder** | Converts token IDs back to text | `[9906, 1917]` → `"hello world"` |
+
+Each component is *independent*. You can swap normalizers or change the algorithm without rewriting everything else.
```
pcuenca: I'd stress that there are multiple variants of all these things, lest the reader think all tokenizers/models always work in lowercase. A link to something like the [normalizers docs](https://huggingface.co/docs/tokenizers/en/api/normalizers) might be eye opening. They are just the method signatures, but still.

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+
+```py
+from transformers import AutoTokenizer
+
+tokenizer = AutoTokenizer.from_pretrained("google/gemma-3-270m-it")
+
+print(f"{tokenizer._tokenizer.normalizer=}")
+# Replace(...)
```
pcuenca, line 104: Do you think it is worth anticipating that `tokenizer` is the Python wrapper for the `_tokenizer` Rust implementation?

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+print(f"{tokenizer._tokenizer.post_processor=}")
+# TemplateProcessing(...)
+
+print(f"{tokenizer._tokenizer.decoder=}")
+# Sequence(decoders=[Replace(...), ByteFallback(), Fuse()])
+```
+
+## Tokenization algorithms
+
+The following algorithms dominate modern language model tokenizers:
```
pcuenca, line 121: I think this would follow very naturally after we have intuitively introduced the complexity of the task, as suggested above.

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+
+tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")
+print(tokenizer._tokenizer.model)
+
+# WordPiece(...)
+```
+
+## Accessing tokenizers through transformers
+
+The [`tokenizers`](https://github.com/huggingface/tokenizers) library is a Rust-based tokenization engine. It is fast, efficient, and completely model-agnostic. The library handles the mechanics of converting text into token IDs. The `tokenizers` library is a general-purpose tool. It implements the tokenization algorithms, but does not implement the conventions that connect those algorithms to specific models.
```
pcuenca:
```suggestion
The [`tokenizers`](https://github.com/huggingface/tokenizers) library is a Rust-based tokenization engine. It is fast, efficient, and completely model-agnostic. The library handles the mechanics of converting text into token IDs and back. The `tokenizers` library is a general-purpose tool that implements the tokenization algorithms, but does not implement the conventions that connect those algorithms to specific models.
```

pcuenca: Slightly confused, are we talking about tokenizer models or language models?

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+
+>[!TIP]
+> `tokenizers` is what PyTorch is to `transformers`
```
pcuenca: aren't we missing something here? `tokenizers is to xxx what PyTorch is to transformers`? We need a concept that refers to "the tokenization and chat templating engine in the transformers library" 🤔

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+
+print(encodings.ids)
+# [9906, 1917]
+print(encodings.tokens)
+# ['Hello', 'Ġworld']
+```
+
+The output is raw tokenization. You get token IDs and the string pieces they correspond to. Nothing more.
+
+Now consider what's missing. The `SmolLM3-3B` is a *conversational model*. When you interact with it, you typically structure your input as a conversation with roles like "user" and "assistant." The model expects special formatting tokens to indicate these roles. The raw `tokenizers` library has no concept of any of this.
```
pcuenca:
```suggestion
Now consider what's missing. The `SmolLM3-3B` is a *conversational model*. When you interact with it, you typically structure your input as a conversation with roles like "user" and "assistant". The language model expects special formatting tokens to indicate these roles. The raw `tokenizers` library has no concept of any of this.
```

tokenizers.md
pcuenca: yeah, `<|im_start|>` is very frequent across models, no idea it referred to "imaginary monologue" lol

tokenizers.md
pcuenca: Let's skip in my opinion: "PyTorch tensors, numpy arrays, and others"

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+The `transformers` tokenizer adds everything the raw library lacks:
+
+* **Chat template application.** The `apply_chat_template` method formats conversations according to the model's expected format, inserting the correct special tokens and delimiters.  
+* **Automatic special token insertion.** Beginning-of-sequence and end-of-sequence tokens are added where the model expects them.  
+* **Truncation to context length.** You can specify `truncation=True` and the tokenizer will respect the model's maximum sequence length.  
+* **Batch encoding with padding.** Multiple inputs can be padded to the same length with the correct padding token and direction.  
+* **Return format options.** You can request PyTorch tensors (`return_tensors="pt"`), TensorFlow tensors, or NumPy arrays.
+
+> [!NOTE]
+> More than adding, `transformers` defines the API for tokenization that is now most commonly used is the entire ML! (`encode`, `decode`, `convert_tokens_to_ids`, etc.)
```
pcuenca:
```suggestion
> `transformers` implements the tokenization API that is most commonly used in the entire ML community (`encode`, `decode`, `convert_tokens_to_ids`, etc.)
```

(trying to make it sound less like we invented it, feel free to revert)

tokenizers.md
pcuenca: Maybe `tokenization api`? `inputs preparation`?

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+| The class hierarchy for tokenizers inside transformers |
+
+### `PreTrainedTokenizerBase` defines the common interface for all tokenizers
+
+[`PreTrainedTokenizerBase`](https://github.com/huggingface/transformers/blob/7f52a2a4ea8ab49b7f069df7fac58a5b280d4919/src/transformers/tokenization_utils_base.py#L964C7) is the abstract base class for all tokenizers in `transformers`. It defines the interface that every tokenizer must implement.
+
+The base class handles functionality that doesn't depend on the tokenization backend:
+
+* **Special token properties.** Properties like `bos_token`, `eos_token`, `pad_token`, and `unk_token` are defined here. These properties provide access to the special tokens that models use to mark sequence boundaries and handle unknown inputs.  
+* **Encoding interface.** The `__call__` method, `encode`, `encode_plus`, and `batch_encode_plus` methods are defined here. These methods accept text input and return token IDs along with attention masks and other metadata.
```
pcuenca: I thought `encode_` and `batch_encode_` were going to be simplified into one?

tokenizers.md
pcuenca: And to Merve's comment, yes, perhaps we can introduce special tokens giving the example of the "bos" and "eos" tokens, and how they are useful for the model to learn where a new sequence starts and ends.

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+        ...
+```
+
+When you call encoding methods on a `TokenizersBackend` tokenizer, the class delegates the actual tokenization to the Rust backend:
+
+```py
+def _batch_encode_plus(self, batch_text_or_text_pairs, ...):
+    encodings = self._tokenizer.encode_batch(batch_text_or_text_pairs, ...)
+    # Convert Rust Encoding objects to transformers BatchEncoding
+    return BatchEncoding(...)
```
pcuenca: I don't understand this comment. Maybe skip and just show the call to `_tokenizer` and then `...` to show that more stuff comes later? Your call!

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+* `LlamaTokenizer`  
+* `GemmaTokenizer`
+
+These model-specific classes configure the backend with the correct vocabulary, merge rules, special tokens, and normalization settings for their respective models.
+
+### `PythonBackend` provides a pure-Python mixin
+
+[`PythonBackend`](https://github.com/huggingface/transformers/blob/7f52a2a4ea8ab49b7f069df7fac58a5b280d4919/src/transformers/tokenization_python.py#L400) inherits from `PreTrainedTokenizerBase` and implements tokenization in pure Python. The class is aliased as [`PreTrainedTokenizer`](https://github.com/huggingface/transformers/blob/7f52a2a4ea8ab49b7f069df7fac58a5b280d4919/src/transformers/tokenization_python.py#L1400C1).
+
+The pure-Python backend exists for several reasons:
```
pcuenca, line 279: Didn't we say in the very beginning that there was a single "fast" backend? Does it refer to a single-vs-multiple fast implementation, whereas we used to have multiple fast tokenizers? If so, it can be confused with meaning that "there are no longer slow implementations".

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+
+```py
+class SentencePieceBackend(PythonBackend):
+    def __init__(self, vocab_file, ...):
+        self.sp_model = spm.SentencePieceProcessor()
+        self.sp_model.Load(vocab_file)
+        ...
+```
+
+Models that use SentencePiece tokenization inherit from this backend. Examples include:
```
pcuenca, line 306: Don't we have sentencepiece models that are implemented with fast tokenizers too?

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+```py
+from transformers import AutoTokenizer
+
+tokenizer = AutoTokenizer.from_pretrained("gpt2")
+```
+
+Behind the scenes, `AutoTokenizer` performs these steps:
+
+1. **Download the tokenizer configuration.** The `from_pretrained` method fetches `tokenizer_config.json` from the Hub (or from a local directory).  
+2. **Identify the model type.** The configuration contains metadata that identifies the model type (e.g., "gpt2", "llama", "bert").
```
pcuenca: Maybe link to the corresponding line in a config file here.

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+# Instantiate the architecture
+tokenizer = LlamaTokenizer()
+
+# Train on your own data to fill in vocab and merges
+tokenizer.train(files=["my_corpus.txt"])
+```
+
+The tokenizer class now explicitly declares its structure. Looking at `LlamaTokenizer` in v5, you can immediately see:
+
+* It uses **BPE** as its tokenization model
```
pcuenca: Maybe link to the line, in addition to the screenshot.

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+# Train on your own data to fill in vocab and merges
+tokenizer.train(files=["my_corpus.txt"])
+```
+
+The tokenizer class now explicitly declares its structure. Looking at `LlamaTokenizer` in v5, you can immediately see:
+
+* It uses **BPE** as its tokenization model  
+* It may add a **prefix space** before text  
+* Its special tokens (`unk`, `bos`, `eos`) sit at specific vocabulary positions  
+* It does **not normalize** input text
```
pcuenca: Same, the link to `normalizer = None` would be cool.

tokenizers.md
pcuenca: I'm still confused about SentencePiece going through the legacy path. I would clarify it's not necessary.

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+        yield dataset[i : i + batch]["text"]
+
+trained_tokenizer = tokenizer.train_new_from_iterator(
+    text_iterator=get_training_corpus(),
+    vocab_size=32000,
+    length=len(dataset),
+    show_progress=True,
+)
+
+trained_tokenizer.save_pretrained("./my_custom_tokenizer")
```
pcuenca: consider push_to_hub :)

tokenizers.md
```diff
@@ -0,0 +1,494 @@
...
+
+## Summary
+
+Transformers v5 brings three improvements to tokenization:
+
+1. **One file per model** instead of separate slow/fast implementations  
+2. **Visible architecture** so you can inspect normalizers, pre-tokenizers, and decoders  
+3. **Trainable templates** that let you create custom tokenizers matching any model's design
+
+The wrapper layer between `tokenizers` and Transformers remains essential. It adds model awareness, context lengths, chat templates, special tokens, that raw tokenization doesn't provide. V5 just makes that layer clearer and more customizable.
```
pcuenca: I think we could also have a Resources section with links.

Which reminds me, we have some excellent posts that we could link to from various places throughout the text:
- https://huggingface.co/blog/qgallouedec/gotchas-in-tokenizer-behavior
- https://huggingface.co/blog/chat-templates (a bit old, but good as an intro to chat templates)

pcuenca (approved):
Great work, covering a lot of ground! 🔥

tokenizers.md
ariG23498: I would like to nominate this as one of the greatest catches in the history of GH reviews. 🙏

tokenizers.md
ariG23498: We do mention that `self._tokenizer` is the rust tokenizer, but I think mentioning it again would not hurt.

tokenizers.md
itazap: we plan to deprecate `tokenizer_config.json`, in v5 we don't really rely on the params here as much apart from reading the tokenizer class to load. So we can probably skip over this!

tokenizers.md
itazap: yes its as it sounds! we read the `vocab` and `merges` as parameters  from `tokenizers` or. `sentencepiece` files (`tokenizer.jso`n vs `sentencepiece.model` respectively)  and  pass the params to create a tokenizer.

We also support training your own tokenizer but that is unrelated to loading one with `.from_pretrained`! Similar how in PyTorch you can define a model with nn.Embedding, and fill it with weights or you can initialize it and train those weights

tokenizers.md
itazap: Okay we shouldn't refer to python backend as `slow` in v5 but we should explicitly state that TokenizersBackend classes were the `fast` classes from v4, and PythonBackend classes were the `slow` ones

If it's not too wordy:

v5 consolidates the two-file system into a single file per model. LlamaTokenizer now inherits from TokenizersBackend, which wraps the Rust-based tokenizer that was previously exposed as the “fast” implementation and is now the default. The former “slow” Python implementation lives explicitly behind PythonBackend, and SentencePieceBackend remains for models that require it, but Rust-backed tokenization is the preferred default.

tokenizers.md
itazap: we don't call backends 'slow' and 'fast' anymore we separate them by the library that its using. There was only ever a single "fast" backend - and that was `tokenizers`. In v5 we made it the default and models only have 1 tokenization file (in v4 we had `tokenization_llama.py` and `tokeniation_llama_fast.py` which was the `tokenizers` backend. we only kept the latter!

tokenizers.md
itazap: (we try to  get the params from the tokenizer file itself)

tokenizers.md
itazap: yes we deprecated `batch_encode` (now just supported by `encode` and `encode_plus` (its internal now, `__call__` calls `_encode_plus` (private function)

tokenizers.md
pcuenca: Yes, sorry, I was using `fast` and `slow` as shortcuts. My concern comes from this sentence at the top of the post, inside the blue box:

Screenshot 2025-12-18 at 17 11 00

I think it's a bit confusing because then we show that there's the legacy Python backend path, which is furthermore used by the SentencePieceBackend. I think we should clarify:
- The "fast" (or Rust-based, or TokenizersBackend) is the default, and it's great.
- There's a pure Python implementation that's only meant for legacy or special purposes.
- Despite the existence of the SentencePiece subclass, sentencepiece tokenizers can also be implemented with the TokenizersBackend (unless I'm wrong).

Otherwise it looks a bit confusing that we refer to a single backend but then talk about others.

PR #3234, opened 2025-12-18
Attempt to fix wrong author

pcuenca:
Screenshot 2025-12-18 at 13 04 44


pcuenca:
Cursor helped me find a few suspects #3235. There are also some true 404s. I can review them all unless we plan to go to case insensitive.

Wauplin:
Note that we can't avoid 404 in theory since users can rename themselves (but still better to have most of them fixed)

PR #3238, opened 2025-12-20
Tokenizer v5: Fix apply chat template + typo 

qgallouedec:
Avoid a common mistake, see 

https://huggingface.co/blog/qgallouedec/gotchas-in-tokenizer-behavior#7-chat-template-and-tokenization-dont-compose-due-to-special-tokens

### DIFF
--- a/tokenizers.md
+++ b/tokenizers.md
@@ -200,7 +200,7 @@ print(text)
 model_inputs = tokenizer([text], add_special_tokens=False, return_tensors="pt")
 ```
 
-Notice how the special tokens like `<|im_start>` and `<|im_end>` are applied to the prompt before tokenizing. This is useful for the model to learn where a new sequence starts and ends.
+Notice how the special tokens like `<|im_start|>` and `<|im_end|>` are applied to the prompt before tokenizing. This is useful for the model to learn where a new sequence starts and ends.
 
 The `transformers` tokenizer adds everything the raw library lacks:
 


tokenizers.md
```diff
@@ -197,10 +197,10 @@ print(text)
 # Give me a brief explanation of gravity in simple terms.<|im_end|>
 # <|im_start|>assistant
 
-model_inputs = tokenizer([text], return_tensors="pt")
+model_inputs = tokenizer([text], add_special_tokens=False, return_tensors="pt")
```
pcuenca, line 200: Yes, I agree this is dangerous in general (although it does not affect this particular case).

In fact, it'd be useful if we could recommend `apply_chat_template` with `tokenizer=True` as the preferred way to prepare conversational inputs cc @ariG23498. (No need to change this example which is illustrative, but we could add a comment in an appropriate point).

tokenizers.md
```diff
@@ -197,10 +197,10 @@ print(text)
...
 ```
 
-Notice how the special tokens like `<|im_start>` and `<|im_end>` are applied to the prompt before tokenizing. This is useful for the model to learn where a new sequence starts and ends.
+Notice how the special tokens like `<|im_start|>` and `<|im_end|>` are applied to the prompt before tokenizing. This is useful for the model to learn where a new sequence starts and ends.
```
pcuenca, line 203: oh yes

PR #3350, opened 2026-04-14
feat: sync blog to HF bucket on push to main

rtrompier:
Add a GitHub Action that syncs the blog repo to the HF bucket `hf-doc-build/blog` on every push to main.

After the sync, the workflow waits for the hub FUSE mount to poll (~30s) and then notifies the hub via `/api/webhooks/blog` to refresh the blog metadata in MongoDB (`updateCanonicalBlogData`). This avoids the race where the hub would read stale data if notified immediately on push.

Used in combination with the upcoming moon-landing change that splits blog-and-doc into two separate bucket-backed deployments.

### DIFF
--- a/.github/workflows/sync-to-bucket.yml
+++ b/.github/workflows/sync-to-bucket.yml
@@ -24,8 +24,11 @@ jobs:
         run: sleep 35
 
       - name: Notify hub to refresh blog metadata
+        env:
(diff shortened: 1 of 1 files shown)


jagwar (approved):
LGTM

paulinebm (approved):

PR #3354, opened 2026-04-16
ST Multimodal: Qwen3 no more revision needed, add new models

tomaarsen:
Hello!

## Pull Request overview
* Remove `revision` for Qwen3 models
* Add new models that are integrated (or not yet)

## Details
I'm not planning on updating this forever, but it seems good to make the list a bit more extensive for the time being.

- Tom Aarsen


ariG23498 (approved):
Thanks for the updates.

PR #3372, opened 2026-05-04
Unlocking asynchronicity in continuous batching

remi-or:
This PR adds the `Unlocking asynchronicity in continuous batching` blog post.


_blog.yml
```diff
@@ -5315,3 +5315,12 @@
     - enterprise
     - partnerships
     - hub
+
+- local: continuous_async
+  date: April 28, 2026
```
pcuenca: reminder to update before release

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+thumbnail: /blog/assets/continuous_async/thumbnail.png
+authors:
+- user: ror
+---
+
+# Unlocking asynchronicity in continuous batching
+
+![Title card](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/banner.png)
+
+*TL;DR: in this blog post, we are going to explain how to separate CPU and GPU workloads to get a massive performance boost for inference.*
```
pcuenca:
```suggestion
*TL;DR: we explain how to separate CPU and GPU workloads to get a massive performance boost for inference.*
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+*This is the second post in a series on efficient LLM inference. The [first post](https://huggingface.co/blog/continuous_batching) covered continuous batching from first principles. In the rest of this post, we will assume the reader has a working knowledge of LLM inference (KV cache, FlashAttention, attention masks, etc.).*
```
pcuenca:
```suggestion
*This is the second post in a series on efficient LLM inference. The [first post](https://huggingface.co/blog/continuous_batching) covered continuous batching from first principles. It introduces some concepts we build upon: KV cache, FlashAttention, attention masks, etc.*
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+Everyone likes owning their own tools. And with AI becoming the largest productivity multiplier around, everyone wants to own their model. But owning a model is not enough: you also need to run it.
+Most likely, on a modern GPU, which does not run cheap. So when you are renting a GPU to run your own models, you want to use it at the fullest of its capacities.
```
pcuenca: Not fully sold on this, perhaps a more direct way to refer to maximizing capacity would work. Will come back to it later.

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+Everyone likes owning their own tools. And with AI becoming the largest productivity multiplier around, everyone wants to own their model. But owning a model is not enough: you also need to run it.
+Most likely, on a modern GPU, which does not run cheap. So when you are renting a GPU to run your own models, you want to use it at the fullest of its capacities.  
+
+We already covered how **continuous batching** helps you schedule batches of requests efficiently, so you waste no compute on padding.  
+But what about the gaps in between GPU batches, when the CPU takes over? During that time, the CPU is doing some work, while your GPUs sit idle. This is not acceptable: we want our GPUs to be working 100% of the time, always doing actual compute.  
+
+To achieve this, we can use **asynchronous batching**. We are going to disentangle CPU batch preparation from GPU batch compute, so we can hide CPU latency and always have a productive GPU.
```
pcuenca:
```suggestion
But what happens in between GPU batches? If we work in the default synchronous mode, we'll see the CPU taking over (processing the last batch from the GPU, preparing the next one) while the GPUs sit idle. We can do better: we want our GPUs to be working 100% of the time, always doing actual compute.  

To achieve this, we can use **asynchronous batching**. We are going to disentangle CPU batch preparation from GPU batch compute, so both can run in parallel and we always have a productive GPU 🔥
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+Most likely, on a modern GPU, which does not run cheap. So when you are renting a GPU to run your own models, you want to use it at the fullest of its capacities.  
+
+We already covered how **continuous batching** helps you schedule batches of requests efficiently, so you waste no compute on padding.  
+But what about the gaps in between GPU batches, when the CPU takes over? During that time, the CPU is doing some work, while your GPUs sit idle. This is not acceptable: we want our GPUs to be working 100% of the time, always doing actual compute.  
+
+To achieve this, we can use **asynchronous batching**. We are going to disentangle CPU batch preparation from GPU batch compute, so we can hide CPU latency and always have a productive GPU.  
+
+## Synchronous batching
+
+Let us remind ourselves of how synchronous batching works.
```
pcuenca:
```suggestion
This is how naive synchronous batching works:
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+![Synchronous batching](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/synchronous_batching.png)
+
+The figure above shows how a synchronous batch step runs. The CPU starts by preparing the batch: it selects which requests to include, updates the KV cache table, evicts requests that finished in the previous step, and admits new ones to fill the freed space. Once that is done, it transfers the inputs to the GPU. The GPU runs its forward pass and samples (i.e. chooses) a new token for each request. The results come back to the CPU, so it knows what token each request just produced. Then the whole cycle starts again.
```
pcuenca:
```suggestion
When the CPU prepares a new batch, it selects which requests to include, updates the KV cache table, evicts requests that finished in the previous runs, and admits new ones to fill the freed space. Once that is done, it transfers the prepared inputs to the GPU. The GPU runs its forward pass and samples (i.e. chooses) a new token for each request. The results come back to the CPU, so it knows what token each request just produced, then the whole cycle repeats again.
```

Illustration nits: perhaps we can use "batch N" instead of batch 0, and potentially show that this is a block repeated all the time. We can, for example, replicate half of it left and right, and enclose the area under interest in a rectangle. But it might become too busy perhaps. Also, the red "GPU idle" message is applicable to the left area, if the illustration shows it as part of a continuous cycle.

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+Notice the red annotation on the right: after the GPU finishes computing, it goes idle. The next batch cannot start until the CPU has gone through its update step: sampling the output tokens, updating request states, re-scheduling the batch. The CPU sends the next batch to the GPU only once all of that is done. The GPU, meanwhile, waits.
```
pcuenca:
```suggestion
Notice the red annotation on the right: after the GPU finishes computing, it goes idle. The next batch cannot start until the CPU has gone through its update step: sampling the output tokens, updating request states, re-scheduling the batch.
```

I think it's enough to make the point, but feel free to stress if necessary.

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+This is the core inefficiency of synchronous batching: the CPU and GPU take turns. While the GPU is computing, the CPU is idle. While the CPU is updating, the GPU is idle. Neither is ever doing useful work at the same time. For a single forward pass this might seem like a small price to pay, but in a continuous batching loop running hundreds of steps per second, these idle gaps accumulate into real throughput loss.
```
pcuenca:
```suggestion
This is the core inefficiency of synchronous batching: the CPU and GPU take turns. While the GPU is computing, the CPU is idle. While the CPU is updating, the GPU is idle. In no circumstances they are both doing useful work at the same time. For a single forward pass this might seem like a small price to pay, but in a continuous batching loop running hundreds of steps per second, these idle gaps accumulate into real throughput loss.
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+To showcase this, we profile the time spent on CPU and GPU when generating 8K tokens with a batch size of 32 using an 8B model:
+
+![CPU and GPU activity timeline](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/cpu_gpu_phases_sync.png)
+
+The timeline alternates between green (GPU active, CPU idle) and red (CPU active, GPU idle): the two never overlap. Total generation time is 300.6 seconds, with 24.0% of that spent with an idle GPU waiting for the CPU to finish. Nearly a quarter of all generation time is wasted. This is the pessimistic way of viewing things.
```
pcuenca:
```suggestion
The timeline alternates between green (GPU active, CPU idle) and red (CPU active, GPU idle): the two never overlap. Total generation time is 300.6 seconds, with 24.0% of that spent with an idle GPU waiting for the CPU to finish. Nearly a quarter of all generation time is wasted, from the point of view of the GPU. This is the pessimistic way of viewing things.
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+The optimistic way of viewing things is that if we eliminate CPU overhead entirely, generation time would drop from 300 to 228 seconds. That is a free 24% speedup, which does not come often. Especially since this requires zero new kernel or model change, just careful coordination of hardwares.
```
pcuenca:
```suggestion
The optimistic way is that generation time would drop from 300 to 228 seconds (a free 24% speedup!), if we could eliminate CPU overhead entirely. This requires zero new kernel or model changes, just careful coordination of hardware.
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+Fundamentally, the idea is simple: we need to figure out how to run batch preparation for batch N+1 while batch N is computing. But this simple idea hides a few technical difficulties that need to be addressed first:
```
pcuenca:
```suggestion
Fundamentally, the idea is simple: we need to figure out how to run batch preparation for batch N+1 while batch N is computing. But this simple idea hides a few technical difficulties:
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+- How can we launch something on the GPU and get back CPU control?
+- How can we make sure data is ready, for either CPU or GPU tasks, before they are launched?
+- How can we prepare batch N+1 if it is based on the predictions of batch N?
```
pcuenca:
```suggestion
- How can we launch something on the GPU and get back control to the CPU?
- How can we make sure data is ready, for either CPU or GPU tasks, by the time each task is launched?
- How can we prepare batch N+1 if it is based on the predictions of batch N?
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+By answering those questions, we are going to build the asynchronous batching feature that powers the [transformers](https://github.com/huggingface/transformers) library's continuous batching today.
```
pcuenca:
```suggestion
By answering those questions, we are going to build asynchronous batching from scratch. We followed the same steps to implement it as part of continuous batching in the [transformers](https://github.com/huggingface/transformers) library. Feel free to check the code and compare!
```

maybe?

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Creating concurrency
+
+To recap, our end goal is to have **concurrent** execution of CPU and GPU operations, i.e. both kinds of operations running at the same time. To do this, we need a way to categorize our operations, so we can let the machine know which operations can run concurrently. We can achieve this with CUDA streams.
```
pcuenca:
```suggestion
Our end goal is to have **concurrent** execution of CPU and GPU operations. We need a way to categorize our operations, so we can let the machine know which operations can run concurrently. We can achieve this using CUDA streams.
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Default and non-default streams
+
+If you have never explicitly used CUDA streams in PyTorch, you might be surprised they exist at all. A typical PyTorch script never mentions them, and it does not *feel* like GPU operations are asynchronous: the CPU seems to wait for the GPU to finish before moving on. That feeling is accurate, and it comes from the **default stream**.
+
+When you call a PyTorch operation without specifying a stream, it lands on the default stream. The default stream has one special property: it is **synchronizing**. If an operation is scheduled on the default stream, it waits for **all other streams to be flushed**. Ie. all work on the GPu has to be over before a single operation on the default stream can start. The reverse is also true: any operation, regardless of its stream, waits for the default stream to be flushed before it launches.
```
pcuenca:
```suggestion
When you call a PyTorch operation without specifying a stream, it lands on the default stream. The default stream has one special property: it is **synchronizing**. If an operation is scheduled on the default stream, it waits for **all other streams to be flushed**, i.e. all work on the GPU has to be over before a single operation on the default stream can start. The reverse is also true: any operation, regardless of its stream, waits for the default stream to be flushed before it launches.
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+We established that no GPU operation should land on the default stream. But the question remains: if we are not using the default stream, what streams should we use? Let us go back to the synchronous batching figure:
+
+![Synchronous batching](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/synchronous_batching.png)
+
+We can identify three distinct GPU operations:
+1. Transfer of inputs from CPU to GPU
+2. Compute on the GPU
+3. Transfer of outputs from the GPU to the CPU
+
+This means we need three streams: one for compute, one for CPU-to-GPU transfers, and one for GPU-to-CPU transfers. The transfers are independent, so there is no reason to serialize them, and each gets its own stream.
```
pcuenca:
```suggestion
This means we need three streams: one for compute, one for CPU-to-GPU transfers, and one for GPU-to-CPU transfers. The transfers are independent, so there is no reason to serialize them, and each one gets its own stream.
```

(But there is dependency on data, isn't it? I guess we'll see it later)

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+A note on nomenclature: when talking about CPUs and GPUs, the convention (used throughout the CUDA documentation) is to call the CPU the **host** and the GPU the **device**. We will use that convention from now on. CPU-to-GPU transfers are called **host-to-device** (H2D) transfers, and GPU-to-CPU transfers are called **device-to-host** (D2H) transfers. Hence, the three streams are the H2D stream, the compute stream, and the D2H stream.
```
pcuenca:
```suggestion
A note on nomenclature: when talking about CPUs and GPUs, the convention used throughout the CUDA documentation is to call the CPU the **host** and the GPU the **device**. We will use that convention from now on. CPU-to-GPU transfers are called **host-to-device** (H2D) transfers, and GPU-to-CPU transfers are called **device-to-host** (D2H) transfers. Hence, the three streams are the H2D stream, the compute stream, and the D2H stream.
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+Let us now try to use streams to asynchronously launch a batch on the GPU and get back CPU control. From the CPU, we do the following:
+
+1. Prepare the batch input data on the CPU (no stream, CPU-only operations)
+2. Transfer it to the GPU (using the H2D stream)
+3. Run compute on the GPU (using the compute stream)
+4. Retrieve the batch outputs (using the D2H stream)
+5. Take a look at the results (no stream)
+
+If we do this using only CUDA streams, the results are available almost instantly and they are incorrect. To understand why, let us look at what happened:
```
pcuenca, line 106: aha, nice

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+The operations are all running correctly in isolation. The problem is that we never told the streams to wait for each other. We know that compute must start after H2D completes, and that D2H must start after compute completes, but we did not enforce that ordering. We need a mechanism to say "do not start this operation until that one is done" across stream boundaries.
+
+## Enforcing synchronization
+
+To enforce synchronization between the streams, we are going to use **CUDA events**.
+
+### What is a CUDA event?
+
+A CUDA event is a marker that can be recorded into a stream. When the GPU reaches that marker during execution, it sets the event as completed. Any other stream can then be told to wait for that event before starting its next operation. Concretely, there are two operations: `stream.record(event)`, which inserts the marker into a stream at the current position, and `stream.wait(event)`, which blocks a stream from proceeding until the event is marked complete. Importantly, `wait` blocks the *stream*, not the CPU: the CPU call returns immediately, and only the stream is held back.
```
pcuenca:
```suggestion
A CUDA event is a marker that can be recorded into a stream. When the GPU reaches that marker during execution, it sets the event as completed. Any other stream can then be told to wait for that event before starting its next operation. Concretely, there are two operations: `stream.record(event)`, which inserts the marker into a stream at the current position, and `stream.wait(event)`, which blocks a stream from proceeding until the event is marked complete. Importantly, `wait` blocks the *stream*, not the CPU or other streams running in parallel: the CPU call returns immediately, and only the waiting stream is held back.
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+![CUDA events](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/events.png)
```
pcuenca, line 122: Conceptually, perhaps the "event" could appear at the end of stream 1 and the "wait" could appear at the beginning of stream 2. Actually, stream 2 could perhaps start with a waiting period until the event happens.

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+The figure above shows a single event synchronizing two streams. The CPU issues three operations in rapid succession (the three small blocks): launch input preparation on stream 1, record the event on stream 1, then tell stream 2 to wait for it. Then the CPU continues immediately. Stream 1 runs its operation, and when it completes, the event is set. Stream 2 is held at the wait marker the whole time, and only starts compute once the event is marked complete. The CPU was not involved in any of this: the ordering was enforced entirely on the GPU side.
+
+### Using events in CB
```
pcuenca: Not sure if `CB` has been defined before; let's do it earlier otherwise.

pcuenca: Not sure if `CB` has been defined before; let's do it earlier otherwise.

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+![Successful asynchronous batching](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/success_async.png)
+
+The figure above shows how this unfolds. The CPU prepares the batch, then quickly enqueues all the GPU work: the H2D transfer, the forward pass, the D2H transfer, with `record` and `wait` calls inserted between each stage. After that, the CPU is free. The GPU takes over, executing each stream in order as its dependency event is set. Notice the green annotation on the right: once the D2H transfer completes, the CPU comes back and reads the results. This final synchronization is the only point where the CPU blocks in the whole step. To implement it, we record a third event on the D2H stream after the output transfer, then call `d2h_done_event.synchronize()` on the CPU side. `synchronize` blocks the CPU until the D2H stream reaches that marker.
+
+This is the key difference from synchronous batching. Before, the CPU blocked after every operation. Now it blocks once, at the very end, only to read results it genuinely needs at that moment. Everything else runs in the background.
```
pcuenca: Yes, but so far we are effectively using parallel computation to serialize tasks lol. Perhaps we could call this out a bit more explicitly so the reader is prepared to see the magic in a moment.

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+We can create any number of CUDA graphs in the same pool and the total memory usage is still capped at the maximum across graphs. This is showcased below.
+
+![CUDA graph memory pool](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/memory_pool.png)
+
+Now that we know how to prevent data corruption, we can address the second issue: getting the output tokens of batch N into the inputs of batch N+1.  
+
+### Carry-over
+
+Consider a request that appears in both batch N and batch N+1. In batch N, it produces a new token. That token is its input for batch N+1. The problem is that when we are preparing batch N+1's input buffer, we do not have that token yet: batch N is still running.
+To address this, we use a placeholder token when building batch N+1. We will use 0 as a placeholder, for reason that will become apparent later. We replace that placeholder after batch N is done computing and before batch N+1 starts the forward pass. We call that step the **carry-over**, because we are carrying over the new tokens from batch N to batch N+1. The idea behind carry-over is illustrated below:
```
pcuenca:
```suggestion
To address this, we use a placeholder token when building batch N+1. We will use 0 as a placeholder, for reasons that will become apparent later. We replace that placeholder after batch N is done computing and before batch N+1 starts the forward pass. We call that step the **carry-over**, because we are carrying over the new tokens from batch N to batch N+1. The idea behind carry-over is illustrated below:
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+
+## Does it actually work?
+
+To find out, we run the same experiment as before: 8K tokens, batch size 32, 8B model.
+
+![CPU and GPU activity timeline](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/cpu_gpu_phases_async.png)
+
+The timeline is almost entirely dark green: CPU and GPU running at the same time. The occasional light green slivers are moments where the GPU is active but the CPU has already finished its prep and is waiting. The near-invisible red marks the sync point between batches, where the CPU blocks to sample batch N's outputs. The GPU is active for 99.4% of total runtime, up from 76.0%. Total generation time drops from 300.6s to 234.5s, a 22% speedup. We predicted 24% if CPU overhead were fully eliminated. The small remaining gap is that unavoidable sync point. No new kernels, no model changes: letting the CPU and GPU work at the same time.
```
pcuenca:
```suggestion
The timeline is almost entirely dark green: CPU and GPU running at the same time. The occasional light green slivers are moments where the GPU is active but the CPU has already finished its prep and is waiting. The near-invisible red marks are the sync points between batches, where the CPU blocks to sample batch N's outputs. The GPU is active for 99.4% of total runtime, up from 76.0%. Total generation time drops from 300.6s to 234.5s, a 22% speedup. We predicted 24% if CPU overhead were fully eliminated. The small remaining gap is that unavoidable sync point. No new kernels, no model changes: letting the CPU and GPU work at the same time.
```

pcuenca:
```suggestion
The timeline is almost entirely dark green: CPU and GPU running at the same time. The occasional light green slivers are moments where the GPU is active but the CPU has already finished its prep and is waiting. The near-invisible red marks are the sync points between batches, where the CPU blocks to sample batch N's outputs. The GPU is active for 99.4% of total runtime, up from 76.0%. Total generation time drops from 300.6s to 234.5s, a 22% speedup. We predicted 24% if CPU overhead were fully eliminated. The small remaining gap is that unavoidable sync point. No new kernels, no model changes: letting the CPU and GPU work at the same time.
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Conclusion
+
+We started with three questions:
```
pcuenca: Perhaps we could do a higher-level recap rather than the lower-level problems we solved.

pcuenca (commented):
Very nice, did a first pass!

There's a lot of ground to cover and you do a good job at introducing the problem and diving progressively into the details. I got a bit lost during the carry-over discussion (it's easy conceptually but I lost track of the positions lol), but it's looking good!

Happy to take another look later if you want.

pcuenca (commented):
Very nice, did a first pass!

There's a lot of ground to cover and you do a good job at introducing the problem and diving progressively into the details. I got a bit lost during the carry-over discussion (it's easy conceptually but I lost track of the positions lol), but it's looking good!

Happy to take another look later if you want.

continuous_async.md
```diff
@@ -0,0 +1,235 @@
+---
+title: "Unlocking asynchronicity in continuous batching" 
+thumbnail: /blog/assets/continuous_async/thumbnail.png
+authors:
+- user: ror
+---
+
+# Unlocking asynchronicity in continuous batching
+
+![Title card](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/banner.png)
```
ariG23498, line 12: The thumbnail and the banner images are different. Was that intentional?

continuous_async.md
ariG23498: Separation of CPU and GPU workloads during inference could be done inside of different scopes. I think we should specify continuous batching.

continuous_async.md
ariG23498: Probably showcasing hosting big models using [Inference Endpoints](https://endpoints.huggingface.co/) might be a good motivation to think about using GPUs to the fullest capability.

continuous_async.md
ariG23498: I love the suggestions provided by @pcuenca! Very simple edits, but drives the point home really quickly and well.

continuous_async.md
ariG23498: @pcuenca I have tried to create the image according to your instructions. I can make some more changes if you would like, but does this fit the bill?

Image

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+Notice the red annotation on the right: after the GPU finishes computing, it goes idle. The next batch cannot start until the CPU has gone through its update step: sampling the output tokens, updating request states, re-scheduling the batch. The CPU sends the next batch to the GPU only once all of that is done. The GPU, meanwhile, waits.
+
+This is the core inefficiency of synchronous batching: the CPU and GPU take turns. While the GPU is computing, the CPU is idle. While the CPU is updating, the GPU is idle. Neither is ever doing useful work at the same time. For a single forward pass this might seem like a small price to pay, but in a continuous batching loop running hundreds of steps per second, these idle gaps accumulate into real throughput loss.
+
+To showcase this, we profile the time spent on CPU and GPU when generating 8K tokens with a batch size of 32 using an 8B model:
+
+![CPU and GPU activity timeline](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/cpu_gpu_phases_sync.png)
```
ariG23498, line 37: It would be really cool to see how we came to this plot, a small utility for plotting as a GitHub gist might do the trick. I advice this as people would then use the plots and make such plots to showcase where and how they might be affected.

continuous_async.md
ariG23498: @pcuenca do you think we should also link to the [code files](https://github.com/huggingface/transformers/tree/main/src/transformers/generation/continuous_batching) where we mention the reader to "check the code and compare"?

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Default and non-default streams
+
+If you have never explicitly used CUDA streams in PyTorch, you might be surprised they exist at all. A typical PyTorch script never mentions them, and it does not *feel* like GPU operations are asynchronous: the CPU seems to wait for the GPU to finish before moving on. That feeling is accurate, and it comes from the **default stream**.
+
+When you call a PyTorch operation without specifying a stream, it lands on the default stream. The default stream has one special property: it is **synchronizing**. If an operation is scheduled on the default stream, it waits for **all other streams to be flushed**. Ie. all work on the GPu has to be over before a single operation on the default stream can start. The reverse is also true: any operation, regardless of its stream, waits for the default stream to be flushed before it launches.  
+So if you use a non-blocking transfer on the result of a default stream operation, your CPU blocks until all GPU operations have finished, effectively destroying any effort to build concurrency.
```
ariG23498: Does a non-blocking transfer mean a "retrieve" operation?

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+That's why we need to use non-default streams. Enqueuing a kernel launch or a non-blocking memory copy returns control to the CPU immediately. The GPU will run the operation in the background, but the CPU does not wait. This answers our first question: to get back CPU control after launching GPU work, we use a non-default stream.
```
ariG23498, line 76:
```suggestion
That's why we need to use non-default streams. Enqueuing a kernel launch or a non-blocking memory copy returns control to the CPU immediately. The GPU will run the operation in the background, but the CPU does not wait.

> [!IMPORTANT]
> This answers our first question: to get back CPU control after launching GPU work, we use a non-default stream.
```

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+![Blocking vs non-blocking transfer](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/block_or_not.png)
+
+For the rest of this post, we will assume all memory transfers from one device to the other are non-blocking. We will therefore have to synchronize them ourselves.
+
+### Back to CB
```
ariG23498:
```suggestion
### Back to Continuous Batching
```

I think having full forms in the title makes it easier for the reader to navigate.

continuous_async.md
ariG23498: I think this should be an alert, something with start with `> [!NOTE]`. What do you think?

continuous_async.md
ariG23498: this was my exact reaction to this as well

I love this part, which showcases failure!

continuous_async.md
ariG23498: I was a little confused at first, but my understanding reading the rest of the paragraphs and looking at the illustration is as follows:

A CUDA event is a marker. It is like a dummy placeholder that is put inside streams to trigger some action. The operations `stream.record(event)` will trigger an alarm as soon as the stream gets hold of the event. The `stream.wait(event)` will wait till the alarm from a different stream is triggered, until it starts its current operation.

Am I thinking correctly?

continuous_async.md
ariG23498: I would like this to be fully spelled out in the titles and subtitles, this helps with navigation and also in the TOC.

ariG23498 (commented):
I am done reviewing first half of the blog post. I will be able to get the second half done by early tomorrow.

Initial verdict: I really like how the blog post is paced, while going really deep into technicalities, it never feels overwhelming. The ideas introduced are very advanced but the illustrations make it reasonable and intuitive. This is really good job! I got to intuitively understand events and streams for the very first time.

continuous_async.md
remi-or: `for inference with continuous batching` ?

continuous_async.md
remi-or: Agreed!

continuous_async.md
remi-or: Image

I don't want the text to be too small... would that fit

continuous_async.md
remi-or: Yes, that's pretty much it!

continuous_async.md
remi-or: Yes! I wanted the thumbnail to be a little snapshot of the banner, so does not have too much details. WDYT?

ariG23498: I am absolutely fine with that. I wanted to point out so that we are sure about them being different! Looks really good.

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+- user: ror
+---
+
+# Unlocking asynchronicity in continuous batching
+
+![Title card](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/banner.png)
+
+*TL;DR: we explain how to separate CPU and GPU workloads to get a massive performance boost for inference.*
+
+*This is the second post in a series on efficient LLM inference. The [first post](https://huggingface.co/blog/continuous_batching) covered continuous batching from first principles. It introduces some concepts we build upon: KV cache, FlashAttention, attention masks, etc.*
```
ariG23498, line 16: Should we also add the link of this (current) blog post to the one in the past (https://huggingface.co/blog/continuous_batching)?

This would help readers who have come across Part 1, to come to this blog post as a natural succession.

continuous_async.md
ariG23498: I think the current version is great! Thanks for the edits @remi-or

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+
+This is the key difference from synchronous batching. Before, the CPU blocked after every operation. Now it blocks once, at the very end, only to read results it genuinely needs at that moment. Everything else runs in the background.
+
+That free window between dispatching batch N and calling `d2h_done_event.synchronize()` is exactly where we can prepare batch N+1. Let us see how we can do this.
+
+## Filling the vacuum
+
+To prepare batch N+1, we can reuse the same CPU-side objects that prepared batch N. However, we need to pay attention to two things:
```
ariG23498: What do we mean by CPU-side objects? Reading a little further I think we mean the Tensors, but is that right? If so, do you think we should specify it here?

continuous_async.md
```diff
@@ -0,0 +1,235 @@
...
+- the device-side input buffers for batch N+1 cannot be the same as batch N's: we would corrupt data the GPU is still reading
+- if a request is in both batch N and N+1, and it produces a new token in the outputs of batch N, that token is needed in the inputs of batch N+1
```
ariG23498:
```suggestion
- data corruption: the device-side input buffers for batch N+1 cannot be the same as batch N's: we would corrupt data the GPU is still reading
- data transmission: if a request is in both batch N and N+1, and it produces a new token in the outputs of batch N, that token is needed in the inputs of batch N+1
```

This sets the tone for the problems that we should solve, and also names the problems beforehand.

ariG23498 (approved):
I really like the second half. I should also mention that there were moments where I had to read the paragraphs quite a number of times to understand it completely. This was due to my first time exposure to a lot of new terms. I predict that this is will also be the case for a lot of readers. To counter this, I think we should describe technical terms (buffers, slots, graphs, etc) better, and also be coherent with our wording.

This is a really well made blog post!

continuous_async.md
remi-or: Suggestion looks good to me! As for the links, they are in the conclusion, but maybe that is not enough?

continuous_async.md
remi-or: Does this render in the blogs? I removed the parenthesis.

continuous_async.md
remi-or: Since the event is scheduled from CPU, I think it's still important to have it appear in the CPU stream. Still, I can make the event more "visible" in the stream 1, here is what it could look like: 

events_proposed_2

The "wait" at start of stream 2 is a very good idea, personnaly I prefer this version: 

events_proposed

but let me know what you think is best!

continuous_async.md
remi-or: Agreed, done!

continuous_async.md
remi-or: Changed the last paragraph and transition to:

---------------------------------------

This is the key difference from synchronous batching: before, the CPU blocked after every operation. Now, it is free to do "something" while the GPU works.  
We need to figure out what that "something" is, because right now nothing changed from a GPU-utilization standpoint. 

## Filling the vacuum

The window where the CPU is available sits between dispatching batch N and dispatching batch N+1 to the GPU. It's natural use would be to prepare batch N+1's inputs, so we can dispatch them to the GPU and have them be ready once batch N compute is over. Let us see how we can do this.

---------------------

What do you think?

continuous_async.md
remi-or: Thank you!

continuous_async.md
remi-or: Not really, it just means a transfer that does not block the CPU. I tried to make the notion a bit more clear, lmk if you find it better!

continuous_async.md
remi-or: You are right! What about this:

----

To prepare batch N+1, we can reuse the same CPU-side objects that prepared batch N: the list of current requests, the state of the cache, the host-side tensor buffers, etc.

continuous_async.md
remi-or: Good catch

continuous_async.md
remi-or: Changed the conclusion from the 2 paragraphs to: 

-----

We started with a synchronous workload where the CPU and GPU worked one after the other, leaving both underused. By moving from schedule-based dependancies to data-based dependancies and refining synchronization points, we managed to detangle the CPU and GPU workloads, making parallel execution of both hardwares possible. Hence, we were able to saturate the GPU work queue and ensure it is always running. This finaly resulted in a large increase of generation speed while maintaining the accuracy of the model. Pretty much a slam dunk.

continuous_async.md
remi-or: Do you have anything in mind for this? I am not that familiar with Inference Endpoints. Agreed the intro is not that great, since it's not technical it's not my strong suit tbh

continuous_async.md
remi-or: Same question, do these render in the blogs? Otherwise, why not, but then I think I should use it a few times in the post to stay consistent.

continuous_async.md
remi-or: Yes, very good idea! Making the change.

continuous_async.md
ariG23498: I was thinking of posting a screenshot of a deployment page, and letting the reader know how much this would cost to host and use the model. I would then motivate the reader to have ways to make the most of the deployment.

What do you think?

Image

continuous_async.md
ariG23498: It does render in the blog post. Example: https://huggingface.co/blog/moe-transformers, [line](https://github.com/huggingface/blog/blob/5bbde3c78fbae666144adde9f49ba6bbc0c2b62c/moe-transformers.md?plain=1#L30).

continuous_async.md
ariG23498: I like the latter as well!

ariG23498 (approved):
Looks really good!

continuous_async.md
ariG23498: I think link in the conclusion only makes more sense. I don't want the readers to move to other tab and not come back and end reading the post.

continuous_async.md
ariG23498: Yes sir, they do render in the blog post. 🫡

continuous_async.md
ariG23498: > This is the key difference from synchronous batching: before, the CPU blocked after every operation. Noe, it is free to do "something" while the GPU works.

I think "Noe" is a typo (just wanted to point it out)

continuous_async.md
ariG23498: Yes! It makes more sense now.

continuous_async.md
ariG23498: Definitely reads more clear now.

continuous_async.md
```diff
@@ -0,0 +1,229 @@
...
+thumbnail: /blog/assets/continuous_async/thumbnail.png
+authors:
+- user: ror
+---
+
+# Unlocking asynchronicity in continuous batching
+
+![Title card](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/continuous_async/banner.png)
+
+*TL;DR: we explain how to separate CPU and GPU workloads to get a massive performance boost for inference.*
```
ariG23498, line 14:
```suggestion
Continuous Batching series:
1. [Continuous batching](https://huggingface.co/blog/continuous_batching)
2. Unlocking asynchronicity in continuous batching (_this blog post_)

*TL;DR: we explain how to separate CPU and GPU workloads to get a massive performance boost for inference.*
```

continuous_batching.md
ariG23498, line 14:
```suggestion
Continuous Batching series:
1. Continuous batching (_this blog post_)
2. [Unlocking asynchronicity in continuous batching](https://huggingface.co/blog/continuous_async)

*TL;DR: in this blog post, starting from attention mechanisms and KV caching, we derive continuous batching by optimizing for throughput.*
```

ariG23498 (commented):
I think ideating around Continuous Batching Series would be a nice addition. What do you all think?

continuous_async.md
remi-or: Maybe a thin screenshot would be better, but good idea! Nice to put concrete numbers inside the text.

continuous_async.md
remi-or: Ok, resolving this then.

continuous_async.md
remi-or: We will go with that then. Updating it!

continuous_async.md
remi-or: Ok, done! Thanks for the idea

continuous_async.md
remi-or: I am afraid it will clutter the start of the article, plus we only have two articles in this serie. Since there is already a reference to the other article in the intro, I think we can leave as is for now.

continuous_async.md
remi-or: I agree that this would be a good place to put the important tag, that said I don't really see where to put other tags like this one. It's the only point in the article where we explicitly answer a question, otherwise it all flows naturally (or at least that's what I am going for). Do you see what I mean?

continuous_async.md
remi-or: Ok, adding this!

continuous_async.md
remi-or: Just re-modeled the intro, let me know what you think!

continuous_async.md
ariG23498: I agree with you! Will resolve this.

continuous_async.md
ariG23498: I like the current version better, thanks for making the changes.

continuous_async.md
ariG23498: I see! Let's keep it as is, I will resolve this.

continuous_async.md
```diff
@@ -0,0 +1,229 @@
...
+
+When the CPU prepares a new batch, it selects which requests to include, updates the KV cache table, evicts requests that finished in the previous runs, and admits new ones to fill the freed space. Once that is done, it transfers the prepared inputs to the GPU. The GPU runs its forward pass and samples (i.e. chooses) a new token for each request. The results come back to the CPU, so it knows what token each request just produced, then the whole cycle repeats again.
+
+Notice the red annotation on the right: after the GPU finishes computing, it goes idle. The next batch cannot start until the CPU has gone through its update step: sampling the output tokens, updating request states, re-scheduling the batch.
+
+This is the core inefficiency of synchronous batching: the CPU and GPU take turns. While the GPU is computing, the CPU is idle. While the CPU is updating, the GPU is idle. In no circumstances they are both doing useful work at the same time. For a single forward pass this might seem like a small price to pay, but in a continuous batching loop running hundreds of steps per second, these idle gaps accumulate into real throughput loss.
```
sergiopaniego:
```suggestion
This is the core inefficiency of synchronous batching: the CPU and GPU take turns. While the GPU is computing, the CPU is idle. While the CPU is updating, the GPU is idle. In no circumstances are they both doing useful work at the same time. For a single forward pass this might seem like a small price to pay, but in a continuous batching loop running hundreds of steps per second, these idle gaps accumulate into real throughput loss.
```

sergiopaniego (approved):
loved it!

PR #3386, opened 2026-05-22
Add delta weight sync blogpost

AmineDiro:
### 
Sparse safetensors over HF Buckets for async RL weight sync in TRL. ~99% bf16 sparsity at RL learning rates means per-step payload drops from 1.2 GB to 20-35 MB on Qwen3-0.6B. Includes four interactive animations and a disaggregated demo running on HF Spaces.

PR #3393, opened 2026-05-27
Fix dark-mode blockquote, escape tildes, use HF LaTeX syntax

AmineDiro:
Fix to https://github.com/huggingface/blog/pull/3386


kashif (approved):

AmineDiro:
@kashif ❤️

PR #3399, opened 2026-05-30
Fix llama.cpp example model_name mismatch

### LINKED ISSUE #3398: Typo in local Reachy Mini guide: llama.cpp example uses Qwen model_name
## Summary
In local-reachy-mini-conversation.md, the llama.cpp Responses API example appears to mix two different models between Terminal 1 and Terminal 2.

## Location
Section: Option 1: llama.cpp in one terminal, speech-to-speech in the other.

## Current content
- Terminal 1 serves: ggml-org/gemma-4-E4B-it-GGUF
- Terminal 2 uses model_name: unsloth/Qwen3-4B-Instruct-2507-GGUF

## Why this is likely a typo
The example says llama.cpp is serving Gemma, but the client is configured with a Qwen model name. This looks like a copy/paste mismatch from another section.

## Proposed fix
Change Terminal 2 model_name to match Terminal 1, e.g. ggml-org/gemma-4-E4B-it-GGUF.

bixentemal:
## What
- Fixes the model mismatch in the llama.cpp + speech-to-speech example.
- Updates Terminal 2 model_name to match the model served in Terminal 1.

## Why
The section served ggml-org/gemma-4-E4B-it-GGUF in Terminal 1 but used a Qwen model in Terminal 2, which was a typo/copy-paste inconsistency.

Fixes #3398

### DIFF
--- a/local-reachy-mini-conversation.md
+++ b/local-reachy-mini-conversation.md
@@ -135,7 +135,7 @@ speech-to-speech \
   --stt parakeet-tdt \
   --tts qwen3 \
   --llm_backend responses-api \
-  --model_name "unsloth/Qwen3-4B-Instruct-2507-GGUF" \
+  --model_name "ggml-org/gemma-4-E4B-it-GGUF" \
   --responses_api_base_url "http://127.0.0.1:8080/v1"
 ```
 


pcuenca (approved):
Looks good to me, assuming the intention was to use Gemma in this example (other examples use Qwen, the tldr says Gemma)

cc @andimarafioti

julien-c (approved):
i think this should be merged no @andimarafioti?

PR #3402, opened 2026-06-02
rebuilding the `hf` CLI for AI Agents blogpost

hanouticelina:
thumbnail:
hf-cli-banner-light


hf-cli-for-agents.md
```diff
@@ -0,0 +1,214 @@
...
+curl -LsSf https://hf.co/cli/install.sh | bash
+
+# Windows (PowerShell)
+powershell -ExecutionPolicy ByPass -c "irm https://hf.co/cli/install.ps1 | iex"
+```
+
+## How we rebuilt the hf CLI for coding agents
+
+The `hf` CLI serves two users from the same commands: humans and coding agents, and they want opposite things. A human wants rich terminal output: ANSI color, padded tables truncated to fit the screen, a green `✓` on success, `✔` for booleans, progress bars, prose hints. An agent wants
+the inverse: no ANSI, nothing truncated, every value in full, compact and structured to stay light on tokens. It also can't answer a prompt and will happily re-run a command after a timeout. The rest of this section is how `hf` gives each side what it needs.
```
davanstrien:
```suggestion
the inverse: no ANSI, nothing truncated, every value in full since an agent can handle far denser output than a human, kept compact and structured to stay light on tokens. It also can't answer a prompt and will happily re-run a command after a timeout. The rest of this section is how `hf` gives each side what it needs.
```

hf-cli-for-agents.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Coding agents detection
+
+We added an agent-detection mechanism inside `huggingface_hub` so that the `hf` CLI knows whether a human or a coding agent is driving it, by reading the environment variables agents set (`CLAUDECODE`/`CLAUDE_CODE` for Claude Code, `CODEX_SANDBOX` for Codex, plus Cursor, Gemini, Pi, etc.. and the universal `AI_AGENT`). It renders the **same command** accordingly, so an agent gets agent-shaped output without passing a flag.
+
+The same command, two renderings:
```
davanstrien: Tiny thing: the `likes` differ between the two blocks I guess because they were captured at slightly different times. Since it's framed as the same command, might be nice to just force them to be matched?

hf-cli-for-agents.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+A **human** gets an aligned table, truncated to fit the terminal, plus a hint on how to see more, with color cues for status (a green `✓` on success, red on error). An **agent** gets the complete record as TSV: full repo ids, full ISO timestamps, every tag, no ANSI codes, nothing truncated, clean to parse and light on tokens. You can always force the choice with `--format human | agent | json | quiet` (`--json` and `-q` are shorthands).
+
+### Next-command hints
+
+We made `hf` end some commands with a **hint**: the exact next command to run, pre-filled with the ids you just used, so an agent can chain straight to its next step instead of working it out from scratch. A `--dry-run` hands you the line that applies it, creating a Space tells you how to watch it boot, extracting a model card tells you how to fetch just the metadata:
```
davanstrien:
```suggestion
CLI commands rarely run in isolation — one step usually implies the next (`git add`, then `git commit`). So many `hf` commands now end with a **hint**: the exact next command to run, pre-filled with the ids you just used, so a user or agent can chain straight to the next step instead of working it out from scratch. A `--dry-run` hands back the line that applies the change; creating a Space tells you how to watch it boot; extracting a model card tells you how to fetch just its metadata:
```

davanstrien (commented):
Really nice post!  Left a few small inline suggestions for language but feel free to ignore if you prefer existing! 

One bigger thing: it currently ends on the skill chart with no real conclusion. Would be good to close with a short CTA? Happy to suggest something for that if you don't have anything planned yet.

hanouticelina:
> One bigger thing: it currently ends on the skill chart with no real conclusion. Would be good to close with a short CTA? Happy to suggest something for that if you don't have anything planned yet.

thanks @davanstrien! the section is still a work in progress but yes feel free to suggest something here! 😄

hf-cli-for-agents.md
hanouticelina: addressed in [cdec62b](https://github.com/huggingface/blog/pull/3402/commits/cdec62b2ea59e060097c22cab61f1882ab4bc933)

hf-cli-for-agents.md
```diff
@@ -0,0 +1,202 @@
+---
+
+title: "How we rebuilt the hf CLI for AI Agents"
+thumbnail: /blog/assets/hf-cli-for-agents/thumbnail.png
+authors:
+
+- user: celinah
+- user: Wauplin
+
+---
```
Wauplin, line 7: (very nit)

```suggestion
---
title: "How we rebuilt the hf CLI for AI Agents"
thumbnail: /blog/assets/hf-cli-for-agents/thumbnail.png
authors:
- user: celinah
- user: Wauplin
---
```

hf-cli-for-agents.md
```diff
@@ -0,0 +1,202 @@
...
+
+# How we rebuilt the hf CLI for AI Agents
+
+`hf` is the official command-line entrypoint to the Hugging Face Hub. Anything you can do on the Hub from Python, you can do from your terminal: download and upload models, datasets and Spaces; create and manage repos, branches, tags and pull requests; run Jobs on HF infrastructure; manage Buckets, Collections, webhooks and Inference Endpoints; log in and manage your cache. One tool, the whole Hub.
```
Wauplin:
```suggestion
`hf` is the official command-line entrypoint to the Hugging Face Hub. Anything you can do on the Hub from the Python SDK, you can do from your terminal: download and upload models, datasets and Spaces; create and manage repos, branches, tags and pull requests; run Jobs on HF infrastructure; manage Buckets, Collections, webhooks and Inference Endpoints. One tool, the whole Hub.
```

I feel that "log in" at the end of the list is a bit weird. Also managing the cache is useful but not really Hub-related. I would just remove them (the sentence still reads nicely)

Wauplin: "One tool, the whole Hub" doesn't sound natural to me (though I'm not native English^^)

hf-cli-for-agents.md
```diff
@@ -0,0 +1,202 @@
...
+
+It's also, increasingly, used by **coding agents**: Claude Code, Codex, Cursor and more. So we rebuilt `hf` to make it work for both audiences at once. This is what we did, and how we benchmarked it.
```
Wauplin:
```suggestion
The `hf` CLI has been primarily built for our users over the years. But it's now increasingly used by **coding agents**: Claude Code, Codex, Cursor and more. So we rebuilt it to make it work for both audiences at once. This blog post summarizes what we did, and how we benchmarked it.
```

hf-cli-for-agents.md
```diff
@@ -0,0 +1,202 @@
...
+
+<div class="flex justify-center">
+    <img class="block dark:hidden" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/huggingface_hub/chart-skill.png" alt="Mean commands (tool calls) per run, with and without the hf-cli skill, on both agents. Claude Code (Sonnet 4.6): 10.4 without the skill, 6.9 with it. Codex (GPT-5.5): 10.1 without, 7.3 with. Fewer is better." width="100%"/>
+    <img class="hidden dark:block" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/huggingface_hub/chart-skill-dark.png" alt="Mean commands (tool calls) per run, with and without the hf-cli skill, on both agents. Claude Code (Sonnet 4.6): 10.4 without the skill, 6.9 with it. Codex (GPT-5.5): 10.1 without, 7.3 with. Fewer is better." width="100%"/>
+</div>
+
+About ten commands per task without the skill, seven with it, on both agents: fewer wrong turns and retries for the same result.
```
davanstrien:
````suggestion
About ten commands per task without the skill, seven with it, on both agents: fewer wrong turns and retries for the same result.

## Try it yourself

We benchmarked all this because we think it matters. Agents are becoming real users of the Hub: they train models, build and clean datasets, and ship demos as Spaces, almost always on behalf of a person. A Hub that works well for agents is also a Hub that works better for the people using them. The better an agent's tools, the more it can do for you. An `hf` that agents can drive as well as people is one piece of that.

The easiest way to see it in action is to point an agent at `hf`. Log in:

```bash
hf auth login
```

then hand your agent a prompt like:

```text
Run `hf --help`, then use it to list my models, datasets, and Spaces.
Take a look at how I use the Hub and suggest a few ways you could help me.
```

Watch it work out the commands on its own and come back with something useful. To make it more reliable, give it the skill: `hf skills add`.
````

Comment (davanstrien) [reply hf-cli-for-agents.md]: suggested CTA. The prompt is a bit basic but wanted something evergreen that wasn't going to go out of date to fast.

Comment (Wauplin) [inline hf-cli-for-agents.md]: (let's not forget a conclusion + call-to-action section at the end)

File: hf-cli-for-agents.md [hunk 8]
```diff
@@ -0,0 +1,202 @@
...
+```bash
+# for Codex, Cursor, OpenCode, Pi and other agents that load skills from `.agents/skills`
+hf skills add
+# includes the above + Claude Code
+hf skills add --claude
+```
+
+It installs a compact map of the whole command surface, so the agent doesn't have to rediscover it command by command.
+
+In the benchmark above, the skill's value over the bare CLI is **reliability**: the two cost about the same in tokens, but the skill lifts success (0.92 to 0.94 on Sonnet, 0.92 to 0.93 on GPT-5.5), roughly halves the false "done"s, and removes guesswork about which command and argument to reach for. The clearest single view is how many commands each run takes, with the skill and without:
```
Wauplin: I'm not sure to understand the conclusion here. I feel that the success rate is very similar even if slightly better. I'd say that "number of tool calls down from 10 to 7" is by far the best achievement of the skill. i.e. "it doesn't need to try and guess".

Also is it true that token cost is more or less the same? Feels weird if the agent does roughly 30% less tool calls

Wauplin (commented):
Super cool! I like the topic and content of the blog but I think the ordering/transitions between section could be improved to make the text flow smoothly.

For now the plan reads like this:
- `How we rebuilt the hf CLI for AI Agents` => intro, what is hf, it's used by agents, we need to address both audiences => ends by "This is what we did, and how we benchmarked it."
- `AI agent traffic on the Hub` => figure + mention Claude+Codex as main agents + "early numbers but we expect to keep growing"
- `Install` => recommended way to install the CLI => feels off to mention it here
- `How we rebuilt the hf CLI for coding agents` => explains why humans and agents want different things => I like this short part
  - though the title is very similar to the one just above
- `Coding agents detection` => mention agent-detection + env variables + the fact that 1 command == 2 renderings => I feel that "agent detection mechanism" and "1cmd==2renderings" are 2 different topics mixed together. I would rather mention how we detect agent use in the "AI Agent traffic" section so that we can focus on the rendering itself afterwards
- `Next-command hints` => makes sense (maybe a better example with Action A => Action B is more logical, here the example is Action A => "you could have done Action A differently"
- `Non-blocking and safe to retry`
   - "and anything destructive can be previewed with --dry-run first:" => not really true though (most delete commands don't have dry run. I would say commands that will take a long time have dry runs -useful independently of human/agent usage-)

- `Benchmarking the hf CLI for Coding Agents`
  -  `the setup` => describes 3 setups and introduces the "skill" => I think we should have mentioned earlier what the skill is (+it's a good occasion to talk about how auto-generated skill, designed to be token efficient, etc.)
  - `the results` => skill not mentioned anymore
    - => I think that either we should not mention the skill in the setup and leave that part for later or mention explicitly why it's not part of the results. Otherwise it reads as "3 setups, 2 results)
    - (nit) I would add titles on figures so they can easily be shared. At the moment they are not self-contained as they need the description below them to be interpreted
  - `Key findings` => yay! 🎉 
  - `The hf-cli skill` => I like this part. Maybe add how it's generated + mention the skill only at this stage.

hf-cli-for-agents.md
hanouticelina: you're right! removed in[c3c0a95](https://github.com/huggingface/blog/pull/3402/commits/c3c0a95c26abf3f8d2a5cd129ce41ca7e6454209) (not a native English either 🙈)

hf-cli-for-agents.md
hanouticelina: good catch, this one slipped through my review. you re right, leading with "reliability / success lift" oversold a tiny delta. ive reframed the paragraph around the tool call drop instead, ~10 --> ~7 calls. and fewer tool calls don't translate to fewer tokens here, because the skill is a fixed chunk of context the agent loads, so it's more like "the agent stops guessing". not 100% sure if the comparison is worth in the blogpost though, wdyt ?

hf-cli-for-agents.md
```diff
@@ -0,0 +1,211 @@
...
+Qwen/Qwen3-4B   2025-04-27T03:41:29+00:00       14808352        transformers    625     text-generation False   ['transformers', 'safetensors', 'text-generation', 'arxiv:2309.00071', 'arxiv:2505.09388', 'base_model:Qwen/Qwen3-4B-Base', 'base_model:finetune:Qwen/Qwen3-4B-Base', 'license:apache-2.0', 'endpoints_compatible', 'deploy:azure', 'region:us']
+```
+
+A **human** gets an aligned table, truncated to fit the terminal, plus a hint on how to see more, with color cues for status (a green `✓` on success, red on error). An **agent** gets the complete record as TSV: full repo ids, full ISO timestamps, every tag, no ANSI codes, nothing truncated, clean to parse and light on tokens. You can always force the choice with `--format human | agent | json | quiet` (`--json` and `-q` are shorthands).
```
Wauplin:
```suggestion
A **human** gets an aligned table, truncated to fit the terminal, plus a hint on how to see more, with color cues for status (a green `✓` on success, red on error). An **agent** gets the complete record as TSV: full repo ids, full ISO timestamps, every tag, no ANSI codes, nothing truncated, clean to parse and light on tokens.

In practice, we've implemented logging methods like `.table(...)`, `.result(...)`, `.json()`, etc., which take raw data as input and handle the formatting. In addition to human and agent modes, we've introduced `--json` and `--quiet` options to make it easier to pipe commands together. The default mode is automatically chosen based on context, but users can always force the format of their choice with `--format human | agent | json | quiet`.
```

I would add a bit more info about our primitives and the possibility to pipe commands together with --json and --quiet (otherwise they are mentioned out of the blue)

hf-cli-for-agents.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Next-command hints
+
+CLI commands rarely run in isolation — one step usually implies the next (`git add`, then `git commit`). So many `hf` commands now end with a **hint**: the exact next command to run, pre-filled with the ids you just used, so a user or agent can chain straight to the next step instead of working it out from scratch. Start a Job in the background and it points you to its logs; create a Space and it points you to its boot status:
```
Wauplin:
```suggestion
CLI commands rarely run in isolation: one step usually implies the next (`git add`, then `git commit`). Many `hf` commands now end with a **hint**: the exact next command to run, pre-filled with the IDs you just used, so a user or agent can chain straight to the next step instead of working it out from scratch. Start a Job in the background and it points you to its logs; create a Space and it points you to its boot status:
```

(nit to avoid em-dash)

hf-cli-for-agents.md
```diff
@@ -0,0 +1,211 @@
...
+</div>
+
+The bars count distinct users per agent; request volume is the sub-label. Claude Code alone is ~40k users and nearly 49M requests, with Codex close behind. These are early numbers (we only began attributing agent traffic in April 2026), but the scale is already significant, and we expect it to keep growing as coding agents become a standard way to work with the Hub.
+
+## Built for humans and agents
+
+The `hf` CLI serves two users from the same commands: humans and coding agents, and they want opposite things. A human wants rich terminal output: ANSI color, padded tables truncated to fit the screen, a green `✓` on success, `✔` for booleans, progress bars, prose hints. An agent wants
+the inverse: no ANSI, nothing truncated, every value in full since an agent can handle far denser output than a human, kept compact and structured to stay light on tokens. It also can't answer a prompt and will happily re-run a command after a timeout. The rest of this section is how `hf` gives each side what it needs.
+
+### One command, two renderings
```
Wauplin:
```suggestion
### One command, multiple renderings
```

not sure about this one (it's just that we have 4 renderings in total even though 2 of them are the "main ones")

feel free to disregard if you feel differently

hf-cli-for-agents.md
```diff
@@ -0,0 +1,211 @@
...
+
+```
+Error: Not logged in. Run `hf auth login` first.
+```
+
+Hints, warnings and errors all go to stderr while data goes to stdout, so none of this guidance pollutes the output the agent is parsing.
+
+### Non-blocking and safe to retry
+
+`hf` never sits on an interactive prompt waiting for a key an agent can't press. A destructive command still asks a human to confirm, but in agent mode it *fails fast* with the fix in the message (`Use --yes to skip confirmation.`), and `-y`/`--yes` skips it. And because agents retry on timeouts and lost context, operations are built to be safe to repeat: `hf repos create --exist-ok` is a no-op if the repo already exists, and re-running an upload re-commits cleanly. Separately, the commands that move real data take a `--dry-run` that shows exactly what they'll transfer before they run - handy for a human and an agent alike, since neither has to commit to a long download or sync blind:
```
Wauplin:
```suggestion
`hf` never sits on an interactive prompt waiting for a key an agent can't press. A destructive command still asks a human to confirm, but in agent mode it *fails fast* with the fix in the message (`Use --yes to skip confirmation.`), and `-y`/`--yes` skips it. And because agents retry on timeouts and lost context, operations are built to be safe to repeat: `hf repos create --exist-ok` is a no-op if the repo already exists, and re-running an upload re-commits cleanly. Separately, the commands that move real data take a `--dry-run` that shows exactly what they'll transfer before they run, which proves handy for a human and an agent alike, since neither has to commit to a long download or sync blind:
```

hf-cli-for-agents.md
```diff
@@ -0,0 +1,211 @@
...
+Qwen/Qwen3-4B
+```
+
+## Benchmarking the hf CLI for Coding Agents
+
+To find out whether the `hf` CLI is really more efficient for agents, we measured it. We built a small evaluation harness and ran the same set of Hub tasks through each way of driving the Hub, many times over, grading every run against the live Hub.
+
+### The setup
+
+We defined **18 non-trivial Hub tasks**. Not "download a file", but the kind of thing you'd actually ask for: aggregate a trending org's models, inspect a repo's files and their sizes, upload a folder with include/exclude rules, delete files, copy files across repos, open a PR that adds a license, create a repo with a branch and a tag, sync and prune a Bucket, build a Collection. Each task goes to a fresh coding agent with exactly **one** way to talk to the Hub:
```
Wauplin:
```suggestion
We defined **18 non-trivial Hub tasks**. Not "download a file", but the kind of thing you'd actually ask for: aggregate a trending org's models, inspect a repo's files and their sizes, upload a folder with include/exclude rules, delete files, copy files across repos, open a PR that adds a license, create a repo with a branch and a tag, sync and prune a bucket, build a collection. Each task goes to a fresh coding agent with exactly **one** way to talk to the Hub:
```

(nit)

hf-cli-for-agents.md
```diff
@@ -0,0 +1,211 @@
...
+- **On a stronger model, curl and the SDK work but stay wasteful.** On Sonnet they can't finish parts of the job (0.84; they fumble the writes). On GPT-5.5 they mostly work (0.92), hand-rolling the REST calls (or using the SDK) correctly, but still pay ~1.8× the tokens and run slower.
+
+## The hf-cli skill
+
+`hf` ships a **skill**: a compact reference of the whole command surface that an agent loads as context. It's **auto-generated** from the live `hf` command tree — one line per command (its signature, a one-line description, and the flags that matter), grouped by resource, with a short glossary of common options. It deliberately skips the self-explanatory flags so it stays terse and light on context, and it's regenerated every release. Run `hf skills preview` to print it, or install it with:
```
Wauplin:
```suggestion
`hf` ships a **skill**: a compact reference of the whole command surface that an agent loads as context. It's **auto-generated** from the live `hf` command tree, one line per command (its signature, a one-line description, and the flags that matter), grouped by resource, with a short glossary of common options. It deliberately skips the self-explanatory flags so it stays terse and light on context, and it's regenerated every release. Run `hf skills preview` to print it, or install it with:
```

(nit)

hf-cli-for-agents.md
```diff
@@ -0,0 +1,211 @@
...
+# includes the above + Claude Code
+hf skills add --claude
+```
+
+What does it buy you? Mostly, the agent stops guessing. The clearest single view is how many commands each run takes, with the skill and without:
+
+<div class="flex justify-center">
+    <img class="block dark:hidden" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/huggingface_hub/chart-skill.png" alt="Mean commands (tool calls) per run, with and without the hf-cli skill, on both agents. Claude Code (Sonnet 4.6): 10.4 without the skill, 6.9 with it. Codex (GPT-5.5): 10.1 without, 7.3 with. Fewer is better." width="100%"/>
+    <img class="hidden dark:block" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/huggingface_hub/chart-skill-dark.png" alt="Mean commands (tool calls) per run, with and without the hf-cli skill, on both agents. Claude Code (Sonnet 4.6): 10.4 without the skill, 6.9 with it. Codex (GPT-5.5): 10.1 without, 7.3 with. Fewer is better." width="100%"/>
+</div>
```
Wauplin, line 180: Above in the "hf vs curl" table we reported ~7 tool calls for `hf` vs 11-12 for `curl`. And we attributed the difference to the CLI vs SDK. But here the numbers are the same but reported as in a "skill vs no skill" setup, which feel inconsistent to me. Does that mean `hf` CLI above was run with the skill? (that would be ok, just need to be mentioned somehow IMO)

Wauplin: or maybe what you can do it drop the "it's using less tool calls" part from the table and text in the previous section to focus solely on token optimization while keeping the tool calls optimization for the skills section

i.e. have something like:

- CLI vs SDK => you are saving 2x - 6x tokens on complex tasks
- add a skill => you are saving ~30-40% tool call since the agent doesn't have to guess the commands

does that make sense?

hf-cli-for-agents.md
```diff
@@ -0,0 +1,211 @@
...
+
+On both agents that's about ten commands per task down to about seven, roughly 30% fewer tool calls - because the agent isn't probing `--help` to find the right command and argument; it already has the map. The skill won't cut your token bill (it's a fixed slice of context the agent loads up front, so tokens hold steady or tick up) and it won't make an already-reliable CLI much more reliable. What it buys is fewer wrong turns: the agent spends its budget doing the task instead of rediscovering the tool.
+
+## Try it yourself
+
+We benchmarked all this because we think it matters. Agents are becoming real users of the Hub: they train models, build and clean datasets, and ship demos as Spaces, almost always on behalf of a person. A Hub that works well for agents is also a Hub that works better for the people using them. The better an agent's tools, the more it can do for you. An `hf` that agents can drive as well as people is one piece of that.
```
Wauplin:
```suggestion
We benchmarked all this because we think it matters. Agents are becoming real users of the Hub: they train models, build and clean datasets, and ship demos as Spaces, almost always on behalf of a person. A Hub that works well for agents is also a Hub that works better for the people using them. The better an agent's tools are, the more it can do for you.
```

(just a suggestion, I feel the last sentence is a bit too much/unnatural)

hf-cli-for-agents.md
```diff
@@ -0,0 +1,211 @@
...
+Then hand it the skill, so it knows the whole command surface from the first turn:
+
+```bash
+hf skills add            # Codex, Cursor, OpenCode, Pi and other agents that load skills from .agents/skills
+hf skills add --claude   # the above + Claude Code
+```
+
+The full command reference lives in the [`hf` CLI guide](https://huggingface.co/docs/huggingface_hub/guides/cli).
+
+And if you build a coding agent yourself, **register your harness**: that's how `hf` learns to detect it and how the Hub attributes its traffic to you by name, it's how the agents in the chart at the top of this post show up in our numbers. It's a small PR adding an entry to [`agent-harnesses.ts`](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/agent-harnesses.ts) (a display label, optional repo/docs links, and how to detect your harness from the environment); the [Register your agent harness](https://huggingface.co/docs/hub/agents-overview#register-your-agent-harness) guide walks through the fields.
```
Wauplin:
```suggestion
Building an agent harness yourself? **Get it registered!** That's how `hf` learns to detect it, and how the Hub attributes its traffic to you. You simply need to open a small PR adding an entry to [`[agent-harnesses.ts](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/agent-harnesses.ts)`](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/agent-harnesses.ts). Read the [[Register your agent harness](https://huggingface.co/docs/hub/agents-overview#register-your-agent-harness)](https://huggingface.co/docs/hub/agents-overview#register-your-agent-harness) guide for more details.
```

alternative suggestion

Wauplin (commented):
Thanks for the iteration! I feel the blog reads more smoothly like this. Left some additional comments, mostly suggestions

hf-cli-for-agents.md
Wauplin: see suggestion here: https://github.com/huggingface/blog/pull/3402#discussion_r3348521185

I think it's great to have "use hf to save context, add a skill to run fewer tool calls" (if that happens to be the case^^)

(and yes, haven't thought about the additional fixed context size of the skill, makes sense)

hf-cli-for-agents.md
```diff
@@ -0,0 +1,213 @@
...
+Then hand it the skill, so it knows the whole command surface from the first turn:
+
+```bash
+hf skills add            # Codex, Cursor, OpenCode, Pi and other agents that load skills from .agents/skills
+hf skills add --claude   # the above + Claude Code
+```
+
+The full command reference lives in the [`hf` CLI guide](https://huggingface.co/docs/huggingface_hub/guides/cli).
+
+Building an agent harness yourself? **Get it registered!** That's how `hf` learns to detect it, and how the Hub attributes its traffic to you. You simply need to open a small PR adding an entry to [`[agent-harnesses.ts](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/agent-harnesses.ts)`](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/agent-harnesses.ts). Read the [[Register your agent harness](https://huggingface.co/docs/hub/agents-overview#register-your-agent-harness)](https://huggingface.co/docs/hub/agents-overview#register-your-agent-harness) guide for more details.
```
davanstrien:
```suggestion
Building an agent harness? **Get it registered!** That's how `hf` learns to detect it, and how the Hub attributes its traffic to you. You simply need to open a small PR adding an entry to [`[agent-harnesses.ts](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/agent-harnesses.ts)`](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/agent-harnesses.ts). Read the [[Register your agent harness](https://huggingface.co/docs/hub/agents-overview#register-your-agent-harness)](https://huggingface.co/docs/hub/agents-overview#register-your-agent-harness) guide for more details.
```

hf-cli-for-agents.md
```diff
@@ -0,0 +1,213 @@
...
+</div>
+
+The bars count distinct users per agent; request volume is the sub-label. Claude Code alone is ~40k users and nearly 49M requests, with Codex close behind. These are early numbers (we only began attributing agent traffic in April 2026), but the scale is already significant, and we expect it to keep growing as coding agents become a standard way to work with the Hub.
+
+## Built for humans and agents
+
+The `hf` CLI serves two users from the same commands: humans and coding agents, and they want opposite things. A human wants rich terminal output: ANSI color, padded tables truncated to fit the screen, a green `✓` on success, `✔` for booleans, progress bars, prose hints. An agent wants
+the inverse: no ANSI, nothing truncated, every value in full since an agent can handle far denser output than a human, kept compact and structured to stay light on tokens. It also can't answer a prompt and will happily re-run a command after a timeout. The rest of this section is how `hf` gives each side what it needs.
```
davanstrien:
```suggestion
the inverse: no ANSI, nothing truncated, every value in full since an agent can handle far denser output than a human, kept compact and structured to stay light on tokens. It also can't answer a CLI prompt and will happily re-run a command after a timeout. The rest of this section is how `hf` gives each side what it needs.
```

davanstrien: nit but maybe clearer

hf-cli-for-agents.md
```diff
@@ -0,0 +1,213 @@
...
+```
+$ hf models ls --help
+...
+Examples
+  $ hf models ls --sort downloads --limit 10
+  $ hf models ls --search "qwen" --author Qwen
+  $ hf models ls Qwen/Qwen3-4B --tree
+```
+
+The command tree is consistent, **resource + verb** with the obvious aliases (`hf models ls`, `hf repos create`, `hf jobs ps`, `hf collections delete`; `list`/`ls`, `remove`/`rm`), so once an agent learns one command it can guess the rest. And the output composes: `-q` prints one id per line to pipe into the next command, `--json` gives you something to hand to `jq`.
```
davanstrien:
```suggestion
The command tree is consistent, **resource + verb** with the obvious aliases (`hf models ls`, `hf repos create`, `hf jobs ps`, `hf collections delete`; `list`/`ls`, `remove`/`rm`), so once an agent learns one command it can guess the rest. And the output composes: `-q` prints one id per line to pipe into the next command, `--json` gives you something to hand to [`jq`](https://jqlang.org/).
```

davanstrien: maybe overkill but for newer vibe coders who may not know these classic tools!

hf-cli-for-agents.md
```diff
@@ -0,0 +1,213 @@
...
+    <img class="hidden dark:block" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/huggingface_hub/chart-skill-dark.png" alt="Mean commands (tool calls) per run, with and without the hf-cli skill, on both agents. Claude Code (Sonnet 4.6): 10.4 without the skill, 6.9 with it. Codex (GPT-5.5): 10.1 without, 7.3 with. Fewer is better." width="100%"/>
+</div>
+
+On both agents that's about ten commands per task down to about seven, roughly 30% fewer tool calls - because the agent isn't probing `--help` to find the right command and argument; it already has the map. The skill won't cut your token bill (it's a fixed slice of context the agent loads up front, so tokens hold steady or tick up) and it won't make an already-reliable CLI much more reliable. What it buys is fewer wrong turns: the agent spends its budget doing the task instead of rediscovering the tool.
```
davanstrien:
```suggestion
On both agents that's about ten commands per task down to about seven, roughly 30% fewer tool calls - because the agent isn't probing `--help` to find the right command and argument; it already has the map. The skill won't cut your token bill (it's a fixed slice of context the agent loads up front, so tokens hold steady or tick up) and it won't make an already-reliable CLI much more reliable. What it buys is fewer wrong turns: the agent spends its budget doing the task instead of rediscovering the tool. This could be particularly helpful when using `hf` with local models.
```

davanstrien: feel free to ignore but imo could be nice to lightly flag somewhere

hf-cli-for-agents.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+Then hand it the skill, so it knows the whole command surface from the first turn:
+
+```bash
+hf skills add            # Codex, Cursor, OpenCode, Pi and other agents that load skills from .agents/skills
+hf skills add --claude   # the above + Claude Code
+```
+
+The full command reference lives in the [`hf` CLI guide](https://huggingface.co/docs/huggingface_hub/guides/cli).
```
davanstrien, line 216: Wonder if it's still worth adding something very minimal like: then ask your agent how it might help you work with the Hub or something? just to prompt people to try something real and not just stop at adding the skill?

davanstrien (commented):
Looking great! Added a few small suggestions but fine to ignore! Main bigger suggestion would be to maybe still have some kind of suggestion for a first prompt to try with an agent after adding the Skill.

hf-cli-for-agents.md
hanouticelina: make sense to me! addressed in https://github.com/huggingface/blog/pull/3402/changes/057a3800bd74c20456b34979d7a793e13c1e259b..a786a2dc0827abf07ab6eb11f7b14a87ba92a0ca

hf-cli-for-agents.md
hanouticelina: yes! https://github.com/huggingface/blog/pull/3402/changes/b4387631e2b802ed83eb7eeff7d144f4ff0a35a7

hf-cli-for-agents.md
```diff
@@ -0,0 +1,221 @@
...
+```text
+Run `hf --help`, then use it to list my models, datasets, and Spaces.
+Take a look at how I use the Hub and suggest a few ways you could help me.
+```
+
+It'll work out the commands on its own and come back with something useful.
+
+The full command reference lives in the [`hf` CLI guide](https://huggingface.co/docs/huggingface_hub/guides/cli).
+
+Building an agent harness? **Get it registered!** That's how `hf` learns to detect it, and how the Hub attributes its traffic to you. You simply need to open a small PR adding an entry to [`[agent-harnesses.ts](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/agent-harnesses.ts)`](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/agent-harnesses.ts). Read the [[Register your agent harness](https://huggingface.co/docs/hub/agents-overview#register-your-agent-harness)](https://huggingface.co/docs/hub/agents-overview#register-your-agent-harness) guide for more details.
```
Wauplin:
```suggestion
## Register an agent harness

Building an agent harness? **Get it registered!** That's how `hf` learns to detect it, and how the Hub attributes its traffic to you. You simply need to open a small PR adding an entry to [`agent-harnesses.ts`](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/agent-harnesses.ts). Read the [Register your agent harness](https://huggingface.co/docs/hub/agents-overview#register-your-agent-harness) guide for more details.
```

worth a title?  (+ fixes the markdown link)

Wauplin (approved):
Last review on my side I think, thanks for the iterations! 😃

davanstrien (approved):

_blog.yml
```diff
@@ -5376,4 +5376,12 @@
     - profile
     - CUDA
     - open-source
-    - kernels
\ No newline at end of file
+    - kernels
+
+- local: hf-cli-for-agents
+  date: June 3, 2026
```
pcuenca, line 5382:
```suggestion
  date: June 4, 2026
```

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+
+The `hf` CLI has been primarily built for our users over the years. But it's now increasingly used by **coding agents**: Claude Code, Codex, Cursor and more. So we rebuilt it to make it work for both audiences at once. This blog post summarizes what we did, and how we benchmarked it.
+
+## AI agent traffic on the Hub
+
+We started tracking agent usage of the Hub in April 2026. The `hf` CLI (and the `huggingface_hub` Python SDK it's built on) detects when a coding agent is driving it by reading the environment variables agents set: `CLAUDECODE`/`CLAUDE_CODE` for Claude Code, `CODEX_SANDBOX` for Codex, plus Cursor, Gemini, Pi, and the universal `AI_AGENT`. That single signal does two jobs: it shapes the CLI's output (more on that below) and it tags each Hub request with an `agent/<name>` user-agent, so we can attribute traffic to the agent driving it. The two largest by distinct users are **Claude Code and Codex**, well ahead of everything else, and they're the two agents we benchmark later in this article.
+
+<div class="flex justify-center">
```
pcuenca, line 19: I think we don't need the two variants of the image in the blog post (just the light one)

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+    <img class="hidden dark:block" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/huggingface_hub/chart-users-dark.png" alt="Distinct users of the Hugging Face Hub by coding agent since April 2026. Claude Code leads with 39.5k users and 48.6M requests, then Codex with 34.8k users and 36.4M requests, followed by antigravity, cursor-cli, openclaw, cursor, gemini and pi." width="100%"/>
+</div>
+
+The bars count distinct users per agent; request volume is the sub-label. Claude Code alone is ~40k users and nearly 49M requests, with Codex close behind. These are early numbers (we only began attributing agent traffic in April 2026), but the scale is already significant, and we expect it to keep growing as coding agents become a standard way to work with the Hub.
+
+## Built for humans and agents
+
+The `hf` CLI serves two users from the same commands: humans and coding agents, and they want opposite things. A human wants rich terminal output: ANSI color, padded tables truncated to fit the screen, a green `✓` on success, `✔` for booleans, progress bars, prose hints. An agent wants
```
pcuenca:
```suggestion
Humans and coding agents expect different outputs for the same `hf` commands. A human wants rich terminal output: ANSI color, padded tables truncated to fit the screen, a green ✅ on success, `✔` for booleans, progress bars, prose hints. An agent wants
```

(I think we have already established we cater to the two)

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+the inverse: no ANSI, nothing truncated, every value in full since an agent can handle far denser output than a human, kept compact and structured to stay light on tokens. It also can't answer a CLI prompt and will happily re-run a command after a timeout. The rest of this section is how `hf` gives each side what it needs.
+
+### One command, multiple renderings
+
+Because `hf` detects when an agent is driving it (the same signal behind the traffic numbers above), it renders the **same command** two ways, so an agent gets agent-shaped output without passing a flag:
```
pcuenca:
```suggestion
When `hf` auto-detects agent use (via the environment variables mentioned above), it renders the **same command** differently. It optimizes output format for humans or agents without passing a flag:
```

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+
+```
+Error: Not logged in. Run `hf auth login` first.
+```
+
+Hints, warnings and errors all go to stderr while data goes to stdout, so none of this guidance pollutes the output the agent is parsing.
+
+### Non-blocking and safe to retry
+
+`hf` never sits on an interactive prompt waiting for a key an agent can't press. A destructive command still asks a human to confirm, but in agent mode it *fails fast* with the fix in the message (`Use --yes to skip confirmation.`), and `-y`/`--yes` skips it. And because agents retry on timeouts and lost context, operations are built to be safe to repeat: `hf repos create --exist-ok` is a no-op if the repo already exists, and re-running an upload re-commits cleanly. Separately, the commands that move real data take a `--dry-run` that shows exactly what they'll transfer before they run, which proves handy for a human and an agent alike, since neither has to commit to a long download or sync blind:
```
pcuenca:
```suggestion
`hf` never sits on an interactive prompt waiting for a key an agent can't press. A destructive command still asks a human to confirm, but in agent mode it *fails fast* with the fix in the message (`Use --yes to skip confirmation.`), and `-y`/`--yes` skips it. And because agents retry on timeouts and lost context, operations are built to be safe to repeat: `hf repos create --exist-ok` is a no-op if the repo already exists, and re-running an upload re-commits cleanly. Separately, the commands that move real data take a `--dry-run` that shows exactly what they'll transfer before they run, which proves handy for humans and  agents alike, since neither has to commit to a long download or blind sync:
```

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+`hf` is built to be probed: run `hf` to see the resource groups, run `--help` on the one you need, and every `--help` ends with real, copy-pasteable examples (which an agent matches against far faster than it parses a description):
+
+```
+$ hf models ls --help
+...
+Examples
+  $ hf models ls --sort downloads --limit 10
+  $ hf models ls --search "qwen" --author Qwen
+  $ hf models ls Qwen/Qwen3-4B --tree
+```
```
pcuenca, line 104: Super cool. Perhaps it'd be useful to adjust the top-level `--help` to explain it's optimized for agents (so TSVs will be used, no interactive confirmation will be asked, --dry-run exists). I've seen agents use it when they're new to a command.

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+
+The config is deliberately clean: a fresh instance per run, no custom MCP servers, no `CLAUDE.md` or `AGENTS.md`, nothing in context to nudge behavior. The task and the tool go into a single prompt, and the agent finishes with a `TASK_COMPLETE` or `TASK_FAILED` marker, but we don't trust that marker (an agent will report success on work that never landed), so we grade every run independently by **re-querying the live Hub**: did the branch really get created, is the file actually gone, does the bucket exist? Each task/tool combination is run **10 times**, since coding agents are non-deterministic, about **520 runs per agent** (18 tasks × 3 tools × 10 reps, minus a cap on one billable Jobs task) and ~1,000 graded runs in total. We ran the whole thing twice, on the two most popular coding agents (**Claude Code** with Sonnet 4.6 and **OpenAI Codex** with GPT-5.5).
+
+### The results
+
+The ranking came out **identical on both agents**:
+
+
+| agent                        | tool        | mean score | token tax | false "done" |
```
pcuenca:
```suggestion
| agent                        | tool        | mean score | token usage | self-report error |
```

maybe?

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+| ---------------------------- | ----------- | ---------- | ---------------- | ------------ |
+| **Claude Code (Sonnet 4.6)** | `hf` CLI    | **0.94**   | baseline         | **2 / 163**  |
+|                              | curl / Python SDK | 0.84       | **1.6× tokens**  | 11 / 163     |
+| **Codex (GPT-5.5)**          | `hf` CLI    | **0.93**   | baseline         | **3 / 163**  |
+|                              | curl / Python SDK | 0.92       | **1.8× tokens**  | 10 / 163     |
```
pcuenca: Amazing, I was thinking about this while reading the previous sections. A couple of comments:
- Perhaps we could show the table after a small intro in the benchmarking section, before describing the setup in full detail. I was itching to see it while reading "The setup".
- I'd mention that we measured results in one of the intro sections. If we could convey a simple message "we saw up to xx token reduction for complex tasks", or "successful task completion improved by xx%" (or both), that'd be awesome. It could even go in the title.
- Curious about the difference (in tokens, perhaps) between the auto-inferred `--agent` and the default output format.

hf-cli-for-agents.md
pcuenca: Maybe also `success score` rather than `mean score`

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+
+Two pictures carry the result. First, **task success on Sonnet**, the agent where curl and the SDK struggle most:
+
+<div class="flex justify-center">
+    <img class="block dark:hidden" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/huggingface_hub/chart-success.png" alt="Task success on Claude Code with Sonnet 4.6: hf CLI 94%, curl / Python SDK 84%." width="100%"/>
+    <img class="hidden dark:block" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/huggingface_hub/chart-success-dark.png" alt="Task success on Claude Code with Sonnet 4.6: hf CLI 94%, curl / Python SDK 84%." width="100%"/>
+</div>
+
+Without the CLI, curl and the SDK trail by ten points, because on Sonnet they simply can't finish parts of the job (the writes, mostly), while the `hf` CLI clears them. Second, and this is the one to look at, **their token tax on GPT-5.5**, broken down per task. Each bar is the curl/SDK tokens divided by the CLI's on the same task, so `2.4×` means they burned 2.4 times as many tokens to do the same thing:
```
pcuenca:
```suggestion
Without the CLI, curl and the SDK trail by ten points, because on Sonnet they simply can't finish parts of the job (the writes, mostly), while the `hf` CLI clears them.

The second image shows **token impact on GPT-5.5**, broken down per task. Each bar is the curl/SDK tokens divided by the CLI's on the same task, so `2.4×` means the non-hf version burned 2.4 times as many tokens to do the same thing:
```

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+</div>
+
+The shape is the whole point. On a one-shot read (count dataset rows, batch metadata) curl and the SDK are fine, sometimes even lighter. But the moment a task is *real work*, several dependent calls that each need the previous result, the agent has to hand-roll the entire chain of REST calls (or dig through the SDK) and the cost blows up: **2.4× to 6× the CLI** on creating a repo with a branch and tag, deleting files, copying across repos, syncing a bucket. The `hf` CLI folds that chain into one command. This is also exactly why a benchmark built on easy reads makes curl and the SDK look better than they really are: it never reaches the tasks where the CLI pulls away.
```
pcuenca:
```suggestion
On a one-shot read (count dataset rows, batch metadata) curl and the SDK are fine, and sometimes lighter. But as tasks get more complex and involve several dependent steps, the agent has to hand-roll the entire chain of REST calls (or dig through the SDK) and the cost blows up: **2.4× to 6× the CLI's** on creating a repo with a branch and tag, deleting files, copying across repos, or syncing a bucket. The `hf` CLI allows the agent to issue the task as a single piped command, rather than crafting a complex workflow.
```

(this paragraph sounded llm-y to me)

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Key findings
+
+- **The `hf` CLI is far leaner than curl or the SDK.** For the same task, at equal-or-better success, curl and the SDK burn **1.6× the tokens on Sonnet (302k vs 194k) and 1.8× on GPT-5.5 (346k vs 191k)**. On easy reads they're fine, but on real multi-step work they pay **2× to 6×**: the CLI folds a chain of REST calls into one command, while curl or the SDK re-derives that chain by hand every run.
```
pcuenca:
```suggestion
- **The `hf` CLI is far leaner than curl or the SDK.** For the same task, at equal-or-better success, curl and the SDK burn **1.6× the tokens on Sonnet (302k vs 194k) and 1.8× on GPT-5.5 (346k vs 191k)**. On easy reads they're fine, but on real multi-step work they pay **2× to 6×**: the CLI composes a set of REST calls into one command, while curl or the SDK re-derives the chain by hand every run.
```

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+- **On a stronger model, curl and the SDK work but stay wasteful.** On Sonnet they can't finish parts of the job (0.84; they fumble the writes). On GPT-5.5 they mostly work (0.92), hand-rolling the REST calls (or using the SDK) correctly, but still pay ~1.8× the tokens.
```
pcuenca:
```suggestion
- **On a stronger model, curl and the SDK work but stay wasteful.** On Sonnet they can't finish parts of the job (0.84 success average). On GPT-5.5 they mostly work (0.92), hand-rolling the REST calls (or using the SDK) correctly, but still pay ~1.8× the tokens.
```

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+
+## The hf-cli skill
+
+`hf` ships a **skill**: a compact reference of the whole command surface that an agent loads as context. It's **auto-generated** from the live `hf` command tree, one line per command (its signature, a one-line description, and the flags that matter), grouped by resource, with a short glossary of common options. It deliberately skips the self-explanatory flags so it stays terse and light on context, and it's regenerated every release. Run `hf skills preview` to print it, or install it with:
```
pcuenca, line 166: It has been adapted to show `--agent`-formatted invocations and results, I assume?

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+# for Codex, Cursor, OpenCode, Pi and other agents that load skills from `.agents/skills`
+hf skills add
+# includes the above + Claude Code
+hf skills add --claude
+```
+
+What does it buy you? Mostly, the agent stops guessing. The clearest single view is how many commands each run takes, with the skill and without:
+
+<div class="flex justify-center">
+    <img class="block dark:hidden" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/huggingface_hub/chart-skill.png" alt="Mean commands (tool calls) per run, with and without the hf-cli skill, on both agents. Claude Code (Sonnet 4.6): 10.4 without the skill, 6.9 with it. Codex (GPT-5.5): 10.1 without, 7.3 with. Fewer is better." width="100%"/>
```
pcuenca, line 178: But then it would also use less tokens, wouldn't it? What's the token impact between using / not using the skill?

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+    <img class="hidden dark:block" src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/huggingface_hub/chart-skill-dark.png" alt="Mean commands (tool calls) per run, with and without the hf-cli skill, on both agents. Claude Code (Sonnet 4.6): 10.4 without the skill, 6.9 with it. Codex (GPT-5.5): 10.1 without, 7.3 with. Fewer is better." width="100%"/>
+</div>
+
+On both agents that's about ten commands per task down to about seven, roughly 30% fewer tool calls - because the agent isn't probing `--help` to find the right command and argument; it already has the map. The skill won't cut your token bill (it's a fixed slice of context the agent loads up front, so tokens hold steady or tick up) and it won't make an already-reliable CLI much more reliable. What it buys is fewer wrong turns: the agent spends its budget doing the task instead of rediscovering the tool. This could be particularly helpful when using `hf` with local models.
```
pcuenca: Ah I see. Wouldn't it amortize over multiple uses in the same session? I assume you were running tests in isolation; we could have seen lower token usage with the skill by running tasks within the same session (I'm guessing).

pcuenca:
```suggestion
On both agents that's about ten commands per task down to about seven, roughly 30% fewer tool calls - because the agent isn't probing `--help` to find the right command and argument. The skill won't cut your token bill, because it prepends a fixed slice of info to the context, so tokens remain about the same or slightly tick up for the same task. The Skill won't make the CLI more reliable either, but it will help the agent spend time running your task rather than finding out how the tool works. This could be particularly helpful when using `hf` with local models.
```

(or in long sessions, I'd say)

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+```bash
+hf skills add            # Codex, Cursor, OpenCode, Pi and other agents that load skills from .agents/skills
+hf skills add --claude   # the above + Claude Code
+```
+
+Then point your agent at the Hub and let it work. Make sure you're logged in (`hf auth login`), then hand it a prompt like:
+
+```text
+Run `hf --help`, then use it to list my models, datasets, and Spaces.
+Take a look at how I use the Hub and suggest a few ways you could help me.
```
pcuenca: I don't understand this prompt. Why do we instruct it to use `--help` when it already has the skill? What does it mean to "take a look at how I use the Hub"?

hf-cli-for-agents.md
```diff
@@ -0,0 +1,223 @@
...
+```
+
+It'll work out the commands on its own and come back with something useful.
+
+The full command reference lives in the [`hf` CLI guide](https://huggingface.co/docs/huggingface_hub/guides/cli).
+
+## Register an agent harness
+
+Building an agent harness? **Get it registered!** That's how `hf` learns to detect it, and how the Hub attributes its traffic to you. You simply need to open a small PR adding an entry to [`agent-harnesses.ts`](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/agent-harnesses.ts). Read the [Register your agent harness](https://huggingface.co/docs/hub/agents-overview#register-your-agent-harness) guide for more details.
```
pcuenca:
```suggestion
Building an agent harness? **Get it registered!** That's how `hf` learns to detect it, and how the Hub attributes its traffic to your harness. You simply need to open a small PR adding an entry to [`agent-harnesses.ts`](https://github.com/huggingface/huggingface.js/blob/main/packages/tasks/src/agent-harnesses.ts). Read the [Register your agent harness](https://huggingface.co/docs/hub/agents-overview#register-your-agent-harness) guide for more details.
```

pcuenca (approved):
Fantastic. Some general thoughts:

- We could explore being more ambitious with the title: "cut token usage with an agent optimized CLI", "desigining hf CLI for agents improves quality and reduces cost", ...
- Anticipate the results earlier, as part of the intro (just a summary, no need to present the full table)
- Maybe mention the `hf` version where this was added

---

Edit re: title. Not looking at hyping the post, but I think it has general value for anyone building tools. "agent-focused CLIs as optimal tool interfaces" is strong.

hf-cli-for-agents.md
hanouticelina: I added it because it looks better in dark theme:
Image

hf-cli-for-agents.md
davanstrien: I think it might be useful for weaker agents but could also remove the help part and do something like:

```
Use hf to list my Hugging Face Hub models, datasets, and Spaces.
Take a look at how I am using the Hub and suggest a few ways you could help me.
```

For the prompt:

> "Take a look at how I use the Hub and suggest a few ways you could help me."

I wanted something that would increase the chances the agent might actually do 
something useful for the user rather than a very toy example (without making it too specific). Ideally the agent would do something like "seems you have a lot of datasets without any readme, maybe we could add minimal docs for some of your datasets?" or similar. 

Open to other suggestions. Can also just do tell me about the current trending models but IMO its nicer to try and get the agent to actually help user with using the Hub rather than just listing stuff.

hf-cli-for-agents.md
pcuenca: Ah, I thought the blog only worked in light mode lol. I have my system set to light, but Chrome set to dark, and the blog always renders in light mode 🤷 Sorry for the noise!

hf-cli-for-agents.md
pcuenca: Ah sorry, I read it too fast; I thought the first line was just instructions on how to use hf, then the query. It makes sense now. I'd lean towards removing the `--help` in this case.

Perhaps we could add some small hook after the prompt, like the examples you showed: 'the agent may offer to add model cards to your empty repos!'

hf-cli-for-agents.md
```diff
@@ -0,0 +1,225 @@
...
+hf skills add            # Codex, Cursor, OpenCode, Pi and other agents that load skills from .agents/skills
+hf skills add --claude   # the above + Claude Code
+```
+
+Then point your agent at the Hub and let it work. Make sure you're logged in (`hf auth login`), then hand it a prompt like:
+
+```text
+Run `hf --help`, then use it to list my models, datasets, and Spaces.
+Take a look at how I use the Hub and suggest a few ways you could help me.
+```
```
davanstrien:
```suggestion
```text
Use `hf` to list my Hugging Face Hub models, datasets, and Spaces.
Take a look at how I am currently using the Hub and suggest a few ways you could help me.
```

After this prompt, the agent may offer to add model cards to repos that lack a `README.md`. You can, of course, prompt the agent to help you with something else: finding the best model for a task, exploring new datasets, training models via Jobs, etc. 
```

hf-cli-for-agents.md
davanstrien: possible fix: https://github.com/huggingface/blog/pull/3402/changes#r3356410812

hf-cli-for-agents.md
hanouticelina: yes! addressed in [875d971](https://github.com/huggingface/blog/pull/3402/commits/875d971aef5f9927d7e13d740ceb5222e8a13355)

hf-cli-for-agents.md
hanouticelina: no, the skill is just discovery reference (command signatures + the flags + a short options glossary), it has no `--format agent` examples. the CLI auto detects when an agent is running the commands and emits agent-formatted output

hf-cli-for-agents.md
hanouticelina: No, the skill costs +5–11% tokens, even though tool calls drop ~25%. It's a fixed slice of context that's resent every turn (we say exactly this in the following paragraph)

hf-cli-for-agents.md
hanouticelina: yes, we ran each task in a fresh session. I added a one-line honesty in the skill section in [c6da0bc](https://github.com/huggingface/blog/pull/3402/commits/c6da0bc5d0db64b61db48690197d2b7b62a21374)

hf-cli-for-agents.md
hanouticelina: addressed in [875d971](https://github.com/huggingface/blog/pull/3402/commits/875d971aef5f9927d7e13d740ceb5222e8a13355)

PR #3434, opened 2026-06-23
docs: update local models triage authors

osolmaz:
## Summary

The local models triage post needs a small follow-up after landing on `main`.
This change narrows the author list and uses the more precise phrase `zero-shot classify` in the conclusion.
It keeps the correction separate from the original blog PR so reviewers can check the final metadata and wording directly.

## What Changed

Only the blog post metadata and one sentence changed.
The body of the post, figures, benchmark table, and publication metadata are unchanged.

- Removed `pcuenq` and `lysandre` from the frontmatter author list.
- Changed `one-shot classify` to `zero-shot classify` in the conclusion.

## Testing

I checked that the patch is clean and limited to the intended Markdown file.
I did not run the full blog YAML validator because this follow-up does not touch `_blog.yml`.

- `git diff --check`

## Risks

Risk is low because this only changes author metadata and one wording choice.
The main thing to verify is that the final author list is intentional.

### DIFF
--- a/local-models-pr-triage.md
+++ b/local-models-pr-triage.md
@@ -9,11 +9,11 @@ authors:
 
 # We got local models to triage the OpenClaw repo for FREE!*
 
-*\*Free as in beer, excluding the cost of electricity, and assuming you already own the hardware*
+<em>*Free as in beer, excluding the cost of electricity, and assuming you already own the hardware</em>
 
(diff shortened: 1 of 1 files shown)


burtenshaw (approved):

osolmaz:
Just to clarify, I used the following rule for author inclusion: if someone has contributed directly, is cited or gave feedback on the post

PR #3438, opened 2026-06-25
Review from Stas on Profiling Part 1

ariG23498:
CC: @sergiopaniego @pcuenca


torch-mlp-fusion.md
```diff
@@ -13,6 +13,39 @@ authors:
 
 ![Thumbnail of the blog post](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-mlp-fusion/thumbnail.png)
 
+<div style="
```
ariG23498, line 16: Looks something like this.

Image

torch-profiler.md
```diff
@@ -355,7 +382,7 @@ Figure 18 reveals a similar finding. While each kernel was the exact same, they
 
 A reader who only saw the average would conclude that a matmul took ~1 ms (mean of 5 = 1084 µs); a reader who looked at the trace would see that the matmul takes ~580 µs except when the GPU throws a fit. Those are very different mental models, and only one of them is correct.
 
-## Let's see some torch compile at work
+## Let's see some `torch.compile` at work
```
sergiopaniego: not sure this will render correctly 🤔

sergiopaniego (approved):
love it!
this is proof that consistency and hard work *can* yield good results :)

torch-profiler.md
ariG23498: image

Which one do you like better?

PR #3451, opened 2026-07-06
[Profiling in PyTorch] Part 3: Attention is all you profile

ariG23498:
CC: @sayakpaul @remi-or @sergiopaniego @pcuenca


torch-attention-profile.md
```diff
@@ -0,0 +1,410 @@
...
+
+> [!NOTE]
+> The scripts for this blog post live here: [`04_a_naive_attention.py`](https://huggingface.co/datasets/ariG23498/profiling-pytorch/blob/main/04_a_naive_attention.py), [`04_b_inplace_ops_attention.py`](https://huggingface.co/datasets/ariG23498/profiling-pytorch/blob/main/04_b_inplace_ops_attention.py), [`04_c_sdpa_attention.py`](https://huggingface.co/datasets/ariG23498/profiling-pytorch/blob/main/04_c_sdpa_attention.py), and [`04_d_kernels_attention.py`](https://huggingface.co/datasets/ariG23498/profiling-pytorch/blob/main/04_d_kernels_attention.py). Like before, it helps to open them in a separate tab and walk through the code as you read. We use an `NVIDIA A100-SXM4-80GB` GPU to run the scripts. It is really easy to set up a GPU on the Hugging Face infrastructure and experiment with the scripts using [Dev Mode with Spaces](https://huggingface.co/docs/hub/spaces-dev-mode). One could also run the scripts with the [Hugging Face Jobs pipeline](https://huggingface.co/docs/huggingface_hub/en/guides/jobs).
```
sergiopaniego, line 57: If you go to the scripts for this blog, they have:

```python
os.environ["CUDA_VISIBLE_DEVICES"]="1"
```

we could update this lines since otherwise it could cause errors if the scripts are launched directly on a machine with 1 gpu (eg., jobs)

torch-attention-profile.md
```diff
@@ -0,0 +1,410 @@
...
+- `128x64x64`: the tile dimensions it chose.
+
+That one fact, _generated per problem_, explains everything else that looks unusual in the trace.
+
+1. No transposes: The CPU lane goes from `_cudnn_attention_forward` straight to a couple of `aten::empty` allocations and then the kernel, with zero `aten::transpose` (Figures 19, 20 and 21). Flash and efficient each insert four (metadata) transposes to reshape the tensors while cuDNN consumes the native `[B, H, S, D]` layout directly because its generator emits a kernel for that layout.
+
+   | Variant | Trace |
+   | :--: | :--: |
+   | Figure 19: Flash | ![CPU lane of the flash backend showing four aten::transpose ops before the fused attention kernel](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/flash-transpose.png) |
+   | Figure 20: Efficient | ![CPU lane of the efficient backend showing four aten::transpose ops before the fused attention kernel](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/efficient-trasnpose.png) |
```
sergiopaniego, line 377: `efficient-trasnpose.png` --> `efficient-transpose.png`

torch-attention-profile.md
```diff
@@ -0,0 +1,410 @@
...
+
+- a matmul kernel (`Q . K.T`)
+- a mul kernel (the scaling)
+- an operation for the masking
+- a softmax kernel
+- a matmul kernel (`A . V`)
+
+```bash
+uv run 04_a_naive_attention.py
+uvx trace-util -f traces/ -b <hf_name>/traces
```
sergiopaniego:
```suggestion
uvx trace-util -f traces/ -b <hf_uname>/traces
```

torch-attention-profile.md
```diff
@@ -0,0 +1,410 @@
...
+    attn = torch.softmax(scores, dim=-1)
+    out = torch.matmul(attn, v)  # [batch, heads, seq, head_dim]
+    return out
+```
+
+Let's look at the trace and see if something changed.
+
+```bash
+uv run 04_b_inplace_ops_attention.py
+uvx trace-util -f traces/ -b <Hf_uname>/traces
```
sergiopaniego:
```suggestion
uvx trace-util -f traces/ -b <hf_uname>/traces
```

torch-attention-profile.md
sergiopaniego: And they have  `import os` twice

torch-attention-profile.md
```diff
@@ -0,0 +1,410 @@
...
+
+| ![Profiler trace of the SDPA flash backend](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/flash-backend.png) |
+| :--: |
+| Figure 15: The flash backend trace, one fused `pytorch_flash` kernel per forward |
+
+The `void pytorch_flash` kernel (Figure 15) is [FlashAttention-2](https://arxiv.org/abs/2307.08691) (Tri Dao's implementation), vendored into PyTorch.
+
+Before we read the trace any further, it is worth answering the question you should be asking by now: _why is there a whole backend named "flash", and why does it matter so much?_
+
+#### Why flash attention exists
```
sergiopaniego:
```suggestion
#### Why flash attention exists?
```

torch-attention-profile.md
```diff
@@ -0,0 +1,410 @@
...
+| Figure 18: The cuDNN backend trace, a single generated attention kernel per forward. |
+
+By now the pattern is familiar. Like flash and efficient, cuDNN gives us one fused, flash-style kernel per forward (Figure 18). So the natural question is: **if flash already fuses attention, why does PyTorch ship yet another flash backend?** The answer is _who writes the kernel and how it is built_, and that difference is what makes the trace look different.
+
+#### How is cuDNN kernel different
+
+Flash and efficient are **fixed, pre-compiled kernels** vendored into PyTorch. You get the same binary every time. cuDNN is NVIDIA's own deep learning library, and its attention kernel is **generated and tuned for the specific problem** at hand. It is closer in spirit to `torch.compile`'s codegen than to a fixed cuBLAS binary. You can read that straight off the (very long) kernel name:
+
+```bash
+cudnn_generated_fort_native_sdpa_sm80_flash_fprop_wmma_f16_knob_6_128x64x64_4x1x1_cga1x1x1_kernel0_0
```
sergiopaniego, line 361: 😱

sergiopaniego (approved):
great stuff! I loved the series and how you try to explain every single idea in the most simple way, without assuming any previous knowledge 👏

torch-attention-profile.md
ariG23498: Great catch! I have updated the scripts.

torch-attention-profile.md
ariG23498: Valid reaction.

_blog.yml
```diff
@@ -5524,4 +5524,13 @@
     - kernels
     - hub
     - optimization
-    - kernel-builder
\ No newline at end of file
+    - kernel-builder
+
+- local: torch-attention-profile
+  date: July 6, 2026
```
sayakpaul: To change.

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+<div style="
+  border: 1px solid #1f5c48;
+  border-radius: 6px;
+  padding: 1.5rem 2rem;
+  background: #17211f;
+  color: #ddd;
+  font-size: 1.1rem;
+  line-height: 1.7;
+">
```
sayakpaul, line 23: Cool idea!

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+    </li>
+  </ol>
+
+</div>
+
+The series "Profiling in PyTorch" is meant to make you comfortable reading profiler traces and tables. In [Part 1](https://huggingface.co/blog/torch-profiler) we profiled basic math operations like addition and multiplication. We saw how the profiler table uncovers hotspots, and how the profiler trace shows the order in which an algorithm runs over time.
+
+In [Part 2](https://huggingface.co/blog/torch-mlp-fusion) we wrapped that addition and multiplication into a torch linear layer. We then stacked several linear layers on top of each other (a multilayer perceptron) and profiled that. Along the way we also profiled fused and hand-tuned kernels.
+
+The next logical step is another fundamental algorithm, attention. Attention looks like a fairly simple operation, but a lot of clever tricks go into making it fast. Our goal here is not to cover every trick in detail. Instead, we want to see how each one looks different under the profiler.
```
sayakpaul: Only if one cares about the transformer architecture, then it becomes logical. Otherwise, one could ask "why not convolutions, despite them being an important algorithm". 

So, it could lead with something like (with some other suggestions too):


```suggestion
From the perspective of the Transformer architecture, the next logical step for us to profile is yet another fundamental algorithm, attention. While being infamous for its quadratic-time complexity, many clever tricks exist to mitigate that issue and make it fast. Our goal here is not to cover every trick in detail. Instead, we want to see how each one looks different under the profiler.
```

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+## Naive attention
+
+Attention works with Queries (`Q`), Keys (`K`), and Values (`V`). The interaction between them can be written as a short sequence of steps:
+
+1. Build the attention scores `S`: `matmul(Q, K.T)`
+2. Scale the scores: `S * scale`
+3. Apply a causal mask to the scores: `S.masked_fill(mask, "-inf")`
```
sayakpaul: Only when it's not self-attention. I think it's important to make this point clear.

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+## Naive attention
+
+Attention works with Queries (`Q`), Keys (`K`), and Values (`V`). The interaction between them can be written as a short sequence of steps:
+
+1. Build the attention scores `S`: `matmul(Q, K.T)`
```
sayakpaul: Maybe we could map these symbols to how they're present in code? So, `Q` would become `q`, `A` would become `attn`, for example.

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+```bash
+uv run 04_a_naive_attention.py
+uvx trace-util -f traces/ -b <hf_uname>/traces
+```
+
+| ![CPU lane of the naive attention profiler trace, with the `attn_fwd` block expanded to show its matmul, mul, masked_fill and softmax operations](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/cpu-profile-naive.png) |
+| :--: |
+| Figure 1: The CPU lane of the profile trace for naive attention highlighting the discrete operations |
+
```
sayakpaul, line 102: Should we briefly mention the dimensions of the inputs?

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+Attention works with Queries (`Q`), Keys (`K`), and Values (`V`). The interaction between them can be written as a short sequence of steps:
+
+1. Build the attention scores `S`: `matmul(Q, K.T)`
+2. Scale the scores: `S * scale`
+3. Apply a causal mask to the scores: `S.masked_fill(mask, "-inf")`
+4. Normalize the scores with softmax to get the attention weights `A`: `softmax(S)`
+5. Reweight the values with those weights: `matmul(A, V)`
+
+So attention is really a collection of primitive operations. Some of them we already know (the matmuls), and the rest are easy to spot. Let's write a naive attention module in PyTorch and profile it.
```
sayakpaul, line 69: Maybe we don't call it naive here and reveal that it's naive only after presenting the profiling results to keep the suspense aspect?

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+- an operation for the masking
+- a softmax kernel
+- a matmul kernel (`A . V`)
+
+```bash
+uv run 04_a_naive_attention.py
+uvx trace-util -f traces/ -b <hf_uname>/traces
+```
+
+| ![CPU lane of the naive attention profiler trace, with the `attn_fwd` block expanded to show its matmul, mul, masked_fill and softmax operations](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/cpu-profile-naive.png) |
```
sayakpaul, line 99: Ouf this is such a beautiful trace. A sight for sore eyes.

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+| :--: |
+| Figure 1: The CPU lane of the profile trace for naive attention highlighting the discrete operations |
+
+Figure 1 shows the CPU lane of the profile (the GPU lane is folded so it does not overwhelm us). Inside `attn_fwd` (our annotated forward call) we can see exactly the operations we guessed. The matmul is an old friend by now, and the new operations are easy to place:
```
sayakpaul:
```suggestion
Figure 1 shows the CPU lane of the profile (the GPU lane is folded so it does not overwhelm us). Inside `attn_fwd` (our annotated forward call) we can see exactly the operations we guessed. The matmul is an old friend by now, and the new operations are easy to spot:
```

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+Figure 3 lets us read off the individual kernels for one profiler step:
+
+1. matmul (query and key)
+2. mul (scaling)
+3. memory copy 🤔
+4. causal masking
+5. softmax (produces the attention weights)
+6. matmul (attention weights and values)
+
+Five of these are expected. The memory copy is the odd one out, so where does this come from? The clue is that PyTorch has in-place operations. When you operate on a tensor the ordinary (out-of-place) way, PyTorch often makes a copy, applies the operation to it, and returns the copy. Following the sequence of operations, the culprit here is our `masked_fill`.
```
sayakpaul:
```suggestion
Five of these are expected. The memory copy is the odd one out, so where does this come from? The clue is that PyTorch has in-place operations. When you operate on a tensor the ordinary (out-of-place) way, PyTorch often makes a copy, applies the requested operation to it, and returns the copy. Following the sequence of operations, the culprit here is our [`masked_fill`](https://docs.pytorch.org/docs/2.13/generated/torch.Tensor.masked_fill.html).
```

Perhaps we also point to the docs of `masked_fill`?

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+| Type | GPU stream |
+| :--: | :--: |
+| Figure 6: Naive masking | ![GPU kernels for naive attention including a separate Memcpy kernel before the masking](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/each-kernels-naive.png) |
+| Figure 7: In place masking | ![GPU kernels for naive attention with in-place masking, with the Memcpy kernel gone](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/each-kernels-inpace.png) |
+
+On the GPU lane the `Memcpy` kernel is gone for good (Figures 6 and 7). With a one line change we shaved a whole kernel off each forward pass. This may not look like much on its own, but remember this is a single attention operation. In a Large Language Model (LLM) it repeats once per layer and head, and there are many layers and heads, so the saving adds up quickly (and if it earns you a raise, sharing at least 10% with us feels only fair).
```
sayakpaul:
```suggestion
On the GPU lane the `Memcpy` kernel is gone for good (Figures 6 and 7). With a one line change we shaved a whole kernel off each forward pass. This may not look like much on its own, but remember this is a single attention operation. In the context of a transformer based large model (LLMs, Diffusion models, etc.), it repeats once per layer and head, and there are many layers and heads, so the saving adds up quickly (and if it earns you a raise, sharing at least 10% with us feels only fair).
```

sayakpaul: > and if it earns you a raise, sharing at least 10% with us feels only fair

This is a me exclusive phrase, though.

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+uv run 04_c_sdpa_attention.py --backend math
+uvx trace-util -f traces/ -b <hf_uname>/traces
+```
+
+Before we open anything, let's guess. We have replaced hand written attention (matmul, mul, mask, softmax, matmul) with a single one liner, so we should expect the trace to get _simpler and faster_. Fewer kernels, less CPU dispatch, maybe even a fused kernel. Let's check the profiler table first.
+
+| Metric | Naive in-place | SDPA math |
+| :--: | :--: | :--: |
+| *_fwd CUDA time avg | 1.955 ms | 7.239 ms |
+| Self CUDA time total | 7.194 ms | 27.279 ms |
```
sayakpaul: What is "Self CUDA time total"?

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+| | Profiler Trace |
+| :--: | :--: |
+| Figure 8: Profiler trace of naive in-place attention showing five GPU kernel launches for one forward | ![GPU lane of naive in-place attention with five kernel launches for one forward pass](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/inplace-kernel-launches.png) |
+| Figure 9: Profiler trace of the SDPA math backend showing 20 GPU kernel launches for a single attention forward | ![GPU lane of the SDPA math backend with twenty kernel launches for a single attention forward pass](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/math-kernel-launches.png) |
+
+Opening the trace (Figure 9) shows why the alarm bells ring, the math backend launches `20` GPU kernels per forward instead of the `5` launched with our naive attention implementation (Figure 8). This is the opposite of what we guessed. Let's figure out why this happens.
+
+#### Tensor Cores left vacant
+
+In [Part 2](https://huggingface.co/blog/torch-mlp-fusion) we learned to read a kernel name like a fingerprint. Let's use that habit here:
```
sayakpaul: Do we want to provide the link to the specific section that talks about reading kernel names (https://huggingface.co/blog/torch-mlp-fusion#where-did-the-transpose-go-kernel-layouts-and-pre-ops)?

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+This is our first surprise, the one liner is `3.7x` slower.
+
+| | Profiler Trace |
+| :--: | :--: |
+| Figure 8: Profiler trace of naive in-place attention showing five GPU kernel launches for one forward | ![GPU lane of naive in-place attention with five kernel launches for one forward pass](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/inplace-kernel-launches.png) |
+| Figure 9: Profiler trace of the SDPA math backend showing 20 GPU kernel launches for a single attention forward | ![GPU lane of the SDPA math backend with twenty kernel launches for a single attention forward pass](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/math-kernel-launches.png) |
+
+Opening the trace (Figure 9) shows why the alarm bells ring, the math backend launches `20` GPU kernels per forward instead of the `5` launched with our naive attention implementation (Figure 8). This is the opposite of what we guessed. Let's figure out why this happens.
+
+#### Tensor Cores left vacant
```
sayakpaul:
```suggestion
#### Tensor cores left vacant
```

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+In [Part 2](https://huggingface.co/blog/torch-mlp-fusion) we learned to read a kernel name like a fingerprint. Let's use that habit here:
+
+| Run | matmul kernel |
+| :--: | :--: |
+| Figure 10: Naive attention | ![Matmul kernel name for naive attention in Perfetto, carrying the s16816 bfloat16 Tensor-core GEMM signature](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/cuda-core-kernels.png) |
+| Figure 11: SDPA with math backend | ![Matmul kernel name for the SDPA math backend, carrying the sgemm FP32 CUDA-core signature](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/tensor-core-kernels.png) |
+
+The `s16816` in the naive kernel (Figure 10) is the signature of a `bfloat16` Tensor core matmul (the `16x8x16` Tensor core instruction). `sgemm` (Figure 11) is the classic single precision (`FP32`) matmul that runs on the ordinary CUDA cores. So the math backend is not touching the Tensor cores at all.
```
sayakpaul: This is a very important section. I think we should lead this by saying A100s (our GPU for deriving the profile traces) have tensor cores and they are specialized for performing accelerated matmuls (hyperlinking the "tensor cores" phrase) and are known to perform better than regular CUDA cores.

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+How does this matter? A Streaming Multiprocessor (SM) is the compute unit of a GPU, and each SM has two kinds of arithmetic units, the CUDA cores and the Tensor cores. CUDA cores are general purpose and process a handful of elements at a time. Tensor Cores are specialised hardware that multiply and accumulate a whole small matrix tile in a single instruction. So when the math backend deliberately trades speed for numerical accuracy it upcasts tensors to `FP32` (doubling the data moved) and uses the CUDA cores instead of the faster Tensor cores.
```
sayakpaul: Not all SMs have this, though. That should be made clear.

sayakpaul: >  So when the math backend deliberately trades speed for numerical accuracy it upcasts tensors to `FP32` (doubling the data moved) and uses the CUDA cores instead of the faster Tensor cores.

Even when the input is bfloat16!? Wow, TIL! Do you think this should be explicitly mentioned?

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+#### Causal masks built
+
+In the naive version we built the causal mask once and reused it. Here we passed `is_causal=True` and the math backend materialized one for us, on every single call. You can watch it happen on the CPU lane:
```
sayakpaul:
```suggestion
In the naive version we built the causal mask once and reused it. Here we passed `is_causal=True` and the math backend materialized one for us, on _every_ single call. You can watch it happen on the CPU lane:
```

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+| ![CPU lane of the SDPA math backend showing the ops that rebuild the causal mask: aten::ones, aten::tril, aten::scalar_tensor, aten::fill_ and aten::where](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/mask-math.png) |
```
sayakpaul, line 241: Do we want to contrast this to the naive attention mask building?

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+Here is what we see in Figure 12
+
+```bash
+aten::ones -> aten::tril            build a [seq, seq] lower-triangular matrix
+aten::scalar_tensor -> aten::fill_  make the -inf fill value
+aten::where                         turn it into an additive bias (0 or -inf)
+```
+
+On the GPU this shows up as a `triu_tril_kernel`, several `where` kernels, and an `add_`. The convenience flag that let us stop thinking about the mask did not remove the work, it just moved it one layer down, where the mask is rebuilt from scratch every forward.
```
sayakpaul, line 253: 🥲

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+```bash
+uv run 04_c_sdpa_attention.py --backend efficient
+uvx trace-util -f traces -b <hf_uname>/traces
+```
+
+| ![Profiler trace of the SDPA efficient backend showing a single fused fmha_cutlassF attention kernel per forward](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/efficient-backend.png) |
+| :--: |
+| Figure 14: The profiler trace for sdpa with efficient backend |
+
```
sayakpaul, line 281: We should also show the table here like we did for "math"?

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+The `void pytorch_flash` kernel (Figure 15) is [FlashAttention-2](https://arxiv.org/abs/2307.08691) (Tri Dao's implementation), vendored into PyTorch.
+
+Before we read the trace any further, it is worth answering the question you should be asking by now: _why is there a whole backend named "flash", and why does it matter so much?_
+
+#### Why flash attention exists?
+
+Let's go back to the math backend for a moment. Its real problem was not the count of 20 kernels, it was what those kernels handed to each other.
+
+Step 1 builds the full score matrix `S = Q . K.T`, which is `[seq, seq]` **per head**. For a sequence length of 4096 that is roughly 16 million numbers for a single head. That matrix is written out to the HBM (the GPU's main memory), read back to be scaled, written again for the mask, read again for the softmax, and so on. Attention's cost is dominated by this **back and forth traffic to HBM**, not by the matmuls themselves.
```
sayakpaul: Could make sense to do the math in brackets (for 16 million).

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+FlashAttention attacks exactly this. Instead of computing the whole `S` matrix and only then reducing it, it walks over `K` and `V` in **tiles**, keeps a running softmax as it goes (the "online softmax" trick), and accumulates the output one tile at a time. The full `[seq, seq]` score matrix is **never written to HBM**, it only ever lives on-chip. This is the single idea that lets the entire attention pipeline collapse into one fused kernel that stays in bf16 on the Tensor cores.
+
+#### Why flash looks "wrong" under the profiler
+
+| ![Perfetto footprint of the flash kernel reporting an estimated achieved occupancy of 13%](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/flash-occupancy.png) |
```
sayakpaul, line 320: This feels a bit unaccompanied to me without the context on what memory occupancy looks like for the other flavors discussed in this post.

torch-attention-profile.md
sayakpaul: Smol name

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+| ![Profiler trace of the SDPA cuDNN backend showing a single cudnn_generated attention kernel per forward](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/cudnn-backend.png) |
+| :--: |
+| Figure 18: The cuDNN backend trace, a single generated attention kernel per forward. |
+
+By now the pattern is familiar. Like flash and efficient, cuDNN gives us one fused, flash-style kernel per forward (Figure 18). So the natural question is: **if flash already fuses attention, why does PyTorch ship yet another flash backend?** The answer is _who writes the kernel and how it is built_, and that difference is what makes the trace look different.
+
+#### How is cuDNN kernel different
+
+Flash and efficient are **fixed, pre-compiled kernels** vendored into PyTorch. You get the same binary every time. cuDNN is NVIDIA's own deep learning library, and its attention kernel is **generated and tuned for the specific problem** at hand. It is closer in spirit to `torch.compile`'s codegen than to a fixed cuBLAS binary. You can read that straight off the (very long) kernel name:
```
sayakpaul, line 358: That means for ephemeral environments, there might be a factor of cold starts?

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+| Variant | What we changed | Kernels / forward | What the trace revealed |
+| :-- | :-- | :--: | :-- |
+| Naive attention | Attention built by hand from primitives (matmul, mul, mask, softmax, matmul) | 6 | A hidden `Memcpy` from the out-of-place `masked_fill`. |
+| Naive in-place | `masked_fill` → `masked_fill_` | 5 | One line drops the `Memcpy` kernel entirely. |
+| SDPA math | `F.scaled_dot_product_attention` pinned to the math backend | 20 | The reference: FP32 on CUDA cores, mask rebuilt every call, `_safe_softmax`. Correct but ~3.7x slower. |
+| SDPA efficient | Efficient (xformers) backend | 1 | One fused `fmha_cutlassF` kernel, stays in bf16 on Tensor cores. |
+| SDPA flash | Flash backend | 1 | One fused `pytorch_flash` kernel (FlashAttention-2). Fastest, despite "wrong-looking" 13% occupancy. |
+| SDPA cuDNN | cuDNN backend | 1 | A per-problem generated kernel: no transposes, `cuLaunchKernelEx`, but the cost moved to a fat CPU bar. |
+
+## Concluding the series
```
sayakpaul, line 421: Or not yet.

torch-attention-profile.md
ariG23498: I don't agree to the provided suggestion. A large model can be of any kind (without attention), but a LLM should almost always have the attention algorithm in it.

sayakpaul: Then it should say large models like language models and diffusion models. Just saying LLMs sounds exclusive when it doesn't have to be.

torch-attention-profile.md
ariG23498: I don't quite follow. How does causal masking depend of self or cross attention?

torch-attention-profile.md
ariG23498: I wanted to tag a name to this variant upfront, so that I could refer back to this quickly.

torch-attention-profile.md
ariG23498: I don't think that would be necessary, as we do not show the GPU trace or talk about the grid, block, or tile which might directly depend on the input shape.

torch-attention-profile.md
sayakpaul: My bad. I meant to suggest that attention doesn't always have to have masks and there could be a note about that.

torch-attention-profile.md
ariG23498: This is from the profiler table. I see your point here, this is not evident at the first glance. I will add some details here.

torch-attention-profile.md
ariG23498: Great idea! I will rephrase this section in the latest commit.

torch-attention-profile.md
ariG23498: > Even when the input is bfloat16!?

Yes! That is what I figured out, as all my inputs are explicitly in `bf16`.

> Do you think this should be explicitly mentioned?

Will add a line to make this point more visible.

torch-attention-profile.md
ariG23498: In the naive attention pipeline, the mask is passed to the module, so it does not showcase the mask building process in the profile, but only showcases the `masked_fill`. This makes me believe that comparing the two might lead to confusion.

torch-attention-profile.md
ariG23498: I don't follow the "cold starts" point.

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+| Type | GPU stream |
+| :--: | :--: |
+| Figure 6: Naive masking | ![GPU kernels for naive attention including a separate Memcpy kernel before the masking](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/each-kernels-naive.png) |
+| Figure 7: In place masking | ![GPU kernels for naive attention with in-place masking, with the Memcpy kernel gone](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/each-kernels-inpace.png) |
+
+On the GPU lane the `Memcpy` kernel is gone for good (Figures 6 and 7). With a one line change we shaved a whole kernel off each forward pass. This may not look like much on its own, but remember this is a single attention operation. In a Large Language Model (LLM) it repeats once per layer and head, and there are many layers and heads, so the saving adds up quickly (and if it earns you a raise, sharing at least 10% with us feels only fair).
```
remi-or:
```suggestion
On the GPU lane the `Memcpy` kernel is gone for good (Figures 6 and 7). With a one line change we shaved a whole kernel off each forward pass. This may not look like much on its own, but remember this is a single attention operation. In a Large Language Model (LLM) it repeats once per layer, and there are many layers, so the saving adds up quickly (and if it earns you a raise, sharing at least 10% with us feels only fair).
```
Attention does happen per-head but the operations we are talking about only happen once per layer (since heads are folded inside the tensor)

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+> [!NOTE]
+> Out-of-place is PyTorch's default for a reason. To compute gradients, autograd has to remember the tensor values it saw on the forward pass, because many backward formulas reuse them. An in-place operation overwrites those values in memory, so the backward pass would read the wrong numbers. Due to the fact that we run `forward` under `torch.no_grad`, in-place is safe for us, with no backward pass and nothing to corrupt.
```
remi-or: Maybe balance that statement with the fact that in-place, when possible, saves time and memory? I know we already hinted at the former, but the latter is also really important for large tensors like logits!

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+The `void pytorch_flash` kernel (Figure 15) is [FlashAttention-2](https://arxiv.org/abs/2307.08691) (Tri Dao's implementation), vendored into PyTorch.
+
+Before we read the trace any further, it is worth answering the question you should be asking by now: _why is there a whole backend named "flash", and why does it matter so much?_
+
+#### Why flash attention exists?
+
+Let's go back to the math backend for a moment. Its real problem was not the count of 20 kernels, it was what those kernels handed to each other.
+
+Step 1 builds the full score matrix `S = Q . K.T`, which is `[seq, seq]` **per head**. For a sequence length of 4096 that is roughly 16 million numbers for a single head. That matrix is written out to the HBM (the GPU's main memory), read back to be scaled, written again for the mask, read again for the softmax, and so on. Attention's cost is dominated by this **back and forth traffic to HBM**, not by the matmuls themselves.
```
remi-or: Maybe add that there may not be enough memory to even materialize the matrix ? 
```suggestion
Step 1 builds the full score matrix `S = Q . K.T`, which is `[seq, seq]` **per head**. For a sequence length of 4096 that is roughly 16 million numbers for a single head. That matrix is written out to the HBM (the GPU's main memory), if there is even enough space to do so. Then, it is read back to be scaled, written again for the mask, read again for the softmax, and so on. Attention's cost is dominated by this **back and forth traffic to HBM**, not by the matmuls themselves.

File: torch-attention-profile.md [hunk 34]
```diff
@@ -0,0 +1,429 @@
...
+
+#### Why flash looks "wrong" under the profiler
+
+| ![Perfetto footprint of the flash kernel reporting an estimated achieved occupancy of 13%](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/torch-attention-profile/flash-occupancy.png) |
+| :--: |
+| Figure 16: Estimated occupancy of flash kernel is seen to be 13% |
+
+Here is where flash surprises people who read profiler footprints. It is the fastest backend, yet the profiler reports it with very **low occupancy** (shown in Figure 16). To see why that is fine, we need three quick definitions.
+
+A CUDA kernel is launched as a **grid** of **blocks** (of threads). Each block contains many **threads**, and the hardware executes those threads in groups of 32 called **warps**.
```
remi-or: I think we need to explain what a thread is before what a block is. Just a one line statement would be enough, something like: 

A GPU kernel is essentially a series of instructions executed by many small execution units. Execution units are called `threads`: they take care of loading variables, adding them together, storing them back, etc. For each kernel, we launch many, many threads, and to keep track of them, we group them by **blocks**. 

Or something along those lines (not particularly fan of how the paragraph above looks, but it conveys the right idea at least)

torch-attention-profile.md
```diff
@@ -0,0 +1,429 @@
...
+
+Blocks are scheduled onto Streaming Multiprocessors (SMs), the main compute units of a GPU. A block lives entirely on one SM, and an SM can host multiple blocks at once _if it has enough resources_. Those resources include registers, shared memory, maximum resident threads, and maximum resident warps. So when we say a kernel has low **occupancy**, we mean each SM has fewer resident warps than it could theoretically support.
```
remi-or, line 328: Did we define `shared memory`?

remi-or, line 328: Also, shameless plug, but we can link to this blog post for more details on threads, warps, etc. for more details (feel free to ignore ofc: https://huggingface.co/blog/mi300kernels#a-quick-introduction-to-the-mi300x)

torch-attention-profile.md
sayakpaul: > its attention kernel is **generated and tuned for the specific problem** at hand. It is closer in spirit to `torch.compile`'s codegen than to a fixed cuBLAS binary.

Cold start means that some thing won't deliver what it is expected to deliver (speedup in this case) right off the bat. It will run some additional stuff (tuning in this case) when it's "started".

remi-or (approved):
Very cool ! Just a few comments on my part, imo this is a really good post inside a very relevant part of torch

torch-attention-profile.md
ariG23498: Great catch!

torch-attention-profile.md
ariG23498: We did not. Do you think there is a need to define it, rather than just mentioning it here?

remi-or: We can probably just get away with "memory" so that the reader understands there is a finite amount of it without asking themselves what is "shared memory". But I agree it's self explanatory, so I think both are fine in the end.

PR #3465, opened 2026-07-15
Security incident disclosure — July 2026

XciD:
Public disclosure of the July 2026 security incident (AI-driven intrusion into part of our production infrastructure).

- **Post:** `security-incident-2026.md` → `hf.co/blog/security-incident-2026`
- **Author:** `huggingface` org account (mirrors the previous breach disclosure, `space-secrets-disclosure`)
- **Tags:** `security`

## Preparing the Article

- [x] Add an entry to `_blog.yml`.
- [x] Add a thumbnail. **Not done yet** — `thumbnail.png` still needs to be uploaded to `assets/security-incident-2026/`.
- [x] Check you use a short title and blog path.
- [x] Upload any additional assets to the Documentation Images repo. N/A — no inline images in the post.
- [x] Add metadata (authors) to the `md` file.
- [x] Ensure the publication date is correct (July 15, 2026).
- [ ] Preview the content.

## Getting a Review

Requesting review from @julien-c.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_013CSGC4dWN8FcAPbhKkWYjE)_

PR #3469, opened 2026-07-16
refactor: keep disclosure and analysis in one post, reframed

XciD:
Proposal for the discussion in #3465, incorporating the feedback from the Slack thread. Keeps everything in a single post, no content removed.

1. **Section order**: client-facing content first. "For our community" (token rotation, contact) moves up right after "What we did". The technical analysis ("Analyzing an AI-driven intrusion") and "What this means" close the post. Customers get impact and actions without reading through the analysis; technical readers reach it naturally.
2. **"The asymmetry problem" conclusion reframed as a lesson for defenders**: the closing paragraph no longer makes an industry-level claim ("treat capable open models as part of the security equation"). It states the facts (we don't know which model the attacker used, whether a jailbroken hosted model or an unrestricted open-weight one, and the asymmetry holds either way), turns the takeaway into actionable incident-response preparedness advice, and mentions we are sharing the feedback with the providers concerned.

### DIFF
--- a/security-incident-july-2026.md
+++ b/security-incident-july-2026.md
@@ -37,16 +37,10 @@ We are grateful to the teams across Hugging Face who responded around the clock,
 
 The attack was initially surfaced through AI-assisted detection. Our anomaly-detection pipeline uses LLM-based triage over security telemetry to separate real signals from the daily noise, and it was the correlation of those signals that flagged the compromise.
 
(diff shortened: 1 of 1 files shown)


pcuenca (approved):
I think the neutral objective tone makes it work quite well!

PR #3483, opened 2026-07-28
docs: clarify the intrusion duration in the agent intrusion timeline

XciD:
The post gave two durations three lines apart: "the 4.5-day intrusion" in the replay link, then "over roughly two and a half days" opening the TL;DR. Both are right, they just measure different things: 4.5 days of campaign end to end, of which about 2.5 inside our infrastructure.

Two words fix it:

- replay link: `4.5-day intrusion` becomes `4.5-day campaign`
- TL;DR: `two and a half days` becomes `two and a half days inside our infrastructure`

The timeline figure caption already says "campaign", so it stays consistent.


julien-c:
lgtm

PR #3486, opened 2026-07-29
Clarify ExploitGym/CyberGym's role in the OpenAI-Huggingface incident

wzunknown:
We are CyberGym/ExploitGym team. We have noticed several points concerning CyberGym and ExploitGym in the blog (https://huggingface.co/blog/agent-intrusion-technical-timeline) that we believe should be clarified to avoid mischaracterizing our projects and our involvement in the incident.

We would like to clarify that, although the incident involved an external instance named “CyberGym,” the vulnerable endpoint was not part of our original CyberGym harness and did not reflect our CyberGym’s original design or implementation. As confirmed with Adrien Carreira, our CyberGym does not provide an endpoint that allows agents to submit arbitrary code for execution or run arbitrary commands in this manner. This external instance named “CyberGym” was provided by some unknown third-party unrelated to the original CyberGym team.

The incident occurred during an internal OpenAI evaluation based on the ExploitGym benchmark. OpenAI used its own internal implementation and evaluation infrastructure, rather than directly using ExploitGym's open-source reference framework (https://github.com/sunblaze-ucb/exploitgym), and we were not involved in designing, operating, or monitoring that internal evaluation.

PR #3487, opened 2026-07-29
docs: correct the attribution of the third-party code-execution harness

XciD:
Minimal alternative to #3486, which reports a real error but rewrites more of the post than the correction requires. Credit to the CyberGym/ExploitGym maintainers for the report and for the follow-up suggestions in the review, which are taken here.

The error: the post attributed the abused harness to their project ("the one used in the SunBlaze-UCB/exploitgym benchmark"). The instance was deployed and operated by an unknown third party, not by the maintainers, and the arbitrary-code endpoint the agent abused is not something upstream CyberGym provides.

What this PR changes:

- TL;DR: "an OpenAI cyber-capability evaluation harness called ExploitGym" becomes "an internal OpenAI cyber-capability evaluation based on the ExploitGym benchmark". The old phrasing implied ExploitGym was an OpenAI harness, when it is the maintainers' benchmark and OpenAI ran an internal evaluation based on it. Added, per the review: OpenAI ran this on its own infrastructure, and the ExploitGym maintainers and their infrastructure had no involvement in the deployment or operation of that evaluation environment.
- TL;DR: the five datasets are described by what we can assert from access logs, their names and files suggesting a connection to ExploitGym/CyberGym challenges and solutions, rather than asserting their contents. The impact bound (five datasets, nothing else) is unchanged.
- Stage 1: the harness is described as an instance labeled "CyberGym", deployed by an unknown third party, exposing an arbitrary-code endpoint upstream does not provide, instead of being attributed to the SunBlaze benchmark. The rest of the paragraph, including the `/submit-function` mechanics, is untouched.
- `attack-chain-dark.svg`: the sandbox box reads "third-party cyber-eval harness" instead of "CyberGym harness", per the review, so the figure does not imply an official or upstream instance.

One suggestion from the review not taken: "using an internal implementation rather than ExploitGym's open-source reference framework". That is a claim about OpenAI's internals that we cannot verify and that their July 28 update does not make. The alternative wording proposed in the same comment is what is in the PR.

The replay Space needs no change, it mentions neither project.


pcuenca (approved):

julien-c (approved):

julien-c (approved):

thomwolf (approved):

wzunknown:
Thanks for the new PR, @XciD. We still believe it would be helpful to clarify CyberGym and ExploitGym’s roles in this incident.

For the current revision, we suggest adding the following clarifications:

> The agent was running an internal OpenAI cyber-capability evaluation based on the ExploitGym benchmark, which tasks an AI agent with finding and exploiting software vulnerabilities. ***OpenAI ran this on its own infrastructure using an internal implementation rather than ExploitGym's open-source reference framework, and the ExploitGym maintainers had no involvement in it.***  As far as we were able to infer, across the course of being evaluated on this benchmark, the agent inferred that Hugging Face may host that benchmark's models, datasets, and reference solutions. We believe the entire intrusion was, from the agent's point of view, an attempt to cheat the evaluation: reach our production systems and steal the test solutions rather than solve the challenge on its own.

If you think it is a claim about OpenAI's internals that you cannot verify, do you think it is reasonable to add "***OpenAI ran this on its own infrastructure, and the ExploitGym maintainers had no involvement in deployment or operation of its evaluation environment.***"

> On this external sandbox, the agent abused an existing code-execution harness. ***The instance was labeled "CyberGym" deployed by an unknown third party rather than by the [CyberGym](https://www.cybergym.io/) maintainers, and exposing an arbitrary-code endpoint that upstream CyberGym does not provide.*** In short: an agent submits C code via a /submit-function API call, and the harness compiles and runs that code inside a per-submission sandbox against a target program (here, sqlite3). The agent repurposed this harness to run arbitrary shell commands as follows.

We also suggest changing the figure label from “third-party CyberGym deployment” to ***“third-party cyber-eval harness”*** to avoid implying that the deployment was an official or upstream CyberGym instance. Really appreciate it!

PR #3496, opened 2026-08-10
fix outro and follow-ups

merveenoyan:


alvarobartt (approved):

ariG23498 (approved):
The collections are still empty. Do we want to add them here?

PR #3500, opened 2026-08-10
[Muse Glimmer] Revised code snippets

ariG23498:
CC: @merveenoyan @pcuenca


muse-glimmer.md
```diff
@@ -161,18 +189,83 @@ inputs = processor.apply_chat_template(
...
-#### Multimodal tool calling
+### Video Inference
+
+To work with videos we recommend installing `torchcodec` into the environment.
+
+```bash
+pip install torchcodec
+```
+
+Muse Glimmer can answer complex questions about videos *without audio*. You can do video inference as follows, here’s an example from VideoMME2, which is the most popular video question answering benchmark.
```
pcuenca, line 205: I assume that audio is ignored, not that it fails if input videos contain audio, is that right?

pcuenca (approved):

muse-glimmer.md
ariG23498: Yes, the audio is just ignored.

pcuenca: Then we could say so in the text.

PR #3504, opened 2026-08-14
Use SVG charts in State of Open Models post

mishig25:
## Summary

- replace the thirteen body-chart PNG references with SVGs
- keep the thumbnail as PNG because blog frontmatter does not accept SVG thumbnails
- host the SVGs in the documentation-images dataset

## Validation

- all thirteen SVG URLs return HTTP 200 with SVG content
- the uploaded files exactly match the verified local SVGs
- blog entry and YAML/frontmatter validation passed
- whitespace checks passed


mishig25:
not needed

PR #3522, opened 2026-09-01
Introducing @huggingface/kernels: 200+ WebGPU Kernels for Local AI

nico-martin:
Adds the new post about @huggingface/kernels and Fleet


xenova (approved):
✅

PR #3527, opened 2026-09-09
asyncgrpo lora running on hfjobs

AmineDiro:
AsyncGRPO with LoRA across Hugging Face Jobs.

Trainer and vLLM replicas run as separate Jobs and exchange adapters through a mounted Storage 🪣 

Media is in huggingface/documentation-images under blog/asyncgrpo-lora-hfjobs.


assets/asyncgrpo-lora-hfjobs/thumbnail.png
burtenshaw: <3

_blog.yml
```diff
@@ -5749,3 +5749,14 @@
...
+- local: asyncgrpo-lora-hfjobs
+  date: September 9, 2026
+  tags:
+    - llm
+    - rl
+    - trl
+    - lora
+    - open-source
+    - hub
+    - storage
```
burtenshaw:
```suggestion
    - storage
    - jobs
    - compute
    - training
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
+---
+title: "Async GRPO with LoRA across Hugging Face Jobs: a bucket, a proxy, and no NCCL"
+thumbnail: /blog/assets/asyncgrpo-lora-hfjobs/thumbnail.png
+authors:
+  - user: aminediroHF
+  - user: qgallouedec
+---
+
+# Async GRPO with LoRA across Hugging Face Jobs: a bucket, a proxy, and no NCCL
+
```
burtenshaw: I would add a tl;dr that says 'the thing that we did'

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+LoRA support recently landed in TRL's `AsyncGRPOTrainer` with [PR #7017](https://github.com/huggingface/trl/pull/7017). The asynchronous trainer can now train an adapter instead of the full model, and it syncs only the LoRA adapter to vLLM. This post is about a real-world project built on top of it, once training and inference no longer share a machine.
```
kashif:
```suggestion
LoRA support recently landed in TRL's `AsyncGRPOTrainer` with [PR #7017](https://github.com/huggingface/trl/pull/7017). The asynchronous trainer can now train an adapter instead of the full model, and it syncs only the LoRA adapter to vLLM. This post covers a real-world project built on top of it, where training and inference no longer share a machine.
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+
+## The architecture: leveraging Hugging Face Jobs and Storage Buckets 🪣
+
+[TRL PR #7017](https://github.com/huggingface/trl/pull/7017) adds an adapter-only sync path to `AsyncGRPOTrainer`. The trainer does not send tensors to vLLM. Every few optimizer steps, it saves the adapter under `<output_dir>/.vllm_lora/trl-policy-v{N}`, publishes the directory with an atomic rename, then sends its path to vLLM's `/v1/load_lora_adapter` endpoint. vLLM loads the files from disk, so the rollout worker can then request `model="trl-policy-v{N}"`.
+
+This is how runtime adapter loading already works in vLLM. The endpoint takes a path, not tensors, so the trainer and the server are expected to share a filesystem. On a Slurm cluster, that is the network filesystem. On Jobs, we get the same thing by mounting a [Storage Bucket](https://huggingface.co/docs/hub/storage-buckets) as a volume at the same path in every Job, like we mentioned earlier. Under the hood, it uses [`hf-mount`](https://github.com/huggingface/hf-mount), which exposes the bucket as a POSIX filesystem inside the container:
```
kashif:
```suggestion
This is how runtime adapter loading already works in vLLM. The endpoint takes a path, not tensors, so the trainer and the server are expected to share a filesystem. On a Slurm cluster, that is the network filesystem. On Jobs, we get the same thing by mounting a [Storage Bucket](https://huggingface.co/docs/hub/storage-buckets) as a volume at the same path in every Job, as we mentioned earlier. Under the hood, it uses [`hf-mount`](https://github.com/huggingface/hf-mount), which exposes the bucket as a POSIX filesystem inside the container:
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+
+### The dataset choice: the sanity set
+
+We chose [`sail/Sanity-Test-R1D-1.5B`](https://huggingface.co/datasets/sail/Sanity-Test-R1D-1.5B), the dataset from [Defeating the Training-Inference Mismatch via FP16](https://arxiv.org/pdf/2510.26788) (Qi et al., 2025). The reproduction code is in [`sail-sg/Precision-RL`](https://github.com/sail-sg/Precision-RL).
+
+The authors generated 40 answers for each MATH problem with DeepSeek-R1-Distill-Qwen-1.5B. They kept the problems with a success rate between 20 % and 80 %, which gives 1,460 questions. This dataset is really good for RL validation as the questions are neither already solved nor completely hopeless for that model, which means a model can get a good signal early on to train on and improve.
```
kashif:
```suggestion
The authors generated 40 answers for each MATH problem with DeepSeek-R1-Distill-Qwen-1.5B. They kept problems with a success rate between 20% and 80%, yielding 1,460 questions. This dataset is really good for RL validation because the questions are neither already solved nor completely hopeless for that model, meaning the model can get a good early signal to train on and improve.
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+- The decode phase then produces one token at a time, and each new token attends to the keys and values of all the tokens before it.
+
+Those keys and values are the KV cache. Because attention is causal, the KV of a token depends only on the tokens before it, not on what comes after. Two requests that share a prefix therefore share the KV of that prefix, and a replica that already has it in cache can **skip that part of the prefill entirely**. The whole game now is to find that replica, so a request can benefit from landing on the replica that has already seen its prefix.
+
+vLLM stores its [prefix KV cache](https://github.com/vllm-project/vllm/blob/main/vllm/v1/core/kv_cache_utils.py) in blocks of 16 tokens. Because of GRPO, the rollout worker sends `G` requests with the same prompt (in our case `G=8`). If they all reach the same replica, the first request computes the prefill and the next seven reuse it. With round-robin routing, half of them would go to a replica that does not have the prefix cached and those four requests would redo the prefill work and waste valuable GPU compute.
```
kashif:
```suggestion
vLLM stores its [prefix KV cache](https://github.com/vllm-project/vllm/blob/main/vllm/v1/core/kv_cache_utils.py) in blocks of 16 tokens. Because of GRPO, the rollout worker sends `G` requests with the same prompt (in our case `G=8`). If they all reach the same replica, the first request computes the prefill and the next seven reuse it. With round-robin routing, half would go to a replica that doesn't have the prefix cached, and those four requests would redo the prefill work and waste valuable GPU compute.
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
+---
+title: "Async GRPO with LoRA across Hugging Face Jobs: a bucket, a proxy, and no NCCL"
```
burtenshaw: What if you use the term "Ultra-scale RL". Maybe something like "Ultrascale RL 4 All: We built async GRPO and LoRA on Hugging Face jobs ". The current title could be the tldr.

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+    -- vllm/vllm-openai:v0.27.1 \
+    vllm serve Qwen/Qwen2.5-Math-1.5B --host 0.0.0.0 --port 8000 \
+        --max-model-len 4096 --logprobs-mode processed_logprobs --generation-config vllm \
+        --enable-lora --max-lora-rank 1 --max-loras 6      # max_staleness=4 -> 4+2 adapter slots
+done
+```
+
+There is another possible design where the trainer keeps only the latest adapter and always publishes it under the same name. We did not go that way, because vLLM keys its prefix cache by adapter name. With a single name, KV blocks computed under the previous weights would still match after the swap, so the prefill would not be redone and a rollout could get its prefix from one policy version and its decode from the next. The trainer would have no way to tell, and it would show up as `ratio` drifting away from 1. Versioned names make this impossible: a name always means one set of weights, and a cached prefix can never match a newer version.
+
+### The dataset choice: the sanity set
```
kashif:
```suggestion
### The dataset choice: the Sanity set
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+
+<figure class="image text-center">
+  <video controls autoplay loop muted playsinline style="max-width: 100%; margin: auto;">
+    <source src="https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/asyncgrpo-lora-hfjobs/kv-prefix-router.mp4" type="video/mp4">
+    Your browser does not support the video tag.
+  </video>
+  <figcaption style="font-size: 12px; color: #6b7280; margin-top: 4px;">The routing decision for two prompts and four requests on two replicas: 16-token blocks, chained hashes, the common prefix, one affinity hit and one spill.</figcaption>
+</figure>
+
+The video plays through the whole decision process of choosing a replica. The steps below go through a real 135-token completion request example (from the sanity dataset problems):
```
kashif:
```suggestion
The video walks through the entire decision process for choosing a replica. The steps below go through a real 135-token completion request example (from the Sanity dataset problems):
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+- TRL [PR #7017](https://github.com/huggingface/trl/pull/7017): PEFT/LoRA support for `AsyncGRPOTrainer` with adapter-only vLLM sync.
+- Hugging Face [Jobs](https://huggingface.co/docs/huggingface_hub/guides/jobs) and [Storage Buckets](https://huggingface.co/docs/hub/storage-buckets); [`hf-mount`](https://github.com/huggingface/hf-mount).
+- [`hf-mount-repro`](https://github.com/AmineDiro/hf-mount-repro): the two-script reproduction of the 30-second negative-cache stall.
+- The [trackio dashboard](https://huggingface.co/spaces/aminediroHF/async-grpo-lora-buckets) for every run in this post.
+- Penghui Qi, Zichen Liu, Xiangxin Zhou, Tianyu Pang, Chao Du, Wee Sun Lee, Min Lin, [Defeating the Training-Inference Mismatch via FP16](https://arxiv.org/pdf/2510.26788), arXiv:2510.26788, 2025. Source of the sanity dataset [`sail/Sanity-Test-R1D-1.5B`](https://huggingface.co/datasets/sail/Sanity-Test-R1D-1.5B) and of the LoRA recipe, [`sail-sg/Precision-RL`](https://github.com/sail-sg/Precision-RL), `oat/scripts/lora/bf16_grpo_tis_lora.sh`.
```
kashif:
```suggestion
- Penghui Qi, Zichen Liu, Xiangxin Zhou, Tianyu Pang, Chao Du, Wee Sun Lee, Min Lin, [Defeating the Training-Inference Mismatch via FP16](https://arxiv.org/pdf/2510.26788), arXiv:2510.26788, 2025. Source of the Sanity dataset [`sail/Sanity-Test-R1D-1.5B`](https://huggingface.co/datasets/sail/Sanity-Test-R1D-1.5B) and of the LoRA recipe, [`sail-sg/Precision-RL`](https://github.com/sail-sg/Precision-RL), `oat/scripts/lora/bf16_grpo_tis_lora.sh`.
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+
+**3. Compare two prompts.** Problem 1 has 103 tokens. Both prompts start with the same 23-token chat template. Their first block is identical, but block 2 already contains the problem text. The hashes differ from there.
+
+**4. Store the owners.** For every hash, the router remembers which replicas served it and which hashes came after it (we cap the successor set at two because we only need to know whether a block has one continuation or several). After a few prompts, the template block `h1` is owned by both replicas and already has several successors, `h2` to `h8` are owned by A only and each has a single successor, and `h2'` to `h6'` are owned by B only.
+
+In practice, every prompt in a run starts with the same tokens. Here it is the chat template and the system prompt, which amount to the first 23 tokens of all 1,460 problems. In an agent setting it would be the tool descriptions, and in a multi-turn environment it would be the shared conversation history. These blocks are in every replica's cache within seconds, so matching on them tells us nothing about where a particular prompt lives.
```
kashif:
```suggestion
In practice, every prompt in a run starts with the same tokens. Here, it is the chat template and the system prompt, which amount to the first 23 tokens of all 1,460 problems. In an agent setting, it would be the tool descriptions, and in a multi-turn environment it would be the shared conversation history. These blocks are in every replica's cache within seconds, so matching on them tells us nothing about where a particular prompt lives.
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+
+A block is *common* if every replica has served it, or if it has more than one successor. Common blocks are ignored during routing because they do not identify a particular prompt. More on this below!
+
+**5. Pick a replica.** The router counts how many leading blocks match on each replica and removes the common prefix. What is left is the number of blocks specific to this prompt. Then:
+
+- If one replica has specific blocks and it is not swamped, meaning it is at most 8 requests ahead of the least-loaded replica, the request goes there. We call this an **affinity** hit.
+- If a replica has specific blocks but it is more than 8 requests ahead, we give up on the cache and send the request to the least-loaded replica. We call this a **spill**.
+- If no replica has specific blocks, this is a new prompt. It goes to the least-loaded replica, round-robin on ties. We call this **unmatched**.
+
+Here is the rule applied to four requests. Start from a state where replicas A and B both have 3 requests in flight, and only the template block `h1` is known, on both replicas.
```
kashif:
```suggestion
Here is the rule applied to four requests. Start from a state where replicas A and B both have 3 requests in flight, and only the template block `h1` is known on both replicas.
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+| whole sync | 30.8 s | **8.5 s** (min 6.6, max 9.2) |
+| of which: pause both replicas | 0.3 s | 0.3 s |
+| adapter all-gather and save to the bucket | 0.6 s | 1.1 s |
+| both replicas accept the adapter | ~29 s | ~7 s |
+
+All 252 adapter loads succeeded 🎉: 126 syncs times 2 replicas. Six succeeded on the second attempt and 246 on the third.
+
+### Routing
+
+At the end of the run, after 64 728 rollouts, the proxy's counters read:
```
kashif:
```suggestion
At the end of the run, after 64,728 rollouts, the proxy's counters read:
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+
+```
+routed [31928, 32800]  affinity 54712  spilled 820  unmatched 9196
+```
+
+With 8 rollouts per prompt, at least one request out of eight must be cold. The theoretical minimum is therefore 12.5 %. The router gets 14.2 % unmatched requests, 84.5 % affinity hits and 1.3 % spills. There is not much left to gain here unless we start looking at each replica's load using deeper inference-side metrics based on real measured load.
```
kashif:
```suggestion
With 8 rollouts per prompt, at least one of the eight requests must be cold. The theoretical minimum is therefore 12.5 %. The router gets 14.2 % unmatched requests, 84.5 % affinity hits, and 1.3 % spills. There is not much left to gain here unless we start looking at each replica's load using deeper inference-side metrics based on real measured load.
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+
+| per optimizer step, p50 | |
+|---|---|
+| step | 22.9 s |
+| forward + backward | 21.9 s |
+| waiting for rollouts | 0.02 s |
+| rollout queue occupancy | 476 of 512 |
+| trainer MFU | 3.9 % |
+
+The rollout queue stays full and the worker is mostly blocked by backpressure. The second replica is basically useless in this configuration. We'll see later in the post how we went through runs that move the bottleneck between training and generation to make the run 3.9× faster.
```
kashif:
```suggestion
The rollout queue stays full, and the worker is mostly blocked by backpressure. The second replica is basically useless in this configuration. We'll see later in the post how we went through runs that moved the bottleneck between training and generation to make the run 3.9× faster.
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,386 @@
...
+
+`perf/fwd_s` is 1.34 s while `perf/fwd_bwd_s` is 5.6 s. A normal backward costs roughly twice the forward, and with frozen base weights it should be closer to once. A ratio of 3.2 is suspicious.
+
+The reason is that `AsyncGRPOConfig` defaults to `gradient_checkpointing=True`. Every microbatch recomputes its forward during the backward. This also explains why a 16k-token row only uses 25 GB on a 141 GB H200! It's a memory optimization for this trainer, but we don't need it in this specific case: the model is small enough to fit into VRAM with the activations kept for the backward.
+
+![Figure 3](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/asyncgrpo-lora-hfjobs/fig3-tb16k-vs-nockpt-crossover.png)
+
+*Figure 3. trackio runs `r1-dp2-tb16k` and `r1-dp2-tb16k-nockpt` overlaid over their first 134 steps. Panels: `perf/fwd_s`, `perf/fwd_bwd_s`, `perf/weight_sync_s`, `sample/rollout_queue_size`, `perf/rollout_wait_s`, `perf/mfu_fwd_bwd`. Forward+backward drops by one forward; the queue falls from ~420 to ~60 and rollout wait rises from 0.02 s to 0.5 s: the bottleneck crosses to generation.*
```
kashif:
```suggestion
*Figure 3. trackio runs `r1-dp2-tb16k` and `r1-dp2-tb16k-nockpt` overlaid over their first 134 steps. Panels: `perf/fwd_s`, `perf/fwd_bwd_s`, `perf/weight_sync_s`, `sample/rollout_queue_size`, `perf/rollout_wait_s`, `perf/mfu_fwd_bwd`. Forward+backward drops by one forward; the queue falls from ~420 to ~60, and rollout wait rises from 0.02 s to 0.5 s: the bottleneck shifts to generation.*
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,387 @@
...
+thumbnail: /blog/assets/asyncgrpo-lora-hfjobs/thumbnail.png
+authors:
+  - user: aminediroHF
+  - user: qgallouedec
+  - user: kashif
+---
+
+# Async GRPO with LoRA across Hugging Face Jobs: a bucket, a proxy, and no NCCL
+
+LoRA support recently landed in TRL's `AsyncGRPOTrainer` with [PR #7017](https://github.com/huggingface/trl/pull/7017). The asynchronous trainer can now train an adapter instead of the full model, and it syncs only the LoRA adapter to vLLM. This post covers a real-world project built on top of it, where training and inference no longer share a machine.
```
qgallouedec: Maybe mentin the release (version) along with the PR

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,387 @@
...
+
+```sh
+for replica in 1 2; do
+hf jobs run --detach --flavor h200 --timeout 8h --secrets HF_TOKEN \
+    --expose 8000 \                                          # reachable at https://<job_id>--8000.hf.jobs
+    -v "hf://buckets/${BUCKET}:/lora:ro" \                   # read-only: the server only reads adapters
+    -e VLLM_ALLOW_RUNTIME_LORA_UPDATING=1 \                  # enables /v1/load_lora_adapter
+    -e VLLM_SERVER_DEV_MODE=1 \                              # enables /pause, /resume, /server_info (TRL needs all three)
+    -- vllm/vllm-openai:v0.27.1 \
+    vllm serve Qwen/Qwen2.5-Math-1.5B --host 0.0.0.0 --port 8000 \
```
qgallouedec, line 63: Nice
although vllm changes fast, so it might make sense to mention somewhere the version used

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,387 @@
...
+        --enable-lora --max-lora-rank 1 --max-loras 6      # max_staleness=4 -> 4+2 adapter slots
+done
+```
+
+There is another possible design where the trainer keeps only the latest adapter and always publishes it under the same name. We did not go that way, because vLLM keys its prefix cache by adapter name. With a single name, KV blocks computed under the previous weights would still match after the swap, so the prefill would not be redone and a rollout could get its prefix from one policy version and its decode from the next. The trainer would have no way to tell, and it would show up as `ratio` drifting away from 1. Versioned names make this impossible: a name always means one set of weights, and a cached prefix can never match a newer version.
+
+### The dataset choice: the Sanity set
+
+We chose [`sail/Sanity-Test-R1D-1.5B`](https://huggingface.co/datasets/sail/Sanity-Test-R1D-1.5B), the dataset from [Defeating the Training-Inference Mismatch via FP16](https://arxiv.org/pdf/2510.26788) (Qi et al., 2025). The reproduction code is in [`sail-sg/Precision-RL`](https://github.com/sail-sg/Precision-RL).
```
qgallouedec, line 75: Nice

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,387 @@
...
+
+The authors generated 40 answers for each MATH problem with DeepSeek-R1-Distill-Qwen-1.5B. They kept problems with a success rate between 20% and 80%, yielding 1,460 questions. This dataset is really good for RL validation because the questions are neither already solved nor completely hopeless for that model, meaning the model can get a good early signal to train on and improve.
```
qgallouedec: Link to the model

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,387 @@
...
+
+This is awesome as a robust end-to-end test: if one vLLM replica silently serves the base model under an adapter name, we want to see that in the curve within a few dozen steps. Also, this dataset is small enough to cycle through in less than two hours.
+
+We also take the hyperparameters from the paper's LoRA scripts in [`oat/scripts/lora`](https://github.com/sail-sg/Precision-RL/tree/main/oat/scripts/lora): `Qwen/Qwen2.5-Math-1.5B`, LoRA rank 1 with alpha 2, a learning rate of 4e-5, 8 samples per prompt, 128 completions per step, a maximum of 3,000 generated tokens and a 4,096-token context.
+
+### The trainer
+
+The trainer uses the same image with TRL installed from the PR branch (now from main). The training script is a normal `AsyncGRPOTrainer` script. The only Job-specific values are the output directory and the server URL.
```
qgallouedec: Now from the latest version 😉

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,387 @@
...
+> [!NOTE]
+> During initialization, TRL calls `/server_info`. If it finds a `lora_config`, it uses adapter-only sync. Configurations vLLM cannot serve directly, such as DoRA, `modules_to_save`, or a rank above `--max-lora-rank`, fall back to merged-weight sync with a warning. The log should contain `Adapter-only vLLM sync enabled`.
+
+## The proxy
+
+Now onto the fun stuff. We need a proxy between the trainer and the vLLM Jobs for two reasons:
+
+1. Exposed Job ports require an `Authorization: Bearer <HF token>` header on every request. The proxy is where that header gets added, so TRL does not need to know about it.
+
+2. TRL refuses adapter-only sync when vLLM runs with `--data-parallel-size > 1`. This is a vLLM limitation rather than a TRL one. A call to `/v1/load_lora_adapter` only reaches the replica that answers it, so the other replicas would keep serving the base model under the new policy name.
```
qgallouedec: Why would you use dp>1?

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,387 @@
...
+
+results = await asyncio.gather(*(load_one(u) for u in ups))
+if any(st != 200 for _, st, _ in results):
+    await asyncio.gather(*(send(u, "POST", "/v1/unload_lora_adapter", headers, unload) for u, st, _ in results if st == 200))
+    return web.Response(status=504 if timed_out else st, text="rolled back on the others")
+```
+
+We also broadcast `/pause`, `/resume` and `/v1/unload_lora_adapter` in the same way. `/health` returns 200 only if every replica is healthy. `/server_info` and `/v1/models` only need one answer. From TRL's point of view, the proxy is a single `data_parallel_size=1` server, so it selects adapter-only sync.
+
+We initially wondered whether a Python asyncio proxy would become a bottleneck. It does not (at least at this scale). There are at most 128 non-streaming JSON requests in flight, and routing only computes a few hashes. One thread handles this easily. A more refined router that needs to handle more traffic would probably need to be written in a faster language (I see you 🦀).
```
qgallouedec, line 209: 👍

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,387 @@
...
+
+*Figure 5. All five trackio runs (`r1-dp2`, `r1-dp2-tb16k`, `r1-dp2-tb16k-nockpt`, `r1-dp3-tb16k-nockpt`, `r1-dp3-inflight384`) overlaid, x-axis in steps. Panels: `perf/step_s`, `reward`, `sample/rollout_queue_size`, `sample/staleness_mean`. Step time falls from 22.9 s to 4.8 s across the series while the reward curves stay on top of each other; the last run's queue refills to ~690 of 768 and its staleness settles at 2. Runs 2 to 4 were stopped early once the dashboard had answered the question.*
+
+With 384 requests in flight, each replica gets 128. The queue quickly fills to around 690 out of 768 and stays there. Backpressure returns to 5 seconds and rollout wait falls to 0.03 seconds. Training is the bottleneck again! Forward and backward take 4.6 seconds, weight sync adds an amortized 1.5 seconds, and median step time is 4.8 seconds.
+
+Mean staleness rises from 1.5 to 2.0 versions because samples wait longer in the larger queue. This is still below `max_staleness=4`, and `ratio` remains very close to 1.000.
+
+### The scoreboard
```
qgallouedec, line 341: Nice flow for this section

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,387 @@
...
+![Figure 6](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/asyncgrpo-lora-hfjobs/fig6-reward-vs-wallclock.png)
+
+*Figure 6. trackio runs `r1-dp2` and `r1-dp3-inflight384`, reward against wall-clock minutes since the first optimizer step. Same recipe, same 500 steps, same final reward; run 5 gets there in 52 minutes instead of 3 h 26 min.*
+
+The final run is 3.9× faster and trains on 31 % more samples, with basically the same reward curve. Packing, disabling checkpointing and raising the in-flight limit made the difference. In each case, the dashboard made this clear within the first ten minutes.
+
+## Try it
+
+```sh
+git clone https://github.com/AmineDiro/hfjobs-lora-buckets && cd hfjobs-lora-buckets
```
qgallouedec, line 364: We could try to have this in trl/examples

qgallouedec (commented):
Looks good, finally this is acheived, and quite clean!!

asyncgrpo-lora-hfjobs.md
AmineDiro: I was waiting for the PR to land, will do 👍🏼

asyncgrpo-lora-hfjobs.md
AmineDiro: just to have multiple vllm instances serving the model for higher inference throughput

asyncgrpo-lora-hfjobs.md
AmineDiro: good idea, I'll push a PR

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
+---
+title: "Async GRPO with LoRA across Hugging Face Jobs: a bucket, a proxy, and no NCCL"
+thumbnail: /blog/assets/asyncgrpo-lora-hfjobs/thumbnail.png
+authors:
+  - user: aminediroHF
+  - user: qgallouedec
+  - user: kashif
+---
+
+# Async GRPO with LoRA across Hugging Face Jobs: a bucket, a proxy, and no NCCL
```
sergiopaniego, line 10:
```suggestion
# Async GRPO with LoRA across HF Jobs: a bucket, a proxy, and no NCCL
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+
+LoRA support recently landed in TRL's `AsyncGRPOTrainer` with [PR #7017](https://github.com/huggingface/trl/pull/7017), and ships with TRL v1.14. The asynchronous trainer can now train an adapter instead of the full model, and it syncs only the LoRA adapter to vLLM. This post covers a real-world project built on top of it, where training and inference no longer share a machine.
```
sergiopaniego, line 12: - TL;DR as start of para maybe
- link to AsyncGRPOTrainer's docs?

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+
+LoRA training is particularly suited for RL, as stated in the amazing Thinking Machines post [LoRA Without Regret](https://thinkingmachines.ai/blog/lora/). They show that LoRA can match full fine-tuning for policy-gradient RL, even with rank 1. This stems from the fact that the advantage function only gives `~O(1)` bits of information per episode, so there is not that much to learn from each step, from a total-bits-of-information point of view. A rank-1 adapter has enough capacity to absorb it.
```
sergiopaniego, line 14:
```suggestion
LoRA training is particularly suited for RL, as shown in Thinking Machines's blog [LoRA Without Regret](https://thinkingmachines.ai/blog/lora/). They show that LoRA can match full fine-tuning for policy-gradient RL, even with rank 1. This stems from the fact that the advantage function only gives `~O(1)` bits of information per episode, so there is not that much to learn from each step, from a total-bits-of-information point of view. A rank-1 adapter has enough capacity to absorb it.
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+
+There is also a nice systems consequence of LoRA training. A rank-1 adapter for a 1.5B model is a few megabytes, while the full model is around 3 GB. Instead of sending the full policy to the inference workers after every update, we can just send the adapter. vLLM can also keep several adapters loaded at once. Old rollouts finish with the policy they started with, while new rollouts use the latest one.
```
sergiopaniego, line 16:
```suggestion
There is also a systems consequence of LoRA training. A rank-1 adapter for a 1.5B model is a few megabytes, while the full model is around 3 GB. Instead of sending the full policy to the inference workers after every update, we can just send the adapter. vLLM can also keep several adapters loaded at once. Old rollouts finish with the policy they started with, while new rollouts use the latest one.
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
...s share a filesystem or can form an NCCL group. What we want is to run the same setup with [Hugging Face Jobs](https://huggingface.co/docs/huggingface_hub/guides/jobs). Essentially, an HF Job is one container running on one VM. This means that one Job cannot spawn multiple nodes (at least for now) to hold a trainer and a fleet of vLLM servers (we are limited to 8xH200 at most per node). The `AsyncGRPOTrainer` is built for exactly that kind of scale, so the question became: how far can we get if we drop the requirement that the trainer and the inference servers share a node? Well, with a full-weight sync, the answer would be "not far". Every update would have to move gigabytes between machines, which is what NCCL is for in a dense cluster, but Jobs can't communicate _across_ nodes. There is no shared local disk and obviously no shared `localhost`. With LoRA, a sync is only a few megabytes. For the filesystem part, HF Jobs provide volumes backed by [Storage Buckets](https://huggingface.co/docs/hub/storage-buckets)! These buckets can then be mounted as a FUSE filesystem in every Job and are enough to work as a shared FS between nodes. No network path between the Jobs is needed at all.
```
sergiopaniego, line 18: This paragraph could be divided, it's too long

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+The setup ended up being quite small:
+
+- a **trainer Job** running `AsyncGRPOTrainer` with LoRA (and FSDP, more on that later),
+- **two vLLM Jobs**, each serving the base model plus whatever adapter the trainer last published,
+- a **Storage Bucket** mounted in all three at the same path, which is how the adapter gets from the trainer to the servers,
+- a **proxy server**. We'll dive deeper into why we need one, but at a high level we need a proxy that routes each rollout to the replica most likely to hold its KV cache, and broadcasts every adapter update to all vLLM replicas.
+
+## The architecture: leveraging Hugging Face Jobs and Storage Buckets 🪣
+
+[TRL PR #7017](https://github.com/huggingface/trl/pull/7017) adds an adapter-only sync path to `AsyncGRPOTrainer`. The trainer does not send tensors to vLLM. Every few optimizer steps, it saves the adapter under `<output_dir>/.vllm_lora/trl-policy-v{N}`, publishes the directory with an atomic rename, then sends its path to vLLM's `/v1/load_lora_adapter` endpoint. vLLM loads the files from disk, so the rollout worker can then request `model="trl-policy-v{N}"`.
```
sergiopaniego, line 29: I'd avoid repeating the PR with its number

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+
+results = await asyncio.gather(*(load_one(u) for u in ups))
+if any(st != 200 for _, st, _ in results):
+    await asyncio.gather(*(send(u, "POST", "/v1/unload_lora_adapter", headers, unload) for u, st, _ in results if st == 200))
+    return web.Response(status=504 if timed_out else st, text="rolled back on the others")
+```
+
+We also broadcast `/pause`, `/resume` and `/v1/unload_lora_adapter` in the same way. `/health` returns 200 only if every replica is healthy. `/server_info` and `/v1/models` only need one answer. From TRL's point of view, the proxy is a single `data_parallel_size=1` server, so it selects adapter-only sync.
+
+We initially wondered whether a Python asyncio proxy would become a bottleneck. It does not (at least at this scale). There are at most 128 non-streaming JSON requests in flight, and routing only computes a few hashes. One thread handles this easily. A more refined router that needs to handle more traffic would probably need to be written in a faster language (I see you 🦀).
```
sergiopaniego, line 209:
```suggestion
We initially wondered whether a Python asyncio proxy would become a bottleneck. It does not (at least at this scale). There are at most 128 non-streaming JSON requests in flight, and routing only computes a few hashes. One thread handles this easily. A more refined router that needs to handle more traffic would probably need to be written in a faster language (We see you 🦀).
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+```
+
+We pin vLLM to `v0.27.1`. vLLM moves fast, and the flags above and the runtime LoRA endpoints are the ones that version exposes, so treat the version as part of the recipe.
+
+There is another possible design where the trainer keeps only the latest adapter and always publishes it under the same name. We did not go that way, because vLLM keys its prefix cache by adapter name. With a single name, KV blocks computed under the previous weights would still match after the swap, so the prefill would not be redone and a rollout could get its prefix from one policy version and its decode from the next. The trainer would have no way to tell, and it would show up as `ratio` drifting away from 1. Versioned names make this impossible: a name always means one set of weights, and a cached prefix can never match a newer version.
+
+### The dataset choice: the Sanity set
+
+We chose [`sail/Sanity-Test-R1D-1.5B`](https://huggingface.co/datasets/sail/Sanity-Test-R1D-1.5B), the dataset from [Defeating the Training-Inference Mismatch via FP16](https://arxiv.org/pdf/2510.26788) (Qi et al., 2025). The reproduction code is in [`sail-sg/Precision-RL`](https://github.com/sail-sg/Precision-RL).
```
sergiopaniego, line 75:
```suggestion
We chose [`sail/Sanity-Test-R1D-1.5B`](https://huggingface.co/datasets/sail/Sanity-Test-R1D-1.5B), the dataset from [Defeating the Training-Inference Mismatch via FP16](https://huggingface.co/papers/2510.26788) (Qi et al., 2025). The reproduction code is in [`sail-sg/Precision-RL`](https://github.com/sail-sg/Precision-RL).
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+- TRL [PR #7017](https://github.com/huggingface/trl/pull/7017): PEFT/LoRA support for `AsyncGRPOTrainer` with adapter-only vLLM sync, released in TRL v1.14.
+- Hugging Face [Jobs](https://huggingface.co/docs/huggingface_hub/guides/jobs) and [Storage Buckets](https://huggingface.co/docs/hub/storage-buckets); [`hf-mount`](https://github.com/huggingface/hf-mount).
+- [`hf-mount-repro`](https://github.com/AmineDiro/hf-mount-repro): the two-script reproduction of the 30-second negative-cache stall.
+- The [trackio dashboard](https://huggingface.co/spaces/aminediroHF/async-grpo-lora-buckets) for every run in this post.
+- Penghui Qi, Zichen Liu, Xiangxin Zhou, Tianyu Pang, Chao Du, Wee Sun Lee, Min Lin, [Defeating the Training-Inference Mismatch via FP16](https://arxiv.org/pdf/2510.26788), arXiv:2510.26788, 2025. Source of the Sanity dataset [`sail/Sanity-Test-R1D-1.5B`](https://huggingface.co/datasets/sail/Sanity-Test-R1D-1.5B) and of the LoRA recipe, [`sail-sg/Precision-RL`](https://github.com/sail-sg/Precision-RL), `oat/scripts/lora/bf16_grpo_tis_lora.sh`.
```
sergiopaniego, line 380:
```suggestion
- Penghui Qi, Zichen Liu, Xiangxin Zhou, Tianyu Pang, Chao Du, Wee Sun Lee, Min Lin, [Defeating the Training-Inference Mismatch via FP16](https://huggingface.co/papers/2510.26788), arXiv:2510.26788, 2025. Source of the Sanity dataset [`sail/Sanity-Test-R1D-1.5B`](https://huggingface.co/datasets/sail/Sanity-Test-R1D-1.5B) and of the LoRA recipe, [`sail-sg/Precision-RL`](https://github.com/sail-sg/Precision-RL), `oat/scripts/lora/bf16_grpo_tis_lora.sh`.
```

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+    --expose 8000 \                                          # reachable at https://<job_id>--8000.hf.jobs
+    -v "hf://buckets/${BUCKET}:/lora:ro" \                   # read-only: the server only reads adapters
+    -e VLLM_ALLOW_RUNTIME_LORA_UPDATING=1 \                  # enables /v1/load_lora_adapter
+    -e VLLM_SERVER_DEV_MODE=1 \                              # enables /pause, /resume, /server_info (TRL needs all three)
+    -- vllm/vllm-openai:v0.27.1 \
+    vllm serve Qwen/Qwen2.5-Math-1.5B --host 0.0.0.0 --port 8000 \
+        --max-model-len 4096 --logprobs-mode processed_logprobs --generation-config vllm \
+        --enable-lora --max-lora-rank 1 --max-loras 6      # max_staleness=4 -> 4+2 adapter slots
+done
+```
```
sergiopaniego, line 67: can a user copy this directly with the `\` and the `#`s?

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+This is awesome as a robust end-to-end test: if one vLLM replica silently serves the base model under an adapter name, we want to see that in the curve within a few dozen steps. Also, this dataset is small enough to cycle through in less than two hours.
+
+We also take the hyperparameters from the paper's LoRA scripts in [`oat/scripts/lora`](https://github.com/sail-sg/Precision-RL/tree/main/oat/scripts/lora): `Qwen/Qwen2.5-Math-1.5B`, LoRA rank 1 with alpha 2, a learning rate of 4e-5, 8 samples per prompt, 128 completions per step, a maximum of 3,000 generated tokens and a 4,096-token context.
+
+### The trainer
+
+The trainer uses the same `vllm/vllm-openai:v0.27.1` image with TRL installed on top. We ran the PR branch at the time; the same code now ships in TRL v1.14. The training script is a normal `AsyncGRPOTrainer` script. The only Job-specific values are the output directory and the server URL.
+
+```python
+config = AsyncGRPOConfig(
```
sergiopaniego, line 88: from trl.experimental... ?

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+| rollout queue occupancy | 476 of 512 |
+| trainer MFU | 3.9 % |
+
+The rollout queue stays full, and the worker is mostly blocked by backpressure. The second replica is basically useless in this configuration. We'll see later in the post how we went through runs that moved the bottleneck between training and generation to make the run 3.9× faster.
+
+### Reward
+
+![Figure 0](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/asyncgrpo-lora-hfjobs/fig0-r1-dp2-reward.png)
+
+*Figure 0. trackio run `r1-dp2`. Panels: `reward` with its 20-step rolling mean and 50-step block means, and `ratio` on a 0.99 to 1.01 axis. Reward climbs from 0.15 to 0.44 over 500 steps; `ratio` stays between 0.9993 and 1.0004 throughout.*
```
sergiopaniego, line 254: We could start the count from Fig. 1, no? Instead of Fig. 0

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+![Figure 6](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/blog/asyncgrpo-lora-hfjobs/fig6-reward-vs-wallclock.png)
+
+*Figure 6. trackio runs `r1-dp2` and `r1-dp3-inflight384`, reward against wall-clock minutes since the first optimizer step. Same recipe, same 500 steps, same final reward; run 5 gets there in 52 minutes instead of 3 h 26 min.*
+
+The final run is 3.9× faster and trains on 31 % more samples, with basically the same reward curve. Packing, disabling checkpointing and raising the in-flight limit made the difference. In each case, the dashboard made this clear within the first ten minutes.
+
+## Try it
+
+```sh
+git clone https://github.com/AmineDiro/hfjobs-lora-buckets && cd hfjobs-lora-buckets
```
sergiopaniego, line 364: link 404

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+  MAX_INFLIGHT=384 QUEUE_MAXSIZE=768 MAX_STEPS=500 ./run_all.sh --wait   # run 5: same recipe, ~55 min
+```
+
+## References
+
+- John Schulman et al., [LoRA Without Regret](https://thinkingmachines.ai/blog/lora/), Thinking Machines Lab, September 2025. The case that rank-1 LoRA matches full fine-tuning for policy-gradient RL, and why.
+- TRL, [`AsyncGRPOTrainer`](https://huggingface.co/docs/trl/en/async_grpo_trainer) and its [logged metrics](https://huggingface.co/docs/trl/en/async_grpo_trainer#logged-metrics).
+- TRL [PR #7017](https://github.com/huggingface/trl/pull/7017): PEFT/LoRA support for `AsyncGRPOTrainer` with adapter-only vLLM sync, released in TRL v1.14.
+- Hugging Face [Jobs](https://huggingface.co/docs/huggingface_hub/guides/jobs) and [Storage Buckets](https://huggingface.co/docs/hub/storage-buckets); [`hf-mount`](https://github.com/huggingface/hf-mount).
+- [`hf-mount-repro`](https://github.com/AmineDiro/hf-mount-repro): the two-script reproduction of the 30-second negative-cache stall.
```
sergiopaniego, line 378: 404

asyncgrpo-lora-hfjobs.md
```diff
@@ -0,0 +1,389 @@
...
+- Penghui Qi, Zichen Liu, Xiangxin Zhou, Tianyu Pang, Chao Du, Wee Sun Lee, Min Lin, [Defeating the Training-Inference Mismatch via FP16](https://arxiv.org/pdf/2510.26788), arXiv:2510.26788, 2025. Source of the Sanity dataset [`sail/Sanity-Test-R1D-1.5B`](https://huggingface.co/datasets/sail/Sanity-Test-R1D-1.5B) and of the LoRA recipe, [`sail-sg/Precision-RL`](https://github.com/sail-sg/Precision-RL), `oat/scripts/lora/bf16_grpo_tis_lora.sh`.
+
+```bibtex
+@article{qi2025precisionrl,
+  title={Defeating the Training-Inference Mismatch via FP16},
+  author={Qi, Penghui and Liu, Zichen and Zhou, Xiangxin and Pang, Tianyu and Du, Chao and Lee, Wee Sun and Lin, Min},
+  journal={arXiv preprint arXiv:2510.26788},
+  year={2025}
+}
+```
```
sergiopaniego, line 389:
```suggestion
```

asyncgrpo-lora-hfjobs.md
AmineDiro: Yeah this should work just fine in  a bash terminal

asyncgrpo-lora-hfjobs.md
AmineDiro: I have deep feelings about starting indexing from 1 🤣  but I do agree with you starting wiht 1 may be better

asyncgrpo-lora-hfjobs.md
AmineDiro: why remove the reference ?

kashif (approved):

